PR #5, opened 2015-07-23
Using ngBind instead of {{ expression }} to avoid header text flicker

vbhv:
Use ngBind for setting title, header and footer text

### DIFF
--- a/server/www/index.html
+++ b/server/www/index.html
@@ -10,7 +10,7 @@
         <link rel="stylesheet" href="/css/rtl-style.css" ng-disabled="! rtlEnabled">
         <link rel="stylesheet" href="/css/vendor.css">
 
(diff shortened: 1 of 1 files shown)


server/www/index.html
```diff
@@ -10,7 +10,7 @@
         <link rel="stylesheet" href="/css/rtl-style.css" ng-disabled="! rtlEnabled">
         <link rel="stylesheet" href="/css/vendor.css">
 
-        <title>{{pageTitle}}{{siteTitle}}</title>
+        <title ng-bind-template="{{pageTitle}} {{siteTitle}}"></title>
```
rjmackay: Do you think it'd make sense to add some default text here though? ie. `<title ng-bind-template="{{pageTitle}} {{siteTitle}}">Ushahidi</title>`

PR #7, opened 2015-08-27
T1744 iconpicker

willdoran:
@rjmackay @jasonmule This is the client side commit for the iconpicker - I need to updated the platform side to change the allowed format of the icon saved.


app/common/services/endpoints/NounEndpoint.js
```diff
@@ -0,0 +1,23 @@
+module.exports = [
```
rjmackay: Can we remove this since its not used?

app/common/services/icon-manager.js
```diff
@@ -0,0 +1,42 @@
...
+    var IconManager = {
+
+        getIconSetArray: function (iconLibName) {
+          var iconSetArray = [];
+          var iconLibNames = iconLibName.split(' ');
+
+          // @todo: simplify this to reduce loops
+          // push array construction to the endpoint
+          // and/or store values with prefix already attached
+          _.each(iconLibNames, function (libName) {
```
rjmackay: I'd like to see something similar.. but as long as this works lets land it and move on.

rjmackay:
Looking pretty good.
:+1: happy to land.

PR #8, opened 2015-09-02
T638 - Restrict views to those enabled in backend

hollycorbett:
Views stuff works fine in concert with the backend changes.  Datasources still needs love.

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/8)

PR #10, opened 2015-09-04
Restrict views to those enabled in backend

rjmackay:
- Check view availability via features config
- [x] Disable data sources based on features: Ideally this should show the data source in settings, but with a 'not available, please upgrade' message.

Replace #8
Refs ushahidi/platform#638

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/10)

### DIFF
--- a/app/common/locales/en.json
+++ b/app/common/locales/en.json
@@ -7,7 +7,7 @@
         "forbidden": "Sorry, you're not allowed to do that."
     },
     "feature_limits": {
-        "view_unavailable" : "This feature is not currently available. To gain access to this feature it is necessary to upgrade to a higher tier of Ushahidi"
+        "view_unavailable" : "This view isn't available on your current plan. Upgrade your plan"
     },
     "form": {
         "add_field" : "Add New Field",
@@ -396,7 +396,7 @@
             "provider_options": "Options",
             "save_changes": "Save Changes",
             "saving_changes": "Saving",
-            "unavailable": "This feature is unavailable at your Ushahidi tier. To get access to this feature upgrade to a higher tier."
+            "unavailable": " is unavailable for your Ushahidi plan. To get access to this feature upgrade your Ushahidi plan."
         },
         "generic_field_error": "Invalid entry, please try again.",
         "skip" : "Skip this step",
--- a/app/common/locales/en.json
+++ b/app/common/locales/en.json
@@ -392,7 +392,8 @@
             "hint_plugin_needs_configuring": "You need to configure this data source before you can enable it.",
             "provider_options": "Options",
             "save_changes": "Save Changes",
-            "saving_changes": "Saving"
+            "saving_changes": "Saving",
+            "unavailable": "This feature is unavailable at your Ushahidi tier. To get access to this feature upgrade to a higher tier."
         },
         "generic_field_error": "Invalid entry, please try again.",
         "skip" : "Skip this step",
--- a/app/common/locales/en.json
+++ b/app/common/locales/en.json
@@ -6,6 +6,9 @@
         "report_a_bug": "Report a bug",
         "forbidden": "Sorry, you're not allowed to do that."
     },
+    "feature_limits": {
+        "view_unavailable" : "This feature is not currently available. To gain access to this feature it is necessary to upgrade to a higher tier of Ushahidi"
+    },
     "form": {
         "add_field" : "Add New Field",
         "add_field_instructions": "Add instructions to this field",
--- a/app/common/locales/en.json
+++ b/app/common/locales/en.json
@@ -7,7 +7,7 @@
         "forbidden": "Sorry, you're not allowed to do that."
     },
     "feature_limits": {
-        "view_unavailable" : "This view isn't available on your current plan. Upgrade your plan"
+        "view_unavailable" : "The {{value}} view isn't available on your current plan. Upgrade your plan"
     },
     "form": {
         "add_field" : "Add New Field",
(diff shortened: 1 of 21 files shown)


rjmackay:
@hollycorbett work on this branch if you continue with this. I'll probably work on other tasks for a bit before I'm back to it

rjmackay:
Looking pretty good. Minor tweaks to the messaging needed then we can land this when the backend change lands.
I'm kinda wonder if we should land to a 'cloud' branch or fork rather than into the main repo. What do you think?

---

Reviewed 1 of 2 files at r1, 2 of 3 files at r2, 13 of 13 files at r3.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

<sup>**[app/common/locales/en.json, line 10 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform-client/10#-Jzh-5dX99UNivkq0u8d)** ([raw file](https://github.com/ushahidi/platform-client/blob/38ea128451f1625d0f53ad0e7b07f9b9d17f658a/app/common/locales/en.json#L10)):</sup>
We're using the term 'plan' not tier throughout the UI. Update this to something like
"This feature isn't available on your current plan. <a>Upgrade your plan</a>" 
.. ideally we'd customise the message so it uses the actual feature name..

---

<sup>**[server/www/templates/settings/datasources.html, line 16 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform-client/10#-Jzh0EMoE8eNk7U-fdjW)** ([raw file](https://github.com/ushahidi/platform-client/blob/38ea128451f1625d0f53ad0e7b07f9b9d17f658a/server/www/templates/settings/datasources.html#L16)):</sup>
any reason for repeating the input toggle rather than just adding `available_providers[provider.id]` to the ng-disabled attribute?

---

<sup>**[server/www/templates/settings/datasources.html, line 75 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform-client/10#-Jzh0R3d9ZjpCFskbZ6y)** ([raw file](https://github.com/ushahidi/platform-client/blob/38ea128451f1625d0f53ad0e7b07f9b9d17f658a/server/www/templates/settings/datasources.html#L75)):</sup>
Could we pass the name of the provider into translation values so we can make the message bit better? ie. FrontlineSMS isn't available on your current plan...

---

<sup>**[server/www/templates/views/chart.html, line 4 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform-client/10#-Jzh0bDzPPaHUT7dJlL-)** ([raw file](https://github.com/ushahidi/platform-client/blob/38ea128451f1625d0f53ad0e7b07f9b9d17f658a/server/www/templates/views/chart.html#L4)):</sup>
Like this.. though its possible you could handle this a level higher and only have to do it once?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform-client/10)

willdoran:
Are we ultimately going to maintain two versions? I find the idea of multiple versions terrifying :D I think I'd ultimately vote for a branch as it's more straightforward to reintegrate. I needed to know more about the goal of a cloud branch.

---

Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

<sup>**[server/www/templates/settings/datasources.html, line 16 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform-client/10#-Jzh0EMoE8eNk7U-fdjW)** ([raw file](https://github.com/ushahidi/platform-client/blob/38ea128451f1625d0f53ad0e7b07f9b9d17f658a/server/www/templates/settings/datasources.html#L16)):</sup>
Yep, I tried to use ng-disabled originally but it failed in conjunction with the other variables. It entered a weird state with the "saving" and ng-model settings. I'm not totally sure that saving is actually working.
I think it might be better to revisit it after I land this. Is that acceptable?

---

<sup>**[server/www/templates/views/chart.html, line 4 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform-client/10#-Jzh0bDzPPaHUT7dJlL-)** ([raw file](https://github.com/ushahidi/platform-client/blob/38ea128451f1625d0f53ad0e7b07f9b9d17f658a/server/www/templates/views/chart.html#L4)):</sup>
Hmmm, yeah doing it in the main controller might be possible. I'll try to do that in the next hour.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform-client/10)

rjmackay:
> Are we ultimately going to maintain two versions? I find the idea of multiple versions terrifying :D I think I'd ultimately vote for a branch as it's more straightforward to reintegrate. I needed to know more about the goal of a cloud branch. 

Ideally no. However we have no framework for plugins or tweaking the UI. I'll keep the UI stuff on a branch for now, and either move to a plugin or merge into master later

rjmackay:
Review status: 6 of 18 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

<sup>**[app/post/directives/views/post-view-unavailable-directive.js, line 14 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/10#-JzqnR9VcBqD45ascixU)** ([raw file](https://github.com/ushahidi/platform-client/blob/a9d869a26da360f142aab7ab233e01e3bcad3904/app/post/directives/views/post-view-unavailable-directive.js#L14)):</sup>
Can we lose the console.log? otherwise things get noisy

---

<sup>**[server/www/templates/views/views.html, line 22 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/10#-Jzqo-fTjdePZz8xzwOm)** ([raw file](https://github.com/ushahidi/platform-client/blob/a9d869a26da360f142aab7ab233e01e3bcad3904/server/www/templates/views/views.html#L22)):</sup>
We should be able to just do this once.. 
- Check the view in the post-view-directive
- if it doesn't exist, change it to 'unavailable'
- move this to the an `ng-switch-when="unavailable"`
make sense?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform-client/10)

PR #20, opened 2015-09-24
Cache layer

willdoran:
[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/20)


willdoran:
@rjmackay I've added caching for almost all the settings entites with the exception of the User page. For individual users this is straightforward, however, there is an issue the user list page.

The angular-cache lib determines the key itself and this doesn't appear to be overrideable. The key for the user page is:
'/user?limit=x&offet=y&role=z'
if there is a search on the page q is also added. When a user is created or updated we want to invalidate the user list cache. Doing this is somewhat of high wire act - it's not an easy to access the query params inside the user create/edit. 

The simplest though potentially most horrorific solution is to get the userCache keys - loop through them and invalidate any that contain specifically '/user?' - this will work though is not very nice.

On the plus side the caching makes the page load for all settings entities super fast :D

willdoran:
@jasonmule @rjmackay could you guys give this a look and poke holes in it for me? It covers the settings menu to start with but if it's acceptable then I can expand it to cover everything. It makes a dramatic improvement to page loads.

app/common/services/endpoints/config.js
```diff
@@ -2,24 +2,35 @@ module.exports = [
...
     $resource,
     Util,
-    _
+    _,
+    CacheFactory
 ) {
+    var cache = new CacheFactory('configCache');
+
+    cache.setOnExpire(function (key, value) {
+        ConfigEndpoint.get(value.id);
```
rjmackay: why are we doing this on some endpoints but not others?

app/common/services/endpoints/user-endpoint.js
```diff
@@ -2,11 +2,23 @@ module.exports = [
...
 function (
     $resource,
     $rootScope,
-    Util
+    Util,
+    CacheFactory
 ) {
+    var cache;
+
+    if (!(cache = CacheFactory.get('userCache'))) {
```
rjmackay: why are we doing special handling here but nowhere else? Doesn't CacheFactory have a method to automatically create a cache if it doesn't exist?

app/common/services/endpoints/config.js
```diff
@@ -2,24 +2,35 @@ module.exports = [
...
     '_',
+    'CacheFactory',
 function (
     $resource,
     Util,
-    _
+    _,
+    CacheFactory
 ) {
+    var cache = new CacheFactory('configCache');
```
rjmackay: I don't think you're supposed to `new` CacheFactory.. docs simply have `profileCache = CacheFactory('profileCache');`

app/common/services/util.js
```diff
@@ -14,7 +14,7 @@ function (
             return CONST.API_URL + relative_url;
         },
         transformResponse: function (response, omitKeys) {
-            omitKeys = (omitKeys || []).concat(['url', 'allowed_methods']);
+            omitKeys = (omitKeys || []).concat(['allowed_methods']);
```
rjmackay: Why did this change?

app/setting/controllers/setting-categories-controller.js
```diff
@@ -33,7 +35,10 @@ function (
                 angular.forEach($scope.selectedTags, function (tagId) {
                     calls.push(TagEndpoint.delete({ id: tagId }).$promise);
                 });
-                $q.all(calls).then($scope.refreshView);
+                $q.all(calls).then(function () {
+                    CacheManager.removeCacheGroup('tagCache', '/tags');
```
rjmackay: shouldn't this happen in the endpoint itself? 
components outside of an endpoint shouldn't really have to know about internals of its caching.

app/setting/controllers/setting-categories-edit-controller.js
```diff
@@ -33,6 +37,12 @@ function (
     $scope.saveTag = function (tag) {
         $scope.processing = true;
         TagEndpoint.update({id: $routeParams.id}, tag, function () {
+            CacheManager.updateCacheItem(
```
rjmackay: Can we bump this up into the endpoint itself? there must be some way to catch the responses there..

rjmackay:
In general if we can avoid calling Cache or CacheManager from outside of endpoints then I think we've got the separation of concerns right. We might need to expose a method on endpoints for clearing cache or removing items eventually.. but controllers shouldn't have to know about the cache itself. Does that make sense? Happy to chat about structure if it helps, but I don't want to dictate either.

willdoran:
yeah I entirely agree, I couldn't find a way to intercept within the endpoint in order to invalidate the cache prior to the get. I'll go through it again to see if I can find a way around it or a way to wrap it closer to the endpoint.

rjmackay:
I suspect you could just wrap it.. ie:
```
var resource = $resource(...);

return {
  get: resource.get,
  put: function() {
   var result = resource.put.call(resource, arguements);
   result.then(cacheWipingStuff);
   return result;
  }
};
```

rjmackay:
.. you might be able to do similar to give all the caching better keys, rather than having key = url as it is now.

willdoran:
Ok, neat, I'll try that!

willdoran:
that would definitely be preferable

app/common/services/endpoints/user-endpoint.js
willdoran: I'll check this but I added it because the spec test was failing as the cache had already been created. I'm not sure why that failure would occur but it doesn't seem that the lib checks for existence itself. So I should propagate this to all the endpoints.

willdoran:
@rjmackay, ok, I think I've fixed all of that.  I need to test the delete and add for posts. But I just want to make sure I haven't gone off the rails.

With the exception of the key. I can cheat and intercept everything so that we can control the keys but that would be both gross and I think unnecessary. I realised that for fast load - I think you pointed this out before. We're interested in the large groups of items being cached. We want to - I think - load fresh when editing individual entities.

When an entity is updated we want to load the query cache again and since we will be loading all entities fresh we can just invalidate the cache as a block in those circumstances. I think this will actually strike a good balance between caching for searching and browsing, and editing fresh.

willdoran:
@jasonmule, would you mind taking a look over this to see if you can poke any holes/make any suggestions?

jasonmule:
I am thinking that we should continue to use save, update etc in the controllers as usual and then add cache logic to these methods on the endpoints?

willdoran:
@jasonmule yep, that seems reasonable - I think that's what we have now or are you suggesting a different route?

rjmackay:
Reviewed 2 of 13 files at r2, 7 of 21 files at r3, 2 of 19 files at r4, 19 of 19 files at r5.
Review status: all files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---

<sup>**[app/common/services/endpoints/config.js, line 31 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2JfsANqlNc_qmNo1Hc)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/common/services/endpoints/config.js#L31)):</sup>
If we only pass cache to `get` .. do we end up caching `query()` at all? 
TBH only caching get might actually make the most sense anyway.N


---

<sup>**[app/common/services/endpoints/config.js, line 41 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2JfGE_B3j7rns8eYBN)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/common/services/endpoints/config.js#L41)):</sup>
I like this.. simple API, minimal thought required :+1: 


---

<sup>**[app/common/services/endpoints/config.js, line 46 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2Jf_QSRvAQ03tOj5xF)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/common/services/endpoints/config.js#L46)):</sup>
Could you docblock this? I'm not 100% how/why this is used?
ie. it doesn't just save to cache.. it saves to API, but then clears the cache?


---

<sup>**[app/common/services/endpoints/config.js, line 47 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2JfOlT84R51WmxNxFl)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/common/services/endpoints/config.js#L47)):</sup>
.. you know we could alias this on ALL our endpoints since we do this all over the place. **not now** but something for later / in between tasks.


---

<sup>**[app/common/services/endpoints/config.js, line 49 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2JfmH9_lNQ-0h03UYK)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/common/services/endpoints/config.js#L49)):</sup>
Why are we clearing everything not just this item? 


---

<sup>**[app/common/services/endpoints/data-providers.js, line 13 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2JgpGWB3e3CIpFyc8F)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/common/services/endpoints/data-providers.js#L13)):</sup>
Do we need to check this exists first? as done in other endpoints?


---

<sup>**[app/common/services/endpoints/data-providers.js, line 24 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2JgnnqEZaeHlqvLqqr)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/common/services/endpoints/data-providers.js#L24)):</sup>
only caching query, not get?


---

<sup>**[app/common/services/endpoints/form-attributes.js, line 52 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2JhZRfCUacXMqEW00n)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/common/services/endpoints/form-attributes.js#L52)):</sup>
probably same as below. Save should clear cache...


---

<sup>**[app/common/services/endpoints/form-attributes.js, line 59 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2JhIpK-bTc4rU_013D)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/common/services/endpoints/form-attributes.js#L59)):</sup>
I think this should just replace `delete` .. you'll probably have to rename the local var for `FormAttributeEndpoint` so you can access the original `.get` but I think it makes sense that all delete calls should clear the cache?


---

<sup>**[app/set/services/endpoints/collection.js, line 12 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2Jibn0MRbUS3bHSxhE)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/set/services/endpoints/collection.js#L12)):</sup>
Revert this if we're not using it here yet?


---

<sup>**[app/set/services/endpoints/savedsearch.js, line 10 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2JifCK5WBx48rb9hKz)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/set/services/endpoints/savedsearch.js#L10)):</sup>
Same here. revert the changes on this endpoint?


---

<sup>**[app/setting/controllers/setting-categories-controller.js, line 39 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K-Zk8-BHmDbZ6u7r6cY-r3-39)** ([raw file](https://github.com/ushahidi/platform-client/blob/b2f12eb3c8bc82d6152439a45d7ea21629a2d706/app/setting/controllers/setting-categories-controller.js#L39)):</sup>
This ends up being a no-change now right? Could just go back to
`$q.all(calls).then($scope.refreshView);`


---

<sup>**[app/setting/controllers/setting-categories-edit-controller.js, line 37 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2JjEGQEDPH0wCkfpJ_)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/setting/controllers/setting-categories-edit-controller.js#L37)):</sup>
We could probably maintain the original API, ie. allow callbacks on saveCache and deleteCache too.. Maybe just not that as a `@todo` in the code?


---

<sup>**[app/setting/controllers/setting-datasources-controller.js, line 58 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2Jjaor0yo6bVb6Vq41)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/setting/controllers/setting-datasources-controller.js#L58)):</sup>
Why are we messing with caches in the controller? Can we move this to use getFresh and QueryFresh instead?


---

<sup>**[app/setting/controllers/setting-users-controller.js, line 82 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2JjpTXUltMeEgLwI_3)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/setting/controllers/setting-users-controller.js#L82)):</sup>
Its not clear why this changed?


---

<sup>**[app/setting/controllers/setting-users-controller.js, line 102 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2JjsVnTbtpiyWuFyDO)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/setting/controllers/setting-users-controller.js#L102)):</sup>
.. same here?


---

<sup>**[app/common/services/cache-manager.js, line 0 \[r4\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2Jk7CPE-eYCgnaKNoE)** ([raw file](https://github.com/ushahidi/platform-client/blob/2ffbe2326ee2eacc881d1ecb3d2587d1cbb8e13c/app/common/services/cache-manager.js#L0)):</sup>
is this just reviewable being weird? Does cache-manager actually exist?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform-client/20)

willdoran:
Review status: 15 of 29 files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

<sup>**[app/common/services/endpoints/config.js, line 31 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2JfsANqlNc_qmNo1Hc)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/common/services/endpoints/config.js#L31)):</sup>
Nope, it only caches what is set to cache, so query isn't automatically cached. Though query isn't quite appropriate for config which I think is why it's not used.


---

<sup>**[app/common/services/endpoints/config.js, line 46 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2Jf_QSRvAQ03tOj5xF)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/common/services/endpoints/config.js#L46)):</sup>
Done.


---

<sup>**[app/common/services/endpoints/config.js, line 49 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2JfmH9_lNQ-0h03UYK)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/common/services/endpoints/config.js#L49)):</sup>
The cache is always small at this point, since it would contain only the query and the get at most. Clearing the whole thing is simpler than trying to find keys for some of the more complex relations and gives the same result. But I'll document it with examples to make sure that's clear.


---

<sup>**[app/common/services/endpoints/data-providers.js, line 13 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2JgpGWB3e3CIpFyc8F)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/common/services/endpoints/data-providers.js#L13)):</sup>
Done.


---

<sup>**[app/common/services/endpoints/data-providers.js, line 24 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2JgnnqEZaeHlqvLqqr)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/common/services/endpoints/data-providers.js#L24)):</sup>
Nope, need to add.


---

<sup>**[app/common/services/endpoints/form-attributes.js, line 52 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2JhZRfCUacXMqEW00n)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/common/services/endpoints/form-attributes.js#L52)):</sup>
Done.


---

<sup>**[app/common/services/endpoints/form-attributes.js, line 59 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2JhIpK-bTc4rU_013D)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/common/services/endpoints/form-attributes.js#L59)):</sup>
Done.


---

<sup>**[app/set/services/endpoints/collection.js, line 12 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2Jibn0MRbUS3bHSxhE)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/set/services/endpoints/collection.js#L12)):</sup>
Done.


---

<sup>**[app/set/services/endpoints/savedsearch.js, line 10 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2JifCK5WBx48rb9hKz)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/set/services/endpoints/savedsearch.js#L10)):</sup>
Done.


---

<sup>**[app/setting/controllers/setting-categories-controller.js, line 39 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K-Zk8-BHmDbZ6u7r6cY-r3-39)** ([raw file](https://github.com/ushahidi/platform-client/blob/b2f12eb3c8bc82d6152439a45d7ea21629a2d706/app/setting/controllers/setting-categories-controller.js#L39)):</sup>
I'll check and make sure.


---

<sup>**[app/setting/controllers/setting-categories-edit-controller.js, line 37 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2JjEGQEDPH0wCkfpJ_)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/setting/controllers/setting-categories-edit-controller.js#L37)):</sup>
Cool


---

<sup>**[app/setting/controllers/setting-datasources-controller.js, line 58 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2Jjaor0yo6bVb6Vq41)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/setting/controllers/setting-datasources-controller.js#L58)):</sup>
This should have been removed, will do now.


---

<sup>**[app/setting/controllers/setting-users-controller.js, line 82 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2JjpTXUltMeEgLwI_3)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/setting/controllers/setting-users-controller.js#L82)):</sup>
$scope.filterRole - as far as I can find though I may be missing it - does not connect to anything. So the user list was never refreshing after changes. This fixes that.


---

<sup>**[app/setting/controllers/setting-users-controller.js, line 102 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2JjsVnTbtpiyWuFyDO)** ([raw file](https://github.com/ushahidi/platform-client/blob/f464f2c87b55b8425918c7bed4b0cfcb242aa540/app/setting/controllers/setting-users-controller.js#L102)):</sup>
Same as above


---

<sup>**[app/common/services/cache-manager.js, line 0 \[r4\]](https://reviewable.io:443/reviews/ushahidi/platform-client/20#-K2Jk7CPE-eYCgnaKNoE)** ([raw file](https://github.com/ushahidi/platform-client/blob/2ffbe2326ee2eacc881d1ecb3d2587d1cbb8e13c/app/common/services/cache-manager.js#L0)):</sup>
Yeah, not sure. Can't find it in my local


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform-client/20)

PR #21, opened 2015-09-27
Issue 303 - better feedback after saving post

hollycorbett:
Basic functionality after saving post.

Do we want to push this out to a generic directive based off the modal directive?  I've gone back and forth about it a couple of times for the sake of issue 468, but in this case its without.  Can easily change.

Delete and error notification coming soon after based on same approach.  Let me know if you want a different style from this - I was largely intending on doing a c+p job.

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/21)


rjmackay:
erm.. I think you borked the PR? Its pulling in changes to do with feature toggling.. but nothing to do with posts feedback? Wrong branch?

hollycorbett:
Okay.  This time.  Colour me confuzzled.  Travis still hates me, but lets say thats on my list for figuring out later.  Currently doing the same thing as in this one for deletioning posts.

app/post/controllers/post-detail-controller.js
```diff
@@ -140,13 +142,19 @@ function (
...
+    };
+
+    $scope.cancelDelete = function () {
+        // again - feel there must be a more angular-ish way to do this
+        $scope.userRequestedPostDelete = false;
+    };
+
     $scope.deletePost = function () {
-        $translate('notify.post.destroy_confirm').then(function (message) {
-            if (window.confirm(message)) {
```
rjmackay: This looses any confirmation before deleting a post??

server/www/templates/posts/post-editor.html
```diff
@@ -297,4 +297,13 @@ <h1 translate translate-values="{ type: activeForm.name, title: post.title }" ng
             </p>
         </div>
     </div>
+
+    <modal visible="userSavedPost">
+        <h1 translate>post.save_dlg_title</h1>
+        <div translate translate-values="{ 'post': post.title }">post.you_have_published</div>
+        <div class="modal-submit">
+            <button type="button" class="button-primary" ng-click="navigateToPost()" translate>post.see_post</button>
```
rjmackay: If you need a button to change URL, just switch it to a link `<a href="/posts/id" class="button button-primary"></a>

server/www/templates/posts/post-editor.html
```diff
@@ -297,4 +297,13 @@ <h1 translate translate-values="{ type: activeForm.name, title: post.title }" ng
             </p>
         </div>
     </div>
+
+    <modal visible="userSavedPost">
```
rjmackay: .. this should really be a confirmation message as per the pattern: http://github.ushahidi.org/platform-pattern-library/assets/html/4_blocks/#pl-pattern-confirmation

.. but this is better than alerts for launch

rjmackay:
Looks usable in the short term.. long term we need something more easily reusable (ie. where you can just call a function and get a promise that'll fire when confirmed or cancelled)
I think that probably looks like a single confirmation modal which is triggered from the notify service or something similar.

rjmackay:
@willdoran can you extend this to give us quota messages in time for launch?

willdoran:
I hope so!

app/post/controllers/post-detail-controller.js
hollycorbett: Nope, cause this function now only gets called when the deletion is confirmed in the modal.

rjmackay:
This already landed along with Will's changes .. sorry should have closed this earlier

PR #22, opened 2015-09-29
726 handle hitting quota

willdoran:
[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/22)


rjmackay:
Awesome!

When you save a post:
- let's just have it say "Your post "post_subject" was successfully saved!"
- We don't need to have the "view your post" link because the title of the post will be a link to it
- Also noticed the close button didn't work?
- How hard would it be to switch the save message to use the confirmation style too?

Delete confirmation:
- Almost right.. but theres an exact layout here: http://github.ushahidi.org/platform-pattern-library/assets/html/5_layouts/modal-delete.html
- Just need copy and markup changes.

For the quota message:
- Looks great
- Could we change the last sentence to "To get more simply [upgrade your plan](link)." With upgrade as a link.
- Keep the upgrade button too!

For all the confirmation messages: It should have a X to manually close it, but also close after being open for ~10 secs (we can tweak time once we see how this feels)

rjmackay:
Review status: 0 of 8 files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[server/www/templates/posts/post-editor.html, line 304 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/22#-K-QLVW0vibvqoIf9q35)** ([raw file](https://github.com/ushahidi/platform-client/blob/0224d74dc764b1953e526cdadad39db004abe13b/server/www/templates/posts/post-editor.html#L304)):</sup>
You can replace this with a link using a button class. Using ng-click to navigate is overkill
`<a class="button button-primary" href=""></a>


---

<sup>**[server/www/templates/posts/post-editor.html, line 308 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/22#-K-QLrno6i2R9DRoHCAq)** ([raw file](https://github.com/ushahidi/platform-client/blob/0224d74dc764b1953e526cdadad39db004abe13b/server/www/templates/posts/post-editor.html#L308)):</sup>
remove the 'Congratulations' title


---

<sup>**[server/www/templates/posts/post-editor.html, line 309 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/22#-K-QLvRREKWHG7B0VWPC)** ([raw file](https://github.com/ushahidi/platform-client/blob/0224d74dc764b1953e526cdadad39db004abe13b/server/www/templates/posts/post-editor.html#L309)):</sup>
this should be a `<p>` tag for correct styling 


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform-client/22)

PR #30, opened 2015-10-22
This changes is intended to prevent loading of pages breaking when a …

willdoran:
…sub element is not viewable but the main element is - in this case we ignore the errors for specific types - this is a short term solution and will be rectified under issue #793

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/30)

### DIFF
--- a/app/common/configs/authentication-interceptor.js
+++ b/app/common/configs/authentication-interceptor.js
@@ -55,8 +55,8 @@ function (
 
                 var deferred = $q.defer();
 
-                if (config.headers.ignorable) {
-                    delete config.headers.ignorable;
+                if (_.has(config, 'params') && config.params.ignore403) {
+                    delete config.params.ignore403;
                     config.ignorable = true;
                 }
 
--- a/app/common/configs/authentication-interceptor.js
+++ b/app/common/configs/authentication-interceptor.js
@@ -108,8 +108,22 @@ function (
                     }]);
                 // If its a 403 we've got a token, but it can't get us what we needed
                 } else if (rejection.status === 403) {
-                    // Trigger a forbidden event and show an error page
-                    $rootScope.$broadcast('event:forbidden');
+                    // In the short term, we will handle failures for 
+                    // associated entities - for example posts that contain unviewable tags -
+                    // we will ignore the error. In future this will be rectified under issue:
+                    // https://github.com/ushahidi/platform/issues/793
+                    var allowed_entities = ['tag', 'notifications'];
+                    var ignore_reject = false;
+                    _.find(allowed_entities, function(entity) {
+                        if (rejection.config.url.contains(entity)) {
+                            ignore_reject = true;
+                        }
+                    });
+
+                    if (!ignore_reject) {
+                        // Trigger a forbidden event and show an error page
+                        $rootScope.$broadcast('event:forbidden');
+                    }
                     deferred.reject(rejection);
                 // For anything else, just forward the rejection
                 } else {
--- a/app/common/configs/authentication-interceptor.js
+++ b/app/common/configs/authentication-interceptor.js
@@ -55,6 +55,11 @@ function (
 
                 var deferred = $q.defer();
 
+                if (config.headers.ignorable) {
+                    delete config.headers.ignorable;
+                    config.ignorable = true;
+                }
+
                 if (config.url.indexOf(CONST.API_URL) === -1) {
                     deferred.resolve(config);
                     return deferred.promise;
@@ -108,19 +113,11 @@ function (
                     }]);
                 // If its a 403 we've got a token, but it can't get us what we needed
                 } else if (rejection.status === 403) {
-                    // In the short term, we will handle failures for 
+                    // In the short term, we will handle failures for
                     // associated entities - for example posts that contain unviewable tags -
                     // we will ignore the error. In future this will be rectified under issue:
                     // https://github.com/ushahidi/platform/issues/793
-                    var allowed_entities = ['tag', 'notifications'];
-                    var ignore_reject = false;
-                    _.find(allowed_entities, function(entity) {
-                        if (rejection.config.url.contains(entity)) {
-                            ignore_reject = true;
-                        }
-                    });
-
-                    if (!ignore_reject) {
+                    if (!rejection.config.ignorable) {
                         // Trigger a forbidden event and show an error page
                         $rootScope.$broadcast('event:forbidden');
                     }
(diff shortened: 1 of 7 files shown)


willdoran:
@jasonmule This is the "fix" for the want of a better phrase for the notification and tag error. So the elements won't be displayed because they are not returned. Only their ID is seen by the client nothing more. 

Is there a more generic way I can capture which elements might cause this issue rather than the array ['tags','nofitications']

app/common/configs/authentication-interceptor.js
```diff
@@ -108,8 +108,22 @@ function (
...
                 } else if (rejection.status === 403) {
-                    // Trigger a forbidden event and show an error page
-                    $rootScope.$broadcast('event:forbidden');
+                    // In the short term, we will handle failures for 
+                    // associated entities - for example posts that contain unviewable tags -
+                    // we will ignore the error. In future this will be rectified under issue:
+                    // https://github.com/ushahidi/platform/issues/793
+                    var allowed_entities = ['tag', 'notifications'];
+                    var ignore_reject = false;
+                    _.find(allowed_entities, function(entity) {
```
rjmackay: I think you're just doing _.each here.. if you're not using the result then _.find doesn't make sense.
A combination of map and reduce might work too .. but each is fine.

app/common/configs/authentication-interceptor.js
```diff
@@ -108,8 +108,22 @@ function (
...
+                        if (rejection.config.url.contains(entity)) {
```
rjmackay: Not hugely fond of searching the URL. Could we pass a "magic" param instead somehow? ie. if you push something into the `config` of the $http request it should come back in `rejection.config` I think.. Might require some fiddling with $resource and $http to figure it out.. but feels safer.

rjmackay:
Think this breaks the interceptor tests too..

app/common/configs/authentication-interceptor.js
willdoran: Yep, I can change that.

app/common/configs/authentication-interceptor.js
willdoran: I'll try that

rjmackay:
Reviewed 4 of 4 files at r2.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---

<sup>**[app/common/services/endpoints/tag.js, line 28 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform-client/30#-K2FbD9NqZ26jSye5y_v)** ([raw file](https://github.com/ushahidi/platform-client/blob/70e25b1abe9d58bac27c4f81da6ac40923659d66/app/common/services/endpoints/tag.js#L28)):</sup>
I don't think this can be permanently true.. we need to set it to `false` when we're on the tags admin.. cos in that case a failure would be bad.


---

<sup>**[app/user-profile/services/endpoints/notification.js, line 17 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform-client/30#-K2Fb2-t1DPJBvmd87ai)** ([raw file](https://github.com/ushahidi/platform-client/blob/70e25b1abe9d58bac27c4f81da6ac40923659d66/app/user-profile/services/endpoints/notification.js#L17)):</sup>
Why are we ignoring notifications failure?


---

<sup>**[package.json, line 91 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform-client/30#-K2FbHxUQmBykdUqQGw_)** ([raw file](https://github.com/ushahidi/platform-client/blob/70e25b1abe9d58bac27c4f81da6ac40923659d66/package.json#L91)):</sup>
Please actually tag that commit, and use a tag name, not a commit hash.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform-client/30)

willdoran:
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks broke.

---

<sup>**[app/common/services/endpoints/tag.js, line 28 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform-client/30#-K2FbD9NqZ26jSye5y_v)** ([raw file](https://github.com/ushahidi/platform-client/blob/70e25b1abe9d58bac27c4f81da6ac40923659d66/app/common/services/endpoints/tag.js#L28)):</sup>
I'll make the value passable.


---

<sup>**[app/user-profile/services/endpoints/notification.js, line 17 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform-client/30#-K2Fb2-t1DPJBvmd87ai)** ([raw file](https://github.com/ushahidi/platform-client/blob/70e25b1abe9d58bac27c4f81da6ac40923659d66/app/user-profile/services/endpoints/notification.js#L17)):</sup>
Notification failures occurs on collections - as notifications can be associated with collections and have separate view controls from the collection itself.


---

<sup>**[package.json, line 91 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform-client/30#-K2FbHxUQmBykdUqQGw_)** ([raw file](https://github.com/ushahidi/platform-client/blob/70e25b1abe9d58bac27c4f81da6ac40923659d66/package.json#L91)):</sup>
OK will do.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform-client/30)

willdoran:
Review status: 0 of 7 files reviewed at latest revision, 3 unresolved discussions.

---

<sup>**[app/common/services/endpoints/tag.js, line 28 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform-client/30#-K2FbD9NqZ26jSye5y_v)** ([raw file](https://github.com/ushahidi/platform-client/blob/70e25b1abe9d58bac27c4f81da6ac40923659d66/app/common/services/endpoints/tag.js#L28)):</sup>
Done.


---

<sup>**[app/user-profile/services/endpoints/notification.js, line 17 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform-client/30#-K2Fb2-t1DPJBvmd87ai)** ([raw file](https://github.com/ushahidi/platform-client/blob/70e25b1abe9d58bac27c4f81da6ac40923659d66/app/user-profile/services/endpoints/notification.js#L17)):</sup>
Done.


---

<sup>**[package.json, line 91 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform-client/30#-K2FbHxUQmBykdUqQGw_)** ([raw file](https://github.com/ushahidi/platform-client/blob/70e25b1abe9d58bac27c4f81da6ac40923659d66/package.json#L91)):</sup>
 Done.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform-client/30)

rjmackay:
Nice :boom:

PR #32, opened 2015-10-26
#759 create post ignores map settings

jasonmule:
Fixes ushahidi/platform#759

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/32)

### DIFF
--- a/app/post/directives/post-location-directive.js
+++ b/app/post/directives/post-location-directive.js
@@ -34,26 +34,16 @@ function (
             var markers = {},
                 mapName = $scope.id + '-map';
 
-            // use default values in case there are none in the config
-            var center = {
-                lat: 36.079868,
-                lng: -79.819416,
-                zoom: 4
-            };
-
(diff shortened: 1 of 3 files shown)


app/post/directives/post-location-directive.js
```diff
@@ -29,15 +31,31 @@ function (
...
             var markers = {},
-                center = {
-                    lat: 36.079868,
-                    lng: -79.819416,
-                    zoom: 4
-                },
                 mapName = $scope.id + '-map';
 
+            // use default values in case there are none in the config
+            var center = {
```
rjmackay: I think this should be handled by Maps service?

rjmackay:
- Can we grab map layers from Maps service too? we should be pull most /all config from Maps ideally
- Can we disable scroll zoom on these maps too?

rjmackay:
@willdoran can you review?

PR #36, opened 2015-11-10
804 update dropdown pattern

willdoran:
Changing drop down menus to match platform pattern library

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/36)


server/www/templates/posts/post-view-filters.html
```diff
@@ -1,10 +1,9 @@
 <div class="filters" dropdown auto-close="disabled">
-    <div class="filters-show-hide" ng-class="{ half : halfWidth }">
+    <div class="filters-toggle init" ng-if="! hasAside" ng-class="{ half : halfWidth }" dropdown-toggle>
```
rjmackay: careful with ng-if .. it changes the scope of everything inside this element.

app/post/directives/post-view-filters-directive.js
```diff
@@ -12,6 +12,7 @@ function (
         replace: true,
         scope: {
             halfWidth: '=',
+            hasAside: '=',
```
rjmackay: Can you double check hasAside isn't just a dupe of halfWidth?

rjmackay:
Notice this doesn't change the classes used in the dropdown directive at all?? I thought the PL changes had change the names of the open/closed classes?

server/www/templates/posts/post-view-filters.html
willdoran: that's gross - I'll change it to a show!

app/post/directives/post-view-filters-directive.js
willdoran: Half width does appear to be redundant so I've removed it!

willdoran:
The main changes were order and structural. The class names themselves didn't change but they did move around a bit. The inclusion of the init and active elements for drop downs and then the changes to how filters are order hierarchically.

PR #46, opened 2015-11-20
fixing add user issue

willdoran:
[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/46)

### DIFF
--- a/app/setting/controllers/setting-users-controller.js
+++ b/app/setting/controllers/setting-users-controller.js
@@ -2,6 +2,7 @@ module.exports = [
     '$scope',
     '$translate',
     '$q',
+    '$location',
     '_',
     '$window',
     'Session',
@@ -12,6 +13,7 @@ function (
     $scope,
     $translate,
     $q,
+    $location,
     _,
     $window,
     Session,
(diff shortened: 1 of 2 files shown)


jasonmule:
Looks good.

---







_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform-client/46#-:-K3_wRgCFKbr7M3x-LiW:-1921560736)_

willdoran:
Fixed add user issue

app/setting/controllers/setting-users-controller.js
```diff
@@ -64,6 +66,10 @@ function (
         return false;
     };
 
+    $scope.addUser = function () {
+        $location.path('/settings/users/create');
```
rjmackay: We should really never be doing this. This is clearly a link, so use a link not a button and let browser navigation take car of it. If the CSS doesn't work as a button, thats a bug in the PL.

willdoran: I'll add it as an issue under the PL.

rjmackay: I *think* it actually already works. Using `class="button button-secondary"` on an `<a>` get the right visual

PR #57, opened 2015-12-18
Remove HTML from Post notification messages

jasonmule:
fixes #846

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/57)


rjmackay:
I'm not sure this is ideal, since now we have no link to a new post after it was created.

PR #60, opened 2016-01-11
Don't show registration and add post links for private deployments

jasonmule:
#859

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/60)

### DIFF
--- a/server/www/index.html
+++ b/server/www/index.html
@@ -69,7 +69,8 @@
                             <p class="header-description" ng-bind="site.description"></p>
 
                             <div class="create-admin">
-                                <a class="button button-primary" ng-href="/posts/create">
+                              <a class="button button-primary" ng-href="/posts/create"
+                                 ng-show="!site.private || site.private && loggedin">
                                     <span class="icon-left fa-plus" translate>nav.add_post</span>
                                 </a>
 
--- a/server/www/index.html
+++ b/server/www/index.html
@@ -70,7 +70,7 @@
 
                             <div class="create-admin">
                               <a class="button button-primary" ng-href="/posts/create"
-                                 ng-show="!site.private || site.private && loggedin">
+                                 ng-show="canCreatePost()">
                                     <span class="icon-left fa-plus" translate>nav.add_post</span>
                                 </a>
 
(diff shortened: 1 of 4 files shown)


server/www/index.html
```diff
@@ -69,7 +69,8 @@ <h1 class="beta" ng-bind="site.name"></h1>
                             <p class="header-description" ng-bind="site.description"></p>
 
                             <div class="create-admin">
-                                <a class="button button-primary" ng-href="/posts/create">
+                              <a class="button button-primary" ng-href="/posts/create"
+                                 ng-show="!site.private || site.private && loggedin">
```
rjmackay: Couldn't this just be if `loggedIn || !site.private` ? .. I'd be tempted to move this logic into a function in the controller. So you can just do `ngShow="canCreatePost()"`

server/www/templates/partials/user-menu.html
```diff
@@ -1,6 +1,6 @@
 <div class="user-admin logged-out" ng-hide="currentUser">
     <a ng-href="/login" class="user-login" translate>nav.login</a>
-    <a ng-href="/register" class="user-register" translate>nav.register</a>
+    <a ng-href="/register" class="user-register" translate ng-show="!site.private">nav.register</a>
```
rjmackay: Similar - maybe put this in a function so its `canRegister` .. more understandable then.

rjmackay:
Minor feedback in comments - but looks good otherwise :boom:

rjmackay:
Looks good so far.
I think you need to redirect any non-logged in users to /login though? (on private deployments)
We could do with some frontend tests too

---

Reviewed 3 of 3 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks broke.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform-client/60#-:-K7xAQJNwXtLBAFIpiiH:867836270)*

jasonmule:
I think redirects to /login already happen?

rjmackay:
Looks good.
You're right redirect to login is probably triggered when we attempt to load any content.

---

Reviewed 1 of 1 files at r4.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/ushahidi/platform-client/60#-:-KA7TwTtMabt2hlhJtpz:-738962332)*

willdoran:
This looks good to me, it even has a groovy spec test, which is going to conflict with the test I added yesterday :D 

My only question is whether it should merge into develop or master?

rjmackay:
Please merge to develop

PR #68, opened 2016-01-26
changed cardinaility setting for checkbox, changed attribue add metho…

willdoran:
…d to allow more variability, changed display of checkboxes on post edit to ensure displayed properly

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/68)


willdoran:
This requires a change on platform as well to allow cardinality to be set by the client.

PR #76, opened 2016-02-06
863 checkbox issue

willdoran:
[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/76)

### DIFF
--- a/server/www/templates/posts/post-editor.html
+++ b/server/www/templates/posts/post-editor.html
@@ -149,8 +149,22 @@
                       >
                           {{attribute.label}}
                       </label>
+                      <!-- type: checkbox -->
+                      <div class="form-field checkbox" ng-if="isCheckbox(attribute)" ng-repeat="option in attribute.options">
+                          <input 
+                              type="checkbox" 
+                              checklist-model="post.values[attribute.key]" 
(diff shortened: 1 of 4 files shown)


server/www/templates/posts/post-editor.html
```diff
@@ -149,8 +149,22 @@ <h1 class="page-title"><span class="icon-left fa-pencil" translate translate-val
...
+                              value="option"
+                              ng-class="{ 'error': form['values_' + attribute.key].$invalid && form['values_' + attribute.key].$dirty }">
+                         <label>
+                              {{option}} 
+                          </label>
+                      </div>
+
                       <div ng-repeat="(key, value) in post.values[attribute.key] track by key">
-                          <div ng-class="{ 'input-inline' : canRemoveValue(attribute, key) }">
+                          <div ng-if="!isCheckbox(attribute)">
```
rjmackay: Doesn't this completely lose the canRemoveValue() check and break multivalue fields (that aren't checkboxes)??

rjmackay: ... or maybe that doesn't matter since it seems we never exposed cardinality to the user

PR #78, opened 2016-02-09
Ui fixes

middle8media:
* removes `has-aside` from users template

* sets default select option on CSV import

* fixes alert error on login fail

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/78)


rjmackay:
Is this landed already @willdoran??

PR #81, opened 2016-02-13
Feature/code coverage - first iteration

willdoran:
[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/81)


willdoran:
I think the most effective way to evaluate these changes is to cause regression in the controllers that don't break the build but do break before and see if the tests catch the significant ones.

The tests need to be expanded further, this iteration includes an example directive test, as well as fixture loading. This can be used to increase directive coverage. Which should happen in the next iteration.

With another week I believe I can get overall client coverage to above 80%.

PR #92, opened 2016-02-18
Declare dependencies explicitly for data-mapper/ route resolve function.

rjmackay:
We use strict DI because otherwise dependencies break when compressing JS

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/92)

### DIFF
--- a/app/setting/setting-routes.js
+++ b/app/setting/setting-routes.js
@@ -24,12 +24,12 @@ function (
         controller: require('./controllers/setting-data-mapper-controller.js'),
         templateUrl: 'templates/settings/data-import/data-mapper.html',
         resolve: {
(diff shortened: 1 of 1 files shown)


app/setting/setting-routes.js
```diff
@@ -24,12 +24,12 @@ function (
         controller: require('./controllers/setting-data-mapper-controller.js'),
         templateUrl: 'templates/settings/data-import/data-mapper.html',
         resolve: {
-            initialData: function ($route, DataRetriever) {
+            initialData: ['$route', 'DataRetriever', function ($route, DataRetriever) {
```
rjmackay: @willdoran FYI - this is needed or things break when we compress code.

PR #119, opened 2016-03-08
762 language settings

jasonmule:
- Loads translated languages from a JSON file downloaded from Transifex and falls back to English (US) if the file is not available.

Test these changes by:
- run `gulp transifex-download`
- Log in as admin, go to General settings and check that the languages available are the ones with over 90% translated strings. Currently, 6 languages have over 90% of their strings translated.

Fixes ushahidi/platform#762 .

Ping @ushahidi/platform

[Review on Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/119)


willdoran:
This looks good, the download works.

When I save the change I have to refresh the page in order to see the language update. Can we force the language update to occur after the save?

How many languages should there be? I see 6, English(3), Swahili, Indonesian, Chinese(Taiwan)

jasonmule:
Currently there are 6 languages with over 90% of their strings translated.

willdoran:
Awesome, so I'm seeing the right number!

gulpfile.js
```diff
@@ -364,6 +364,8 @@ gulp.task('transifex-download', function () {
         locales_dir = options.www + '/locales/',
         mode = 'default',
         resource = 'client-en',
+        // Get languages that are at least 90% translated
+        completion_threshold = 90,
```
rjmackay: Love this!

rjmackay:
This is great.  I'm tempted to take anything over 50-70% at least until we have a few more languages.

rjmackay:
I'm going to merge this. Can you fix the language refresh in a follow up PR? (ie. switch language immediately, or just force a page reload)

PR #128, opened 2016-03-16
Feature/intercom

willdoran:
- Expanded information sent to intercom
- Set intercom only to load when user logged in
- Set intercom to unload on user log out

Test these changes by:
- Create a new deployment
- Log in, you should see an intercom welcome message
- Check intercom interface, you should see your new user, and the user's information - name/role etc should be correct

Fixes ushahidi/platform# .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/128)

### DIFF
--- a/app/common/common-module.js
+++ b/app/common/common-module.js
@@ -41,6 +41,7 @@ angular.module('ushahidi.common', [
 .controller('navigation', require('./controllers/navigation.js'))
 .controller('PageMetadata', require('./controllers/page-metadata.js'))
 .controller('notifier', require('./controllers/notifier.js'))
+.controller('intercom', require('./controllers/intercom.js'))
 
 .directive('collectionSelector', require('./directives/collection-selector.js'))
 .directive('roleSelector', require('./directives/role-selector.js'))
--- a/app/common/controllers/intercom.js
+++ b/app/common/controllers/intercom.js
@@ -2,22 +2,33 @@ module.exports = [
     '$scope',
     '$rootScope',
     'UserEndpoint',
+    'ConfigEndpoint',
+    '$q',
 function (
     $scope,
     $rootScope,
-    UserEndpoint
+    UserEndpoint,
+    ConfigEndpoint,
+    $q
 ) {
     $rootScope.$on('event:authentication:login:succeeded', function () {
-        UserEndpoint.getFresh({id: 'me'}).$promise.then(function (user) {
+        $q.all([
+            ConfigEndpoint.get({ id: 'site' }).$promise,
+            UserEndpoint.getFresh({id: 'me'}).$promise
+        ]).then(function (results) {
+            var site = results[0];
+            var user = results[1];
+
             window.Intercom('boot', {
                 app_id: window.ushahidi.intercomAppId,
                 email: user.email,
                 created_at: user.created,
                 user_id: user.id,
+                'company': site.name,
+                'deployment_url': window.ushahidi.apiUrl,
                 'realname' : user.realname,
                 'last_login': user.last_login,
-                'role': user.role,
-                'logins': user.logins
+                'role': user.role
             });
         });
     });
--- a/app/common/controllers/intercom.js
+++ b/app/common/controllers/intercom.js
@@ -23,7 +23,7 @@ function (
                 app_id: window.ushahidi.intercomAppId,
                 email: user.email,
                 created_at: user.created,
-                user_id: user.id,
+                user_id: window.ushahidi.apiUrl + '_' + user.id,
                 'company': site.name,
                 'deployment_url': window.ushahidi.apiUrl,
                 'realname' : user.realname,
--- a/app/common/controllers/intercom.js
+++ b/app/common/controllers/intercom.js
@@ -11,6 +11,8 @@ function (
     ConfigEndpoint,
     $q
 ) {
+    var pattern = /^(?:https?:\/\/)?(?:[^@\/\n]+@)?(?:www\.)?([^:\/\n]+)/g;
+
     $rootScope.$on('event:authentication:login:succeeded', function () {
         $q.all([
             ConfigEndpoint.get({ id: 'site' }).$promise,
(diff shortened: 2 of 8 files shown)


server/www/index.html
```diff
@@ -141,6 +141,12 @@ <h2 ng-bind="site.name"></h2>
                 }(document, 'script'));
             }
         </script>
+        <!--
+        Intercom integration
+        -->
+        <script src="//static.intercomcdn.com/intercom.v1.js"></script>
```
willdoran: This should be stored locally.

app/common/controllers/intercom.js
```diff
@@ -0,0 +1,28 @@
...
+    $rootScope,
+    UserEndpoint
+) {
+    $rootScope.$on('event:authentication:login:succeeded', function () {
+        UserEndpoint.getFresh({id: 'me'}).$promise.then(function (user) {
+            window.Intercom('boot', {
+                app_id: window.ushahidi.intercomAppId,
+                email: user.email,
+                created_at: user.created,
+                user_id: user.id,
```
willdoran: Probably not needed

app/common/controllers/intercom.js
```diff
@@ -0,0 +1,28 @@
...
+                'realname' : user.realname,
+                'last_login': user.last_login,
+                'role': user.role,
+                'logins': user.logins
```
willdoran: what other elements do we want to capture?

app/common/controllers/intercom.js
rjmackay: Seems like a good idea though

app/common/controllers/intercom.js
rjmackay: - I don't think user.logins or last_login are actually incremented on the backend???
- Can we attach users to their deployment? I think intercom has 'companies' I think it'd make sense to create a company per deployment

app/common/controllers/intercom.js
willdoran: We should have the domain/subdomain for the user so we know what deployment they come from.

app/common/common-module.js
```diff
@@ -41,6 +41,7 @@ angular.module('ushahidi.common', [
 .controller('navigation', require('./controllers/navigation.js'))
 .controller('PageMetadata', require('./controllers/page-metadata.js'))
 .controller('notifier', require('./controllers/notifier.js'))
+.controller('intercom', require('./controllers/intercom.js'))
```
rjmackay: So I'd be tempted to take this out of Angular entirely - if someone loads a deployment and hits a major error then angular might never bootstrap.. 
If we load intercom outside angular - then they at least get a support channel directly on the failure page. Does that make sense?

app/common/controllers/intercom.js
willdoran: Yeah, I think we can assign directly to that field. Should we make a list of what should be captured in its own issue?

server/www/index.html
rjmackay: I'd like to do something similar here to the `window.ushahidi.gaEnabled` check below .. ie. don't load the JS if its not enabled. Otherwise we're loading intercom into OSS deployments where its not used.

app/common/controllers/intercom.js
```diff
@@ -0,0 +1,28 @@
+module.exports = [
+    '$scope',
+    '$rootScope',
+    'UserEndpoint',
+function (
+    $scope,
+    $rootScope,
+    UserEndpoint
+) {
+    $rootScope.$on('event:authentication:login:succeeded', function () {
```
rjmackay: Can we trigger some kind of 'event' in intercom to record the login/logout for each user?

app/common/common-module.js
willdoran: Yeah that does make sense. BUT at the moment it's based on login, if the user can't login - if the app fails completely - then we have no way to identify them.

Would it make sense to have an outer intercom that kicks in on failed load that sits outside of angular?

app/common/controllers/intercom.js
willdoran: So we can do anything that is here:
https://docs.intercom.io/configuring-intercom/the-intercom-javascript-api

Intercom does, usefully, record login/out automatically based on boot/shutdown that info will appear in the intercom interface so we get that upfront.

server/www/index.html
willdoran: Yeah that'd be nice, the only issue is that dynamically loading that kind of JS in angular seems messy but may be someone has a nice wrapper for it.

rjmackay:
I notice we're not showing intercom at all for non logged in users. Could we...
- Let non-logged in users open intercom from a 'support' link or something similar
- Swap the link on our 'report a bug' button on cloud to open intercom instead of loading github?

@jshorland would those changes make sense?

willdoran:
That's really easy to do AND it will handle the case of the app failing to load!

app/common/controllers/intercom.js
rjmackay: Most of the tracking can be a future issue. But ideally the user -> deployment link would go in now.. not sure how much work is involved though?

app/common/common-module.js
rjmackay: Can we just load intercom outside angular (in a global) but then have angular send the user details on login/logout?

app/common/controllers/intercom.js
rjmackay: looks like intercom already does some 'Session' tracking itself https://docs.intercom.io/faqs/how-is-a-session-defined

rjmackay:
- Do we need to trigger intercom `update` when the user edits their profile??
- We can do `Intercom('trackEvent', 'invited-friend');` ..  we can make a list of things to track on a future issue.

willdoran:
@rjmackay 
Theoretically, yes, but I think what we need to do is define a somewhat unique id, so it would seem logical to restrict it to something unchanging, depoyment_domain + user_id. If they change email address or any of the other data the above two should remain consistent. So intercom appears to use user_id and/or email to identify individual users.

I think there may be a conflict between users where the id overlaps as at the moment the user_id is just the straight platform user_id.

Does deployment_name + user_id seem logical to create a unique ID? Or is there a better construction?

rjmackay:
> Does deployment_name + user_id seem logical to create a unique ID? Or is there a better construction?

Deployment subdomain + user id works.. maybe include the ushahidi.io portion too? That'll future proof for include OSS deploys. There will be a minor issue if we change deployment URLs.. but thats a special case I think.

jshorland:
Yes, those changes make sense. Non-logged in users might be deemed “visitors” in intercom. If that’s the case, we should at least know which deployment they are on. 

—Jess


Jess Shorland
Product manager
jess@ushahidi.com

willdoran:
Ok, i've added some issues covering this:

[131](https://github.com/ushahidi/platform-client/issues/131)
[132](https://github.com/ushahidi/platform-client/issues/132)
[133](https://github.com/ushahidi/platform-client/issues/133)

Issue #138, opened 2016-03-22
Map view shown as unavailable on first load


rjmackay:
### Expected behaviour
* Map view should appear on first load

### Actual behaviour
* "The map view isn't available on your current plan. Upgrade your plan to get it!"

### Steps to reproduce the behaviour/error
* Clear localStorage (run `localStorage.clear()` in console)
* Reload the home page

This is caused by a race condition where the feature config hasn't loaded yet when the unavailable view is rendered.

PR #139, opened 2016-03-22
Fix race condition on first map load

rjmackay:
- Fix race condition on first map load. Don't run view available check till features config is loaded

Test these changes by:
- Clear localStorage (run localStorage.clear() in console)
- Reload the home page

Fixes ushahidi/platform-client#138 .

Ping @willdoran

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/139)

### DIFF
--- a/app/post/directives/views/post-view-directive.js
+++ b/app/post/directives/views/post-view-directive.js
@@ -20,17 +20,20 @@ function (
                 $scope.currentView = 'map';
             }
 
-            if (!PostViewHelper.isViewAvailable($scope.currentView)) {
-                $scope.unavailableView = $scope.currentView;
-                $scope.currentView = 'unavailable';
-            }
+            // Watch views, in case we get new feature config
+            PostViewHelper.isViewAvailable($scope.currentView).then(function (available) {
+                if (!available) {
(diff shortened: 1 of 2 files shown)


app/post/directives/views/post-view-directive.js
```diff
@@ -20,17 +20,20 @@ function (
                 $scope.currentView = 'map';
             }
 
-            if (!PostViewHelper.isViewAvailable($scope.currentView)) {
-                $scope.unavailableView = $scope.currentView;
-                $scope.currentView = 'unavailable';
-            }
+            // Watch views, in case we get new feature config
+            PostViewHelper.isViewAvailable($scope.currentView).then(function (available) {
+                if (!available) {
```
willdoran: Could this all be:

Features.loadFeatures().then(function (features) {
           if(features.isViewEnabled(view) {....

Do we want to keep the PostViewHelper or could it ultimately be replaced with an improved version of features? Or is the separation of the two better?

rjmackay: I don't have strong opinions either way. This is the only place we use isViewAvailable right now..

PR #142, opened 2016-03-22
fixed limit messages

willdoran:
- Improves limit message
- Improves placement of post type limit message
- Fixes hiding of datasource config when not available
Test these changes by:
- Ensure that limit message appear correctly when appropriate limits set on api

Fixes ushahidi/platform# .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/142)


rjmackay:
> Fixes hiding of datasource config when not available

I'm not clear that that was a bug? shouldn't it be hidden?

PR #157, opened 2016-04-04
Google 2fa enable in client

willdoran:
- Added endpoints for 2fa enable verify
- Added module and template to handle 2fa enable
## Test these changes by:

Fixes ushahidi/platform#980

Ping @ushahidi/platform

---

This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/157)


rjmackay:
Reviewed 2 of 6 files at r1, 3 of 4 files at r3, 3 of 5 files at r4, 6 of 6 files at r5.
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[app/common/directives/google2fa.js, line 27 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/157#-KGjQCKJe7WKyMPHX1iA:-KGjQCKJe7WKyMPHX1iB:652762766) ([raw file](https://github.com/ushahidi/platform-client/blob/786218dcf01bb943789006b2f4c6eff9379a4722/app/common/directives/google2fa.js#L27)):*
its weird to have a 'notX' variable.. could this be named better? ie. is it really 'awaitingVerification'? where its false if 2fa is disabled, and also if its fully enabled+verified.

---

*[app/common/directives/google2fa.js, line 33 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/157#-KGjNfJc30byemLeyunE:-KGjNfJc30byemLeyunF:-730233575) ([raw file](https://github.com/ushahidi/platform-client/blob/786218dcf01bb943789006b2f4c6eff9379a4722/app/common/directives/google2fa.js#L33)):*
I was thinking handling multiple 'steps' that suggests this should really be multiple directives and routes too.. but since this is in a modal I'm not sure that works.

Nothing to change now. Just something to think on for general architecture.

---

*[app/common/directives/google2fa.js, line 50 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/157#-KGjPk-tkJi_icRuSg76:-KGjPk-tkJi_icRuSg77:265410711) ([raw file](https://github.com/ushahidi/platform-client/blob/786218dcf01bb943789006b2f4c6eff9379a4722/app/common/directives/google2fa.js#L50)):*
Why do we have to update the user AND disable 2fa? 

---

*[app/common/directives/google2fa.js, line 86 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/157#-KGjQZF-zPbRedlAeqBl:-KGjQZF-zPbRedlAeqBm:-2103109718) ([raw file](https://github.com/ushahidi/platform-client/blob/786218dcf01bb943789006b2f4c6eff9379a4722/app/common/directives/google2fa.js#L86)):*
Why is this a separate API call? it seems to me that verifying the 2fa code should also trigger the actual enabling of 2fa on the user too..

---

*[app/common/directives/google2fa.js, line 104 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/157#-KGjR1fUrPQKx1ZW2_c-:-KGjR1fUrPQKx1ZW2_c0:-665869013) ([raw file](https://github.com/ushahidi/platform-client/blob/786218dcf01bb943789006b2f4c6eff9379a4722/app/common/directives/google2fa.js#L104)):*
Whats the difference between verificationFailed and running the error callback below? .. seems odd that failed verification would still be a 200 response

---

*[server/www/templates/login.html, line 6 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/157#-KGjOEa2KqNc1HgGlMmx:-KGjOEa2KqNc1HgGlMmy:-1696663621) ([raw file](https://github.com/ushahidi/platform-client/blob/786218dcf01bb943789006b2f4c6eff9379a4722/server/www/templates/login.html#L6)):*
Is it possible to get to the otp step with an incorrect password?
Ideally it shouldn't be.. but if it is we need to show the password fields still.

---

*[server/www/templates/login.html, line 24 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/157#-KGjOGyWMK4x9MPGZMTY:-KGjOGyWMK4x9MPGZMTZ:-112039030) ([raw file](https://github.com/ushahidi/platform-client/blob/786218dcf01bb943789006b2f4c6eff9379a4722/server/www/templates/login.html#L24)):*
Should this just be a 2nd route? so that you just redirect /login to /login/otp ? or does that make it harder to maintain state?

---

*[server/www/templates/settings/google-2fa.html, line 34 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/157#-KGjOoUBqelZIIonO1KF:-KGjOoUBqelZIIonO1KG:1635162585) ([raw file](https://github.com/ushahidi/platform-client/blob/786218dcf01bb943789006b2f4c6eff9379a4722/server/www/templates/settings/google-2fa.html#L34)):*
ng-switch might make more sense for the step switching:
```
<ANY ng-switch="step">
  <ANY ng-switch-when="verified">...</ANY>
  <ANY ng-switch-when="qr_verify">...</ANY>
  <ANY ng-switch-default>...</ANY>
</ANY>
```

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ushahidi/platform-client/157)*

willdoran:
Review status: all files reviewed at latest revision, 8 unresolved discussions, some commit checks failed.

---

*[app/common/directives/google2fa.js, line 27 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/157#-KGjQCKJe7WKyMPHX1iA:-KGmDF60QjjB2Nkf5Dq4:-499146009) ([raw file](https://github.com/ushahidi/platform-client/blob/786218dcf01bb943789006b2f4c6eff9379a4722/app/common/directives/google2fa.js#L27)):*
Done.

---

*[app/common/directives/google2fa.js, line 33 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/157#-KGjNfJc30byemLeyunE:-KGmBNXxgq98EcrpnTqQ:1805143107) ([raw file](https://github.com/ushahidi/platform-client/blob/786218dcf01bb943789006b2f4c6eff9379a4722/app/common/directives/google2fa.js#L33)):*
Yeah, I wanted individual directives per step akin to the way data-import works but within a modal it was non-obvious. 

What we could do is to have the modal behave more like the modify template - where it switches between choose-form and post-editor. That would still mean there is essentially a parent directive that breaks apart into the individual steps.

If we went in that kind of direction it would need to be supported by a state system that makes it easier to maintain state across the steps.

---

*[app/common/directives/google2fa.js, line 50 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/157#-KGjPk-tkJi_icRuSg76:-KGm27GNJR8ha8mGZClP:1981022347) ([raw file](https://github.com/ushahidi/platform-client/blob/786218dcf01bb943789006b2f4c6eff9379a4722/app/common/directives/google2fa.js#L50)):*
The enable/disable/verify are managed by a separate route on the api. So what we could do is trigger a user update via the disable/enable route, the reason I didn't want to do this was because it becomes oblique that a user update is occurring. It's almost like a side effect. 

May be we could chat later about this in terms of what a good pattern would be on platform. 

---

*[app/common/directives/google2fa.js, line 104 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/157#-KGjR1fUrPQKx1ZW2_c-:-KGmZpNE9KDYPYXfdCRI:1916713156) ([raw file](https://github.com/ushahidi/platform-client/blob/786218dcf01bb943789006b2f4c6eff9379a4722/app/common/directives/google2fa.js#L104)):*
The verification step in this case is essentially a remote function call, so it wants the return value of whether or not the code is valid. I think in that sense the verify returning false is not an error, it's the result. Whereas the error condition should handle errors on the server side.

---

*[server/www/templates/login.html, line 6 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/157#-KGjOEa2KqNc1HgGlMmx:-KGm8wiXxy8uKhxmfd6d:-1063232040) ([raw file](https://github.com/ushahidi/platform-client/blob/786218dcf01bb943789006b2f4c6eff9379a4722/server/www/templates/login.html#L6)):*
It is possible to end up in that state, I will change the ordering so that the password is checked first and bounces the user back if it fails, only progressing to otp if it passes.

That actually makes it much nicer, I don't know why I did it the other way around :/

---

*[server/www/templates/login.html, line 24 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/157#-KGjOGyWMK4x9MPGZMTY:-KGmAcZWJuR_4nBing1v:-644136209) ([raw file](https://github.com/ushahidi/platform-client/blob/786218dcf01bb943789006b2f4c6eff9379a4722/server/www/templates/login.html#L24)):*
I decided to keep them together to simplify maintaining state, because the user credentials will need to be resent.

It might be splitable into a separate route but I think what we should do is create an abstraction of login type with special types, login, login-otp, that would mean we could ultimately add other login types. I wanted to avoid building that service upfront.

---

*[server/www/templates/settings/google-2fa.html, line 34 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/157#-KGjOoUBqelZIIonO1KF:-KGm5yUn2lm3DuGKsGma:-499146009) ([raw file](https://github.com/ushahidi/platform-client/blob/786218dcf01bb943789006b2f4c6eff9379a4722/server/www/templates/settings/google-2fa.html#L34)):*
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ushahidi/platform-client/157)*

rjmackay:
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[app/common/directives/google2fa.js, line 50 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/157#-KGjPk-tkJi_icRuSg76:-KGnXzA5Fk2RAzf5rHCH:-1414663921) ([raw file](https://github.com/ushahidi/platform-client/blob/786218dcf01bb943789006b2f4c6eff9379a4722/app/common/directives/google2fa.js#L50)):*
From our call, I think ideally this user update should happen on the backend - it is a side effect but also a concern the client shouldn't have to manage.

However if it takes significant effort to make that happen on the API then the current approach is fine.

---

*[app/common/directives/google2fa.js, line 104 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/157#-KGjR1fUrPQKx1ZW2_c-:-KGnXji3-ncqzmnwKys7:-1029323965) ([raw file](https://github.com/ushahidi/platform-client/blob/786218dcf01bb943789006b2f4c6eff9379a4722/app/common/directives/google2fa.js#L104)):*
Elsewhere in the API we handle this as the difference between a validation error (HTTP 422) vs a server error (HTTP 500) .. I think thats still appropriate here.

---

*[server/www/templates/login.html, line 6 \[r5\]](https://reviewable.io:443/reviews/ushahidi/platform-client/157#-KGjOEa2KqNc1HgGlMmx:-KGnYHpcPNqaQr_9Fcc7:951019167) ([raw file](https://github.com/ushahidi/platform-client/blob/786218dcf01bb943789006b2f4c6eff9379a4722/server/www/templates/login.html#L6)):*
Excellent. Password first matches common OTP flows in other apps (gmail, github etc)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ushahidi/platform-client/157)*

PR #159, opened 2016-04-05
Generate post title from content when post is title-less

rjmackay:
- `<h3>{{post.title}}</h3>`

to 

- `<h3>{{ post.title || post.content | limitTo:50 }}{{ post.content.length > 50 ? ' ...' : ''}}</h3>`

Test these changes by:
- Add a new post without a title
- Post description is required
- Post title should default to post content and should link to post detail

Fixes ushahidi/platform-pattern-library#23

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/158)


server/www/templates/posts/preview.html
```diff
@@ -18,7 +18,7 @@
         </div>
         <div class="post-text">
             <a href="/posts/{{post.id}}">
-                <h3>{{post.title}}</h3>
+                <h3>{{ post.title || post.content | limitTo:50 }}{{ post.content.length > 50 ? ' ...' : ''}}</h3>
```
rjmackay: @middle8media post title can be up to 150 chars. `post.content.length > 50 ? ' ...' : ''` will only show an ellipsis if we're using content. Could you do `post.content.length > 50 || post.title.length > 50 ? ' ...' : ''` instead?

middle8media: Hey @rjmackay,

> post.content.length > 50 ? ' ...' : '' will only show an ellipsis if we're using content

I used the elipsis on the title generated from the content because a mid-word or mid-sentence break felt harsh. The elipsis is not needed when a post has a title. I will change the 50 to 150, then if a post has a title, it will display the whole title, if a post does not have a title, then a title will be generated from the content and will use an ellipsis as needed, depending on character length.

Cool?

PR #164, opened 2016-04-13
Processing ui feedback for post location directive

ihatehandles:
This pull request makes the following changes:
- Modifies the Post Location directive to disable inputs and show a spinning cog when a lookup is running

Test these changes by:
- Running a location search on the post creation route 

Fixes ushahidi/platform-client#150 .

Ping @ushahidi/platform

Will bring these in one at a time to make it easier to correct mistakes since I'm still yet to be comfortable with the Ushahidi code and conventions. Should this also come with tests or it qualifies for the "non-critical path" badge?

Apologizing in advance for any gotchas :)

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/164)

PR #168, opened 2016-04-27
(Some) Tests for the Post Location directive

ihatehandles:
- Adds tests for the Post Location  directive

Test these changes by:
- Running the tests

Fixes ushahidi/platform# .

Ping @ushahidi/platform

This is more for the process of learning your approach to writing tests for the platform client, since this varies from team to team. If it's good I'll add tests for the map events `click` & `locationfound`, and also for the additions in PR #164

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/168)

### DIFF
--- a/test/unit/post/directives/post-location-directive-spec.js
+++ b/test/unit/post/directives/post-location-directive-spec.js
@@ -7,7 +7,8 @@ describe('post location directive', function () {
         isolateScope,
         GlobalFilter,
         Notify,
-        element;
+        element,
+        Geocoding;
 
     beforeEach(function () {
         fixture.setBase('mocked_backend/api/v3');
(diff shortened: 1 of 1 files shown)


willdoran:
This looks really great. The only suggestion I have is to move the mock functions out to a mock service under test/mock/services/

---

Review status: 0 of 1 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[test/unit/post/directives/post-location-directive-spec.js, line 49 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform-client/168#-KGm3CIOjuwa6nVNlI5i:-KGm3CIOjuwa6nVNlI5j:683713083) ([raw file](https://github.com/ushahidi/platform-client/blob/01101a043f035249aebec6fca7e89e32f3eddd8d/test/unit/post/directives/post-location-directive-spec.js#L49)):*
Could you move these mock values into a Leaflet mock in the test/mock/services? As much as possible it would be good to keep any mock jazz outside of the tests.

---

*[test/unit/post/directives/post-location-directive-spec.js, line 85 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform-client/168#-KGm3YClC01nZScyCqxj:-KGm3YCmanAWL4J7cAu5:-1200966096) ([raw file](https://github.com/ushahidi/platform-client/blob/01101a043f035249aebec6fca7e89e32f3eddd8d/test/unit/post/directives/post-location-directive-spec.js#L85)):*
This behaviour could also be moved out to a mock Leaflet service.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ushahidi/platform-client/168#-:-KGm3pjNEW2ABxDUjvkV:-1172963793)*

ihatehandles:
Noted. Yes I noticed all the mocks exist in one folder but wasn't sure what the approval process would be for a service mock to get a permanent seat there. I'll move them.

So for the mock themselves, are they structured how you prefer for this project or I was supposed to use $q for example to mock the promises. I mean do you have any policies regarding how to create the mocks themselves or as long as it's clean and clear it'll do?

Thanks btw for the feedback

willdoran:
@ihatehandles, great, there is no solid pattern right now - the services should really be split up more into sub dirs, there is a third party dir and there is an example leaflet mock already which you could use:
https://github.com/ushahidi/platform-client/blob/feature/google-2fa/test/unit/mock/services/third_party/leaflet.js

The services are evolvable so if you see a better pattern or better structuring do feel free to adjust them and we can review. The more you can decompose stuff the better.

Let me know if you have any issues and I'll help out!

ihatehandles:
Thanks, will reach out if I get stuck. Let me take a stab now and see how I fare

rjmackay:
@ihatehandles anything you need on this?

ihatehandles:
Nope. Forgot about this (been crammed). I'll try and wrap this up this week.Trust you're great, some awesome things you've added recently.

rjmackay:
Merged this to develop as-is

PR #170, opened 2016-04-28
Fixed post status filter not appearing despite user having permissions

willdoran:
- Fixes post status filter which was not enabled even when user had correct permissions

Test these changes by:
- On the post views screen, it should be possible to select status as a filter option, when a user either is an admin or has the manage Posts permission.

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/170)

### DIFF
--- a/app/post/directives/post-choose-form-directive.js
+++ b/app/post/directives/post-choose-form-directive.js
@@ -15,7 +15,7 @@ function (
         },
         templateUrl: 'templates/posts/choose-form.html',
         link: function ($scope) {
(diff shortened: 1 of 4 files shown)


rjmackay:
c/p from hipchat:
any reason not to just do "$scope.hasPermission = $rootScope.hasPermission" ..
it seems confusing to have hasPermission() take a param in some places but not others

rjmackay:
@willdoran Can you redo this PR against master not develop?
Then just merge it yourself.

willdoran:
Yep will do

PR #173, opened 2016-05-01
Feature/csv configure

willdoran:
- Added configure screen for bulk data import
- Split Post Editor directive into sub directives covering post status and post stages
- Added PostEditService to handle common functionality for post editing
- Added additional field fixed values for bulk data import, including all Form fields, tags, title and content
- Added Bulk Data Import mode for Post Editor
- Added switch to handle properties that should/shouldn't be enabled in Post Editor during Bulk Data Import mode

Test these changes by:
- Attempt to import a new csv
- Map the fields
- Set default values for all imported posts via the configure screen
- Check that the default values are correctly set for the newly imported Posts

Fixes ushahidi/platform# 945

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/173)


willdoran:
@rjmackay @jasonmule could you guys give this a review? It looks more involved then it actually is - I will have to merge Jason's changes from the media upload back into this branch once they land in develop.

The most important changes in this PR is the addition of the configure post step to csv. To facilitate this, I broke apart post-editor into: 

- Post-value-edit: this manages the individual attribute field to appear on the post editor
- Post-status: this directive manages the status field - it can be used to replace the duplicate code that exists anywhere post-status currently exists
- Post-stages: this directive manages all the stage related functionality, it consumes post-value edit and is consumed by the post-editor directive
- Post Edit service: this service holds common post edit code which have been split out as they were copied in a number of places
- Post editor: I've reduced the code to a smaller footprint, so post editor is no responsible solely for data load and CRUD. Ultimately, the data load should move further out - ideally onto the route itself.
- Post editor bulk data import mode: There are some specific difference between the way post editor works on the post edit view and how it behaves on configure csv. In the latter, different data needs to be passed in a different fields need to be shown. For example, post title is always present on the post edit page but may be not be present on the configure view.

Pulling out the code into separate directives makes it easier to see what code is there and what it does, hopefully it will make further cleaning it more straightforward.

PR #190, opened 2016-05-19
Mustang/filter system rebase

rjmackay:
- Adds filter system
- Adds implementation of filtering functionality for users

Test these changes by:
- Via the users page attempt to filter the list by search term and/or roles

Fixes ushahidi/platform#1048

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/190)


rjmackay:
Added feedback. Maybe we should pair on this for a bit tomorrow?

<details><summary><i>Previously, rjmackay (Robbie Mackay) wrote…</i></summary>
> #### Mustang/filter system rebase
> 
> This pull request makes the following changes:
> 
> - Adds filter system
> 
> - Adds implementation of filtering functionality for users
> 
> 
> 
> Test these changes by:
> 
> - Via the users page attempt to filter the list by search term and/or roles
> 
> 
> 
> Fixes ushahidi/platform#1048
> 
> 
> 
> Ping @ushahidi/platform
> 
> 
> 
</details>


---

Reviewed 15 of 15 files at r1.
Review status: all files reviewed at latest revision, 12 unresolved discussions, some commit checks failed.

---

*[app/common/common-module.js, line 55 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform-client/190#-KI62pnpINqNSEee2hbU:-KI62pnqS1UI3UT28A7v:1006864635) ([raw file](https://github.com/ushahidi/platform-client/blob/5f2617e9a79f0ea5230e5e0674158f5f850fdb67/app/common/common-module.js#L55)):*
> ```JavaScript
> 
> .directive('filterSystem', require('./directives/filter-system/filter-system.js'))
> .directive('filterUsers', require('./directives/filter-system/filter-users.js'))
> ```

if we're creating filters for categories/users/posts as individual directives I think they should live where they're used: ie. categories and users move to settings module.

---

*[app/common/directives/filter-system/filter-system.js, line 25 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform-client/190#-KI69-l6tiS9NttSBvQS:-KI69-l6tiS9NttSBvQT:2135449273) ([raw file](https://github.com/ushahidi/platform-client/blob/5f2617e9a79f0ea5230e5e0674158f5f850fdb67/app/common/directives/filter-system/filter-system.js#L25)):*
> ```JavaScript
>             _
>         ) {
>             $scope.searchResultsActive = false;
> ```

What is this for?

---

*[app/common/directives/filter-system/filter-system.js, line 39 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform-client/190#-KI66v8coLciE2cr3H7S:-KI66v8coLciE2cr3H7T:1305806324) ([raw file](https://github.com/ushahidi/platform-client/blob/5f2617e9a79f0ea5230e5e0674158f5f850fdb67/app/common/directives/filter-system/filter-system.js#L39)):*
> ```JavaScript
>         }];
>     return {
>         restrict: 'E',
> ```

I suspect we'd be better with a top level filters directive that just applies to an existing form element but doesn't replace it
`<form filter-form filters="foo">`

we could then watch for submit events but otherwise leave the form markup loose

---

*[app/common/directives/filter-system/filter-system.js, line 42 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform-client/190#-KI63KEUASOqjJPwZykq:-KI63KEUASOqjJPwZykr:483377071) ([raw file](https://github.com/ushahidi/platform-client/blob/5f2617e9a79f0ea5230e5e0674158f5f850fdb67/app/common/directives/filter-system/filter-system.js#L42)):*
> ```JavaScript
>         replace: true,
>         templateUrl: 'templates/common/filter-system/filter-system.html',
>         scope: {
> ```

Can we add some usage docs please? I'm not clear whats happening here until I read all the code.

---

*[app/common/directives/filter-system/filter-users.js, line 34 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform-client/190#-KI68WzUaS4gp-hiX_ZK:-KI68WzUaS4gp-hiX_ZL:-1937515990) ([raw file](https://github.com/ushahidi/platform-client/blob/5f2617e9a79f0ea5230e5e0674158f5f850fdb67/app/common/directives/filter-system/filter-users.js#L34)):*
> ```JavaScript
> 
>             $scope.applyFilters = function () {
>                 $scope.filters.roles = $scope.selectedRoles;
> ```

We could probably generalize this pattern for the entire form:
- on submit copy values from a `form` to `filters`

---

*[app/common/directives/filter-system/filter-users.js, line 35 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform-client/190#-KI67wKJrPpu3A0XtBAr:-KI67wKJrPpu3A0XtBAs:-900654722) ([raw file](https://github.com/ushahidi/platform-client/blob/5f2617e9a79f0ea5230e5e0674158f5f850fdb67/app/common/directives/filter-system/filter-users.js#L35)):*
> ```JavaScript
>             $scope.applyFilters = function () {
>                 $scope.filters.roles = $scope.selectedRoles;
>                 $scope.queryFnc();
> ```

I'm still struggling to figure out the best practice, but I think we'd be better to skip passing in a queryFnc and instead watch `filters` from the controller/outside this directive.

---

*[app/common/locales/en.json, line 28 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform-client/190#-KI63w2OvRFOKwIlxXI6:-KI63w2PL9c5afsok5CP:1621926885) ([raw file](https://github.com/ushahidi/platform-client/blob/5f2617e9a79f0ea5230e5e0674158f5f850fdb67/app/common/locales/en.json#L28)):*
> ```JSON
>     },
>     "toolbar": {
>         "searchbar" : {
> ```

unless theres a good reason for a deep hierarchy make this flatter. Its easy to scan through. The deep hierarchy hasn't helped with much so far.

---

*[app/setting/controllers/setting-users-controller.js, line 31 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform-client/190#-KI699HtsmttqYqvYWbi:-KI699HtsmttqYqvYWbj:-799008247) ([raw file](https://github.com/ushahidi/platform-client/blob/5f2617e9a79f0ea5230e5e0674158f5f850fdb67/app/setting/controllers/setting-users-controller.js#L31)):*
> ```JavaScript
>     });
> 
>     $scope.filters = {
> ```

I had in my head that we'd need a FilterService (with instances for UserFilters, CategoryFilters, etc) .. but now I look at this it seems the plan JS object works fine (at least for categories and users but maybe not posts)

---

*[app/setting/controllers/setting-users-controller.js, line 32 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform-client/190#-KI64TZR3sKqdqn2fBs7:-KI64TZSgYQr1JfavzZH:-300683094) ([raw file](https://github.com/ushahidi/platform-client/blob/5f2617e9a79f0ea5230e5e0674158f5f850fdb67/app/setting/controllers/setting-users-controller.js#L32)):*
> ```JavaScript
> 
>     $scope.filters = {
>         roles: [],
> ```

Param name should remain `role` as mentioned on platform diff.

---

*[app/setting/controllers/setting-users-controller.js, line 134 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform-client/190#-KI69RNnURbpL8I7GBm3:-KI69RNoqrjQ0wxdfglL:-1370023419) ([raw file](https://github.com/ushahidi/platform-client/blob/5f2617e9a79f0ea5230e5e0674158f5f850fdb67/app/setting/controllers/setting-users-controller.js#L134)):*
> ```JavaScript
>             offset: ($scope.currentPage - 1) * $scope.itemsPerPage,
>             limit: $scope.itemsPerPage,
>             roles: $scope.filters.roles,
> ```

Could we just do a generic extend(params, filters) here? ie.
```
_.extend({
  offset: ..
  limit: ..
}, $scope.filters)
```
or something like that.. I think we did that with the post filters previously?
This way we can add further filters without touching this code again.

---

*[server/www/templates/common/filter-system/filter-system.html, line 2 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform-client/190#-KI65Sx3RohEp2VgxAkq:-KI65Sx3RohEp2VgxAkr:616898965) ([raw file](https://github.com/ushahidi/platform-client/blob/5f2617e9a79f0ea5230e5e0674158f5f850fdb67/server/www/templates/common/filter-system/filter-system.html#L2)):*
> ```HTML
> <form role="search" class="searchbar" dropdown auto-close="outsideClick">
>     <div class="searchbar-input">
> ```

I think ideally we should move this keyword search field into a directive just to keep things clean.

---

*[server/www/templates/common/filter-system/filter-system.html, line 32 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform-client/190#-KI64geyyzId5sn3ZpOg:-KI64geyyzId5sn3ZpOh:1937789334) ([raw file](https://github.com/ushahidi/platform-client/blob/5f2617e9a79f0ea5230e5e0674158f5f850fdb67/server/www/templates/common/filter-system/filter-system.html#L32)):*
> ```HTML
>     </div>
> 
>     <div ng-switch="entityType">
> ```

I think we should consider these kind of `switch` solutions a warning sign that a directive is serving multiple purposes.

If we split these elements into a bunch of directives:
- keyword-search
- date filter
- role filter
- etc

We'd still have to copy/paste this top level structure, but it becomes easier to mix and match the individual filter controls

We might still be able to use transclude in a top level filter-system directive so we can avoid copy/pasting even the basic structure.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ushahidi/platform-client/190#-:-KI695Wxji67jWj2Oxnt:-2073611111)*

willdoran:
Ok so I've made significant changes to filters. 

Users now has a filter-user template and js. The filter-user directive takes a JSON object 'filters', it adds a removes the user specific filters to this objects.

Within the template is defined a searchbar and filter menu. Searchbar is a resuable directive that can be dropped into any filter directive.

Within filter menu there is a filter-role directive which is also reusable. Similar template like date and status etc can be created. This directive takes a list as param. Using the checklist model lib it updates this list setting or removing roles as appropriate.

When the user has entered a search term or select specific filters they can either hit enter or click Apply Filters. The result is that the local filters, q, roles etc are copied to the filter object.

The page user controller watches $scope.filters. It is set to use "deep watch" which watches for changes not of just the object but of the objects properties. 

On change, the user controller make an endpoint request with the associated filters.

For clarity the user filter is split into its own directive. It could be moved inside the users.html and the filter-user directive could be changed to an attribute. I felt that decomposing it into its own directive was appropriate given that it is a distinct sub element of the page.

Filter-user.js is responsible for collecting the filters and setting them on the filter object watched by the controller. I am not sure how to make this perfectly abstract. It would be preferable to have just a filter directive which takes the entries in the form and applies them to the filter. I'm not sure if this can be achieved for elements like checkboxes. Let me know if you see a better way that could replace the need to specify user/post/etc filters explicitly.

For posts, the pattern should be as follows:
1. Define filter-post.html, this should use identical markup to filter-user.html
2. Define filter-post.js, this should have an apply filter function
3. Create sub directive for filter-date etc

rjmackay:
I think this is pretty much done. Just need to implement the cancel button behaviour.

<details><summary><i>Previously, willdoran (will) wrote…</i></summary>
> Ok so I've made significant changes to filters. 
> 
> 
> 
> Users now has a filter-user template and js. The filter-user directive takes a JSON object 'filters', it adds a removes the user specific filters to this objects.
> 
> 
> 
> Within the template is defined a searchbar and filter menu. Searchbar is a resuable directive that can be dropped into any filter directive.
> 
> 
> 
> Within filter menu there is a filter-role directive which is also reusable. Similar template like date and status etc can be created. This directive takes a list as param. Using the checklist model lib it updates this list setting or removing roles as appropriate.
> 
> 
> 
> When the user has entered a search term or select specific filters they can either hit enter or click Apply Filters. The result is that the local filters, q, roles etc are copied to the filter object.
> 
> 
> 
> The page user controller watches $scope.filters. It is set to use "deep watch" which watches for changes not of just the object but of the objects properties. 
> 
> 
> 
> On change, the user controller make an endpoint request with the associated filters.
> 
> 
> 
> For clarity the user filter is split into its own directive. It could be moved inside the users.html and the filter-user directive could be changed to an attribute. I felt that decomposing it into its own directive was appropriate given that it is a distinct sub element of the page.
> 
> 
> 
> Filter-user.js is responsible for collecting the filters and setting them on the filter object watched by the controller. I am not sure how to make this perfectly abstract. It would be preferable to have just a filter directive which takes the entries in the form and applies them to the filter. I'm not sure if this can be achieved for elements like checkboxes. Let me know if you see a better way that could replace the need to specify user/post/etc filters explicitly.
> 
> 
> 
> We can set the elements on the form to map directly to the filter object at the input point. However, when doing this the $watch fires the moment - for example - a filter is clicked. This is interesting behaviour as it means that the user does not have to click or hit enter for the search to be run but it might be disconcerting and doesn't follow the behaviour defined by the PL.
> 
> 
> 
> For posts, the pattern should be as follows:
> 
> 1. Define filter-post.html, this should use identical markup to filter-user.html
> 
> 2. Define filter-post.js, this should have an apply filter function
> 
> 3. Create sub directive for filter-date etc
</details>


---

Reviewed 21 of 26 files at r2, 1 of 1 files at r3.
Review status: all files reviewed at latest revision, 16 unresolved discussions, some commit checks broke.

---

*[package.json, line 24 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform-client/190#-KIPd5s8OQHpv31kt8uV:-KIPd5s9Dnd6xwGDGpM-:1060675951) ([raw file](https://github.com/ushahidi/platform-client/blob/d98c9232a461cbb716f0ae767e828df5abbe4590/package.json#L24)):*
> ```JSON
>     "jscs-fix": "gulp jscsfix"
>   },
>   "pre-commit": [
> ```

erm.. did we really want to drop these?

---

*[app/common/directives/filter-system/filter-role.js, line 2 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform-client/190#-KIPdERD-2kZZqfKv4_g:-KIPdERD-2kZZqfKv4_h:-309496664) ([raw file](https://github.com/ushahidi/platform-client/blob/085de41af2ef58db52299cc44dbf047025783776/app/common/directives/filter-system/filter-role.js#L2)):*
> ```JavaScript
> /**
>  * Ushahidi Angular Filter System role directive
> ```

:+1: this looks like its probably simple enough to be useful outside of filters too

---

*[app/common/directives/filter-system/filter-searchbar.js, line 35 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform-client/190#-KIPcCurczQlBqP2PhsI:-KIPcCurczQlBqP2PhsJ:448424643) ([raw file](https://github.com/ushahidi/platform-client/blob/085de41af2ef58db52299cc44dbf047025783776/app/common/directives/filter-system/filter-searchbar.js#L35)):*
> ```JavaScript
>         scope: {
>             filters: '=',
>             placeholdeEntity: '='
> ```

typo? `placeholdeR`

---

*[app/setting/users/directives/filter-users.js, line 34 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform-client/190#-KI68WzUaS4gp-hiX_ZK:-KIPdkkinAX7iciTn6Xt:1729485968) ([raw file](https://github.com/ushahidi/platform-client/blob/5f2617e9a79f0ea5230e5e0674158f5f850fdb67/app/setting/users/directives/filter-users.js#L34)):*
<details><summary><i>Previously, rjmackay (Robbie Mackay) wrote…</i></summary>
> We could probably generalize this pattern for the entire form:
> - on submit copy values from a `form` to `filters`
</details>
Looks good.
You could even just bind to something like `$scope.editableFilters.q` then this would just be
`angular.copy($scope.editableFilters, $scope.filters)` .. but I'll leave that to when we implement the post-filters etc..

---

*[app/setting/users/directives/filter-users.js, line 27 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform-client/190#-KIPeDn3o2rCTpJ6O66O:-KIPeDn3o2rCTpJ6O66P:575275764) ([raw file](https://github.com/ushahidi/platform-client/blob/085de41af2ef58db52299cc44dbf047025783776/app/setting/users/directives/filter-users.js#L27)):*
> ```JavaScript
> 
>             $scope.cancel = function () {
>                 //TODO: check expected behaviour - should close dropdown
> ```

I assume it should:
1. close dropdown
2. reset active filters ie. copy $scope.filters values back into the form.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ushahidi/platform-client/190#-:-KIPenoD6oLvuuQG44P1:-304587687)*

PR #197, opened 2016-05-23
Re-organise app/common module into sub-modules

rjmackay:
- Move auth services etc into common/auth
- Start using thing.type.js naming convention
- Use subdirectories within common module

Test these changes by:
- Check app builds

Ping @willdoran @jasonmule

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/197)


rjmackay:
@willdoran @jasonmule what do you think? would this make things clearer? ..

willdoran:
Yeah, I like it alot. I want to split the settings dir into categories, sets etc. I did it for users under the filter system. 

I think as much as possible we should make the dirs a readable path through the app.

rjmackay:
Not really sure what to do with the remaining components that are still in services/directives/etc
One or two could go into a `layout` folder. The rest don't follow much of a theme.

rjmackay:
I'm going to abandon this - I've rearranged the auth code in my register PR - other changes can happen as needed.

PR #198, opened 2016-05-23
New modal directive and service

rjmackay:
- Add `ush-modal-container` directive
- Add `ModalService`
  - Modals can be triggered by `ModalService.open(templateUrl, title ...)`

Todo: 
- [ ] slider service
- [ ] Split notification system templates and use modal/slider services
- [ ] Modify existing `<modal>` to delegate to ModalService

Test these changes by:
- N/A yet..

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/198)

### DIFF
--- a/app/common/common-module.js
+++ b/app/common/common-module.js
@@ -39,6 +39,7 @@ angular.module('ushahidi.common', [
 .service('PasswordReset', require('./services/password-reset.js'))
 .service('IconManager', require('./services/icon-manager.js'))
 .service('FontAwesomeIcons', require('./services/endpoints/FontAwesomeIcons.js'))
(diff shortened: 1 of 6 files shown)


rjmackay:
@willdoran @jasonmule this should avoid the need for many many `<modal>` directive .. but makes it easier to open a modal from anywhere and not worry about where the markup is: its always in its own template.

I probably also need to add a variant which takes the template directly, rather than taking a templateUrl..
I could modify the `ModalService.open()` to do both, but maybe a different method might be easier?

willdoran:
That looks much cleaner.

PR #201, opened 2016-05-25
Mustang/post filters

rjmackay:
- Add post toolbar directive
- Add post filters directive
- Add directives for each filter control
- Add mode context form filter
- Reorganise post views directives/controller etc into `app/posts/views`

To do:
- [ ] Add date picker ushahidi/platform#1078
- [x] Finish wiring date selector
- [x] Finish wiring published to selector
- [x] Add location selector
- [x] Update user settings to leverage ng-model-options
- [x] Wire up context filter to filter state

Test these changes by:
- Not yet..

Fixes ushahidi/platform#1050 ushahidi/platform#1049

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/201)


rjmackay:
@willdoran I'm not done yet - but some feedback would be helpful.
Particularly around:
1. the PostFilterService and how I'm managing state
2. the de-composition of directives etc.
3. the file structure .. I think I've stuffed too much into post/views now :(

server/www/templates/common/filter-system/filter-searchbar.html
```diff
@@ -1,9 +1,9 @@
 <div class="searchbar-input" dropdown auto-close="outsideClick">
     <!-- Search input -->
     <div class="form-field search">
-        <input name="q" type="search" 
-        placeholder="{{ 'toolbar.searchbar.search_entity' | translate:{entity: placeholderText} }}" 
-        ng-model="q">
+        <input name="q" type="search"
+        placeholder="{{ 'toolbar.searchbar.search_entity' | translate:{entity: placeholderText} }}"
+        ng-model="filters.q">
```
willdoran: The way I had done this in users was that in the filter-users applyFilters() it copies the value of the q into filters.q. Wiring it directly to filters.q caused the $watch to fire the moment filters.q was updated so before the user hit enter/applyFilters(). This is because I was using the deep $watch setting. 
Without deep watch the $watch wasn't being fired because it $watch consider the object not to have been changed.
I'm not sure what the best work around for this would be but I would much prefer this to be filters.q and to drop having to ask the form for the value of q

server/www/templates/posts/views/category-select.html
```diff
@@ -0,0 +1,10 @@
+<fieldset ng-show="categories.length">
+    <legend translate="nav.categories">Categories</legend>
+
+    <div class="form-field checkbox" ng-repeat="(index, category) in categories">
+        <label>
+            <input checklist-value="category.id" checklist-model="model" type="checkbox" name="selectedCategories" > {{ ::category.name }}
```
willdoran: For the checklist-model, when you want to clear the filters checklist model UI may not respect model=[]
https://github.com/vitalets/checklist-model#usage
I used their suggested work around of splice but there may be a better way to do it.

willdoran:
I really like the components. For the category/visible_to/date/form/location components, could they live in templates/common/filter-system? Since they are reusable and belong to filters more closely then posts. Alternatively, a sub dir in views that governs filters might make it more readable and would reduce the volume of things in views.
Would a dir structure like this be too much:
/common
    /filter-system
         filter-role.html
         filter-category.html
         filter-visiblity.html
         filter-form.htmll  
/views
   /timeline
   /map
   /toolbar
  /searchbar
      /filters
      /savedsearch
  

I wonder if searchbar should take filters.q so that it's consistent with the others.
When you add the watch, there is a deep watch switch that I used for Users:
http://teropa.info/blog/2014/01/26/the-three-watch-depths-of-angularjs.html
It is expensive for large/complex objects though - not sure if performance wise it would be better to use either the collection watch or specific watches per item.

I added caveat notes to highlight potential issues that i found on filter-user.js

server/www/templates/common/filter-system/filter-searchbar.html
rjmackay: I did some reading. Adding `ng-model-options="{ updateOn : 'submit' }"` to the form.. makes angular queue model updates until you click submit. Which means angular basically handles the entire submit/reset workflow for us :)

willdoran: Wow...that is immensely great :)

server/www/templates/posts/views/category-select.html
rjmackay: Yea thats a general failing of how angular scopes work. I had many issues - this way is working so far. But its possible I'll need to pass the entire `filters` object through or reference `$parent.model`

rjmackay:
> When you add the watch, there is a deep watch switch that I used for Users:
.. I think this is working fine without it since filters is always just a single level (intentionally) but I'll test.

rjmackay:
@willdoran this is done. Can you give it a review?

rjmackay:
Review status: 0 of 41 files reviewed at latest revision, 7 unresolved discussions.

---

*[app/post/post-module.js, line 24 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform-client/201#-KIfHkC6e8PpMoULa_kD:-KIfHkC6e8PpMoULa_kE:1350937130) ([raw file](https://github.com/ushahidi/platform-client/blob/214dba8c90385c2b7f2047948537f8b153a57580/app/post/post-module.js#L24)):*
> ```JavaScript
> .directive('addPostButton', require('./views/add-post-button.directive.js'))
> .directive('filterPosts', require('./views/filters/filter-posts.directive.js'))
> .directive('modeContextFormFilter', require('./views/mode-context-form-filter.directive.js'))
> ```

Could move this under filters - but its different from the rest.

---

*[app/post/post-module.js, line 26 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform-client/201#-KIfH_5urr8oFM7JdGv9:-KIfH_5v-1g0d7_yzePo:1059062626) ([raw file](https://github.com/ushahidi/platform-client/blob/214dba8c90385c2b7f2047948537f8b153a57580/app/post/post-module.js#L26)):*
> ```JavaScript
> .directive('modeContextFormFilter', require('./views/mode-context-form-filter.directive.js'))
> .directive('postToolbar', require('./views/post-toolbar.directive.js'))
> .directive('filterCategory', require('./views/filters/filter-category.directive.js'))
> ```

I've kept these filters in the post module for now since they're not used anywhere else yet. We should move them out if they get reused. 

---

*[app/post/views/post-filters.service.js, line 22 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform-client/201#-KIfI6FOi1hVKQ-E6Qlo:-KIfI6FPaD2UgibntS1F:1055622298) ([raw file](https://github.com/ushahidi/platform-client/blob/214dba8c90385c2b7f2047948537f8b153a57580/app/post/views/post-filters.service.js#L22)):*
> ```JavaScript
> 
>     function activate() {
>         FormEndpoint.query().$promise.then(function (forms) {
> ```

I'm not sure happy that this is happening here. But it'll do for now.

---

*[app/post/views/post-view-list.directive.js, line 62 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform-client/201#-KIfHs-kDnuvAOIQbt6x:-KIfHs-kDnuvAOIQbt6y:-756841115) ([raw file](https://github.com/ushahidi/platform-client/blob/214dba8c90385c2b7f2047948537f8b153a57580/app/post/views/post-view-list.directive.js#L62)):*
> ```JavaScript
>             getPostsForPagination();
>         }
>     }, true);
> ```

Turns out you're right! Definitely needed the deep watch.

---

*[server/www/templates/settings/users/filter-users.html, line 1 \[r2\]](https://reviewable.io:443/reviews/ushahidi/platform-client/201#-KIfI0XsLxBzoKVJHKrs:-KIfI0XsLxBzoKVJHKrt:841380072) ([raw file](https://github.com/ushahidi/platform-client/blob/214dba8c90385c2b7f2047948537f8b153a57580/server/www/templates/settings/users/filter-users.html#L1)):*
> ```HTML
> <form name="usersFiltersForm" role="search" class="searchbar" dropdown is-open="filtersMenuOpen" auto-close="outsideClick" ng-submit="applyFilters()" ng-model-options="{ updateOn : 'submit' }">
> ```

`ng-model-options` does some magic :)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ushahidi/platform-client/201)*

willdoran:
@rjmackay this looks good can we land it? I'd like to pull it into my savedsearch branch

zhalloran:
@rjmackay I know this is already merged, but I'd like to offer a different approach here to the PostFiltersService. As it stands right now, the form endpoint doesn't return any information about the number of posts associated with a given form. We could use the Posts records to drive what appears in the form filters for the mode-context OR we could modify the form endpoint to return information about published posts. I'd actually lean towards adding that info to the form endpoint.

willdoran:
@zhalloran That's actually kind of interesting. At the moment endpoints are very unary. Having a data layer that allows us to dynamically mix associated data would be very useful. I think @rjmackay has already suggested something along these lines. If it mimicked something like a nosql db that allowed you to pull data straight as intelligent groupings that would be good. So you could essentially see the data a vat of goo, and we could window into it and return intelligently associated data. 
It would be difficult to do this without adopting a document or nosql schema or something like elasticsearch or solr both of which are java and both of which are heavy on machines.

rjmackay:
I'd agree it'd be ideal for the form endpoint to return those counts. The simplest way to do that might just be to modify the API to include that info (either in every request or based on a parameter)

>  I'd like to offer a different approach here to the PostFiltersService.
Does this change the details of the PostFiltersService at all? I think it'd just happen alongside it.

PR #205, opened 2016-05-27
NEW: Angular Date Picker

middle8media:
- Adds https://github.com/alongubkin/angular-datepicker

Test these changes by:
- click "Filters" in toolbar
- scroll to date fields
- click "start date" field
- date picker should slide from bottom
- interact with date picker (pick a date, clear, close etc...)
- repeat for "end date"

Fixes ushahidi/platform#1078

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/205)


app/app.js
```diff
@@ -8,7 +8,7 @@ require('angular-resource');
 require('angular-translate');
 require('angular-translate-loader-static-files');
 require('angular-ui-bootstrap');
-require('angular-datepicker');
+require('angular-datepicker/build/angular-datepicker');
```
rjmackay: Try this as just `require(‘angular-datepicker’)` - then it loads whatever is set as `main` in package.json for that package

middle8media: This breaks the app. I can only get it to work with `require('angular-datepicker/build/angular-datepicker');`

image

rjmackay: Yup - you were right to start with. They're package isn't set up properly

rjmackay:
Added this to the post filters and it works perfectly!

PR #207, opened 2016-05-31
Wire up saved searches dropdown and views

rjmackay:
- Wire up saved searches dropdown
- Wire up saved searches view
- Minor fixes to collections

Todo:
- [x] Wire up active filters
- [x] Create saved search
- [x] Update filters on save search
- [x] Fix up breadcrumb title
- [ ] Track state properly when switch map/timeline mode

Ping @ushahidi/platform
Refs ushahidi/platform#1060 ushahidi/platform#1080 ushahidi/platform#1081

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/207)


app/set/directives/savedsearches/mode-context.directive.js
```diff
@@ -0,0 +1,113 @@
...
+        });
+    }
+
+    // Check if we can edit
+    function canEdit(savedSearch) {
+        return _.contains(savedSearch.allowed_privileges, 'update');
+    }
+
+    function editSavedSearch() {
+        ModalService.openTemplate('<saved-search-editor saved-search="savedSearch"></saved-search-editor>', 'set.edit_search_settings', 'star', $scope, false, false);
```
rjmackay: This now works to open a modal. I think for single directive tag templates like this, this is probably the preferred method (vs a template file). Anything else should go into a template file.

app/set/directives/savedsearches/editor-directive.js
```diff
@@ -67,26 +58,14 @@ function (
                 .then(function (savedSearch) {
                     $location.url('/savedsearches/' + savedSearch.id);
                     $scope.savedSearch = _.clone(savedSearch);
-                    $scope.isOpen.data = false;
+                    $scope.$parent.closeModal();
```
rjmackay: `closeModal` is pushed into the scope by modal directive. We use $parent here because this directive has created an isolate scope

rjmackay:
This has gotten away from me a little as the collections, searches and all the post filter logic is very closely coupled.
Its 95% there but there's a bunch of issues with filter bindings breaking.. so I need to rethink how that works :/

rjmackay:
@willdoran @jasonmule squashed and merged but I'd still appreciate review feedback

PR #218, opened 2016-06-03
Fix endpoint get fresh to properly remove cache key

willdoran:
- Properly set params passed to getFresh
- Update tests for getFresh
- Fix removal of cache key based on url

Test these changes by:
- Ensure that cache keys are removed on getFresh for Tags, Roles, Permissions, Forms, Form Attributes, Form Stages and Data provider

Fixes ushahidi/platform# .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/218)

### DIFF
--- a/app/common/services/endpoints/config.js
+++ b/app/common/services/endpoints/config.js
@@ -15,11 +15,6 @@ function (
         cache = new CacheFactory('configCache');
     }
 
-    cache.setOnExpire(function (key, value) {
-        ConfigEndpoint.get(value.id);
-    });
-
-
     var ConfigEndpoint = $resource(Util.apiUrl('/config/:id'), {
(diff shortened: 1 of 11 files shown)


rjmackay:
- This is missing at least the config endpoint
- Also some endpoints also have
```
    cache.setOnExpire(function (key, value) {
        ConfigEndpoint.get(value.id);
    });
```
which never works because `value` is never populated. I think we get the URL in the args so you could fetch that. But TBH we should probably just skip the setOnExpire.

willdoran:
@rjmackay I've added config endpoint, and removed setOnExpire. I'm searched for instances of cache. It is currently only used by settings' entities.

PR #226, opened 2016-06-08
Modifies add post button to automatically redirect to /post/create/{{…

zhalloran:
- Modifies the add post button to automatically redirect to /posts/create/{{form.id}} if there's only one survey type

Test these changes by:
- Load map with only one survey type and click on the add post button

Fixes ushahidi/platform#1119 .

Ping @ushahidi/platform

…form.id}} if there's only one survey.

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/226)


zhalloran:
@rjmackay @jasonmule @willdoran Please review code when you get a chance.

One thing I noticed: when transitioning from /views/map to /posts/create/{{form.id}} for the first time there's a bit of a flicker. I thought it might be my implementation, so I tried adding an href to the template but it exhibited the same behavior. It might just be the result of the first time render of the mode-context for the post-create route. Any ideas?

app/post/views/add-post-button.directive.js
```diff
@@ -29,14 +33,19 @@ function AddPostButtonController(
...
         // Load forms
         $scope.forms = FormEndpoint.query();
+        
     }
 
     function toggleFab() {
-        $scope.fabToggle = !$scope.fabToggle;
-        if ($scope.fabToggle) {
-            $scope.fabOptionsStyle = { opacity: 1, display: 'flex' };
+        if (_.size($scope.forms) == 1) {
```
rjmackay: `$scope.forms.length` should be sufficient. `_.size` is kinda overkill

zhalloran: Yup. That was pretty dumb :)

PR #230, opened 2016-06-09
Notifications refactor

rjmackay:
- Replace notification slide and notification system with new slider directive and server
- Update Notify service with clearer methods
- Update all other code to use new methods
- [x] Open slider inside modal when modal is visible

Test these changes by:
- gulp test

Fixes ushahidi/platform#1120

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/230)


rjmackay:
This is a bit scary.. but in mostly just the same changes over and over again.
I've generally attempted to simplify the notifications code so theres only a few short templates and most logic is in the service or directive. It takes a similar approach to the ModalService.

Next step is to track the state of the modal so we can open the slider IN the modal when its up. But its too later for that tonight.

jasonmule:
Looking good...

PR #240, opened 2016-06-10
Addresses issues where the zoom control position default is enforced …

zhalloran:
- Fixes a bug where the zoom control position defaults to 'topleft' when navigating.
- Fixes bug where attribution prefix is restored when navigating back to /views/map

Test these changes by:
- navigate to /views/map, navigate to another page and then navigate back to /views/map

Fixes ushahidi/platform#1110 .

Ping @ushahidi/platform

…when navigating away from and then back to /views/map. While working on this I noticed that the attribution prefix was also being turned on when navigating away from and back to /views/map. Addressed that issue as well in this commit.

- sets zoomControlPosition to 'bottom right'
- runs map.init() in post-view-map.directive so the attributionPrefix is set to false.

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/240)

### DIFF
--- a/app/common/services/maps.js
+++ b/app/common/services/maps.js
@@ -83,7 +83,7 @@ function (
         getInitialScope: function () {
             return {
                 defaults: {
-	                zoomControlPosition: 'bottomright',
+                    zoomControlPosition: 'bottomright',
                     scrollWheelZoom: false
                 },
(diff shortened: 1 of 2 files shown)


rjmackay:
.. our map code is insane. But this looks good :)

PR #241, opened 2016-06-10
Add post export button

jasonmule:
- Adds export button for posts

Test these changes by:
- Clicking on the export button on the post toolbar.

Fixes ushahidi/platform#1103 .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/241)

### DIFF
--- a/app/post/export/post-export-directive.js
+++ b/app/post/export/post-export-directive.js
@@ -1,0 +1,75 @@
+module.exports = PostExportDirective;
+
+PostExportDirective.$inject = [];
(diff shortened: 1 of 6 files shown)


app/post/export/post-export-directive.js
```diff
@@ -0,0 +1,75 @@
+module.exports = PostExportDirective;
```
rjmackay: I think this should probably still be under app/post/views - since its only used on views right now.

PR #255, opened 2016-06-14
Load images, date and author details for map pop-ups

jasonmule:
- Loads images, dates and author details for map pop-ups

Test these changes by:
- Clicking on a marker to load a pop-up

Fixes ushahidi/platform#1149.

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/255)

### DIFF
--- a/app/common/services/maps.js
+++ b/app/common/services/maps.js
@@ -70,46 +70,23 @@ function (
                 getPostDetails(feature).then(function (details) {
                     var scope = $rootScope.$new();
 
-                    details.content = $filter('truncate')(details.content, 150, '...', true);
+                    // details.content = $filter('truncate')(details.content, 150, '...', true);
                     scope.post = details;
 
-                    var e = $compile('<post-card post="post"></post-card>')(scope);
+                    var el = $compile('<post-card post="post"></post-card>')(scope);
 
-                    that.bindPopup(e[0]).openPopup();
+                    that.bindPopup(el[0], {
+                        'minWidth': '300',
+                        'maxWidth': '300',
+                        'className': 'pl-popup'
+                    }).openPopup();
                 });
             });
         }
     };
 
     var getPostDetails = function (feature) {
-        var attributes = [], deferred = $q.defer(), mediaId;
-
-        PostEndpoint.get({id: feature.properties.id}).$promise.then(function (post) {
-            // Grab form attributes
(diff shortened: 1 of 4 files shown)


jasonmule:
Probably not the most efficient way to pull in extra post details.

app/post/views/post-view-map.directive.js
```diff
@@ -29,12 +37,91 @@ function (
             query = query || PostFilters.getQueryParams($scope.filters);
 
             $scope.isLoading = true;
+
+            var postDetailPromises = [];
+
             return PostEndpoint.geojson(query).$promise.then(function (posts) {
-                map.reloadPosts(posts);
-                $scope.isLoading = false;
+                _.each(posts.features, function (feature) {
```
rjmackay: This is going to be terrible for deployments that have a few posts. We only need the full post for the single post with an open popup, so loading the data for ALL of them is a waste..

app/common/services/maps.js
```diff
@@ -55,22 +55,51 @@ function (
     var geojsonLayerOptions = {
         onEachFeature: function (feature, layer) {
             var description = feature.properties.description || '',
-                title = feature.properties.title || feature.properties.id;
+                title = feature.properties.title || feature.properties.id,
```
rjmackay: I think we should try to use angular $compile and render the `post-card` directive here.. otherwise we're repeating the same code in 2 places. There's very little difference between a post card in the list view, and a post card in the popup.

rjmackay:
Review status: 0 of 2 files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

*[app/common/services/maps.js, line 89 \[r3\]](https://reviewable.io:443/reviews/ushahidi/platform-client/255#-KL-VLRQH7-3SW24PLjJ:-KL-VLRRPGQ51m3vWbtC:506217633) ([raw file](https://github.com/ushahidi/platform-client/blob/36995799cf9a87247ebf994146ea8b829ecbb578/app/common/services/maps.js#L89)):*
> ```JavaScript
>         PostEndpoint.get({id: feature.properties.id}).$promise.then(function (post) {
>             // Grab form attributes
>             FormAttributeEndpoint.get({formId: post.form.id}).$promise.then(function (response) {
> ```

This is all unnecessary now you're using `post-card` right? It only needs the post as-is

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ushahidi/platform-client/255)*

PR #259, opened 2016-06-20
Reorg sets into collections and savedsearches modules

rjmackay:
- Split sets module into saved searches module and collections module

Test these changes by:
- `gulp test`

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/259)


rjmackay:
I want to move the sets/ code into posts/ too since its all tied to posts too. I'm not sure how to structure it though.. thoughts so far

- Collection view is part of post/views
- Saved search view is part of post/views
- Collection modal is post/shared
- Saved search dropdown is post/views or post/views/filters
- Saved search create/update directives are just part of the mode context. Either for vanilla post/views or savedsearch view.

Maybe the split is something like:

- post/views/shared: filters, map, list etc
- post/views/all: just the post-views controller
- post/views/collection: collections controller + mode context
- post/views/savedsearch: saved search controller + mode context

willdoran:
Stylistically, do we need to prepend all the files with post, would it be possible to design the naming scheme so that the directory structure tells you were you are and what you're working with, so, for example,
post/collections/listing.directive.js or post/views/modify/editor.directive.js
I would generally be in favour of more directories and shorter file names.

Instead of shared can we use common? Or can we change any other occurrences of common to be shared.

I'm not positive about the use of "views" I think it might cause confusion with templates that are generally considered views. I would suggest that instead we break it apart,

Post/Collection
Post/Common
Post/Detail
Post/Modify
Post/Posts/Activity
Post/Posts/Map
Post/Posts/Timeline
Post/Savedsearch

It can be really hard to discern from file names what the interrelation is between code and final output, I think the simplest way to do this is to use directories to guide people. Eh, to use an analogy a lot of people have a "random crap" drawer in their house that contains old keys, cables, manuals for obscure kitchen devices, old coins. I used to have a lot of those locations, then I married a German now everything has a place and if it doesn't have a place it runs the risk of being removed from the apartment :)

My fear for shared/common is that it become like common in app that we put things that don't belong to anything specific there until it gets large enough to warrant new dirs being created. Instead, I would suggest that if something exists and it's not explicitly shared between multiple entities that it be placed in its own directory. This will increase the number of directories but hopefully it will make the src more readable.

willdoran:
@rjmackay @jasonmule ^^

rjmackay:
- We can definitely drop to posts-* naming.. I actually meant to but forgot.
- Happy to do `common` not shared.
- I'm not sure about `Post/Posts` and `Posts/Collections`
  - Collections, saved searches and "All posts" are kinda the same level
  - Map and list are shared by both. Also only have 1 file each..
  - Activity kinda stands on its own (doesn't share any code right now). In theory it should expand to encompass more than just posts too.

I think maybe we have
- Posts/ViewAll/[list|map|activity]
- Posts/

.. I wonder if the entire app could be split into:  common, settings, posts.

willdoran:
@rjmackay Yeah that dir structure looks good
So plans would be logical to move under setting, user-profile could go there too. I would quite like a simple top level like with common, settings, posts

PR #262, opened 2016-06-22
Fix empty reply message

jasonmule:
- Fixes issue where reply text is empty when sending a message 

Test these changes by:
- Replying to a message in a thread

Fixes ushahidi/platform#1227 .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/262)

### DIFF
--- a/app/post/detail/post-messages.directive.js
+++ b/app/post/detail/post-messages.directive.js
@@ -73,7 +73,7 @@ function (
                 ModalService.close();
 
                 var reply = {
-                    message: $scope.reply_text,
+                    message: $scope.message.reply_text,
                     direction: 'outgoing',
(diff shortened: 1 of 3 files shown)


jasonmule:
I think this didn't take into account that child scope in the modal would be destroyed.

app/post/detail/post-messages.directive.js
```diff
@@ -94,6 +92,7 @@ function (
             };
 
             $scope.reply = function () {
+                $scope.message = {};
                 $translate('post.messages.send').then(function (title) {
```
rjmackay: title's are auto translated, no need to wrap this.

PR #264, opened 2016-06-23
Fixing redirect loop, adding private placeholder page for private dep…

willdoran:
- Adding logic to avoid redirect loop
- Create private page for private deployments to be shown behind login modal

Test these changes by:
- Set a deployment to private
- Logout, you should see the private page and the login modal
- On login you should be directed to /

Fixes ushahidi/platform#1232
Fixes ushahidi/platform#1269

Ping @ushahidi/platform

…loyments when not logged in

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/264)


app/common/auth/login.controller.js
```diff
@@ -1,7 +1,8 @@
 module.exports = LoginController;
 
-LoginController.$inject = ['Authentication','$location'];
-function LoginController(Authentication, $location) {
+LoginController.$inject = ['Authentication','$location', 'Features'];
+function LoginController(Authentication, $location, Features) {
     Authentication.openLogin();
-    $location.url('/');
+    var path = Features.isFeatureEnabled('private') ? '/private' : '/';
```
rjmackay: This is checking if the deploy is allowed to make the deployment private, not if the deploy IS private. Very different things.

willdoran: Hmmm, that's unfortunate naming, I'll change that.

PR #268, opened 2016-06-24
UUID for survey attribute keys

willdoran:
- Remove key field from survey field editor
- Add uuid generation for attribute key to avoid collision with other attribute keys

Test these changes by:
- Add attributes with the same name to different surveys and ensure that no error is returned when saving the surveys.

Related to [ushahidi/platform/pull/1275](https://github.com/ushahidi/platform/pull/1275l)

Fixes ushahidi/platform#1236

Ping @ushahidi/platform

… prevent user changing it

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/268)

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -21,7 +21,6 @@ window.d3 = require('d3'); // Required for nvd3
 require('./common/wrapper/nvd3-wrapper');
 require('angular-nvd3/src/angular-nvd3');
 require('angular-cache');
-require('angular-uuid');
 
 // Load ushahidi modules
 require('./frame/frame-module.js');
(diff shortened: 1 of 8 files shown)


app/app.js
```diff
@@ -21,6 +21,7 @@ window.d3 = require('d3'); // Required for nvd3
 require('./common/wrapper/nvd3-wrapper');
 require('angular-nvd3/src/angular-nvd3');
 require('angular-cache');
+require('angular-uuid');
```
jasonmule: Sorry, should have mentioned that we should probably generate this at the backend since we no longer need the API user to provide/ change this?

willdoran: Yeah, we can I suppose move it to the backend.

willdoran:
I'll add the uuid element to the api on Monday morning.

rjmackay:
LGTM

rjmackay:
Make sure to move the issue into Testing on the board, not straight to closed

PR #269, opened 2016-06-28
Clear filters when navigating to home

jasonmule:
- Clears filters when user navigates to home

Test these changes by:
- Start a search and navigate elsewhere. Going back to home should clear the search term in the search box and mode context bar.

Fixes ushahidi/platform#1200 .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/269)

### DIFF
--- a/app/common/global/event-handlers.js
+++ b/app/common/global/event-handlers.js
@@ -1,9 +1,11 @@
 module.exports = [
     '$rootScope',
     '$location',
+    'PostFilters',
 function (
     $rootScope,
-    $location
+    $location,
+    PostFilters
 ) {
     $rootScope.switchRtl = function () {
         $rootScope.rtlEnabled = !$rootScope.rtlEnabled;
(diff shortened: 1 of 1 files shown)


app/common/global/event-handlers.js
```diff
@@ -20,4 +22,11 @@ function (
     $rootScope.toggleModalVisible = function (state) {
         $rootScope.modalVisible = (typeof state !== 'undefined') ? state : !$rootScope.modalVisible;
     };
+
+    // Clear filters when navigating to home
+    $rootScope.$on('$locationChangeSuccess', function () {
```
rjmackay: I don't think this should happen here. I'd like to minimize global events over time.. since they make it less obvious where this happens.

Ideally this should happen in the post-views.controller.js instead, however calling `clearFilters()` will clear the filters when we switch map/timeline views. Maybe we can check the previous location there? otherwise we need to look at tracking state (are we on home, collection or saved searches?) in PostFilters..

rjmackay:
Closing this till we have time for a more complete solution

PR #283, opened 2016-07-04
Hide nested options in until parent is selected

jasonmule:
- Hides nested options for fieldset until parent is selected.

Test these changes by:
- Edit a post and select "Who can see this" from the dropdown.
- Select "Specific roles" to see nested options.

Fixes ushahidi/platform#1131 .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/283)


rjmackay:
Tested. Once the roles list is open, it never closes again. I think it should close if you select Everyone or Just you again. That said - we could land as is since this will get replaced in Sprint 8.

PR #285, opened 2016-07-04
Notifications: User can always add contact

rjmackay:
- Don't hide add contact just because initial contact isn't saved.

Fixes ushahidi/platform#1093

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/285)

### DIFF
--- a/app/user-profile/directives/notifications.js
+++ b/app/user-profile/directives/notifications.js
@@ -69,11 +69,6 @@ module.exports = [
                                     isLoginEmail: true
                                 });
                             });
-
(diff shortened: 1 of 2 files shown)


willdoran:
@rjmackay This looks good BUT did you make a change on the backend too? I made the exact same changes last night and I noticed the following:
- Create new user
- Go to notification settings
- Ignore reg email
- Click add account
- save, go back to settings both reg and new account are enabled

Dunno if that's acceptable, it's odd behaviour but not the worst thing in the world

PR #304, opened 2016-08-07
ushahidi/platform#1320 add 'deploy' task to gulp with cachebusting of css and js

tuxpiper:
-
Adds a 'deploy' gulp task that makes a copy of the build, cachebusting some of the resulting assets, specifically: css and js files (tried to do images too, but references are too spread around).

Fixed the 'transifex-download' task to provide feedback of when it's complete (so it can be used as a dependency in other tasks).

The 'release' task is now based on this 'deploy' task so we distribute bundles with cachebusted filenames.

Test these changes by:
-
Run gulp deploy

Fixes ushahidi/platform#1320 .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/304)

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -20,7 +20,11 @@ var gulp         = require('gulp'),
     gzip         = require('gulp-gzip'),
     jscs         = require('gulp-jscs'),
     dotenv       = require('dotenv'),
-    Transifex    = require('transifex');
+    async        = require('async'),
+    Transifex    = require('transifex'),
+    CacheBuster  = require('gulp-cachebust');
+
+var cachebust = new CacheBuster();
 
 // Grab env vars from .env file
 dotenv.load({silent: true});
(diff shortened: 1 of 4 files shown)


gulpfile.js
```diff
@@ -398,7 +403,7 @@ gulp.task('heroku:dev', ['build'], function () {});
  * Task: `transifex-download`
  * Download translations from www.transifex.com
  */
-gulp.task('transifex-download', function () {
+gulp.task('transifex-download', function (gulp_cb) {
```
rjmackay: general convention is that `gulp_cb` should be called `done`

rjmackay:
- How long does cachebusting take?
- I'd prefer to include this in the standard build pipeline so we're always cachebusting
- I'd guess we might need to move template and index.html into `app/` so we're always processing them and outputting to `server/www`.

gulpfile.js
tuxpiper: ah.. the gulp documentation uses `cb`, but I've always found it confusing to have the two different callbacks when using `async`. Anyway, `done` sounds good.

tuxpiper:
In my machine it usually took a few seconds (much less than it takes to run, for instance, the "font" task).

In any case, moving index.html, templates etc to a folder that is not touched at all by gulp sounds like a good idea.

tuxpiper:
Closing in favor of #305

PR #325, opened 2016-08-31
Checkbox and radio button display issues

willdoran:
- Update detail page to ensure that image field appears first for each task
- Fixes layout issues for radio and checkbox buttons

Test these changes by:
- The following steps should be tested:
- [ ] Create a survey with checkbox and radio button fields on both the post and task sheets
- [ ] Ensure that the checkbox and radio buttons display as expected according to the pattern library:
http://preview.ushahidi.com/platform-pattern-library/develop/assets/html/2_fragments/#pl-pattern-forms
- [ ] Save the post, ensure that all checkbox and radio field values are displayed correctly on the Post Detail view
- [ ] Change checkbox and radio button fields to required on your test Survey.
- [ ] Create a new post and test that you are unable to save/validation errors are displayed next to the checkbox and radio fields. 

Fixes ushahidi/platform#1353

Ping @ushahidi/platform

…eckbox on post edit

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/325)

### DIFF
--- a/app/post/modify/post-editor.directive.js
+++ b/app/post/modify/post-editor.directive.js
@@ -127,9 +127,7 @@ function PostEditorController(
                         } else {
                             $scope.post.values[attr.key] = [null];
                         }
-                    } else if (attr.input === 'checkbox') {
-                        $scope.post.values[attr.key] = [];
-                    } else if (attr.input === 'number') {
+                    }  else if (attr.input === 'number') {
                         $scope.post.values[attr.key] = [parseInt(attr.default)];
                     } else if (attr.input === 'date' || attr.input === 'datetime') {
                         $scope.post.values[attr.key] = attr.default ? [new Date(attr.default)] : [new Date()];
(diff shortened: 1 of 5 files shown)


jshorland:
@willdoran When I save post and go back to post detail page, I don't see the full post with all fields, so I'm not sure if radio buttons appear as they should on post detail. I only see title, description and map. Also, all tasks are not appearing below the post detail. Only the first and the last. I went back into post edit mode, completed all tasks, and went back to post detail view. I now see one additional task, but still not all of them. Should I open a new issue for this?

screen shot 2016-09-01 at 9 52 31 am

willdoran:
@jshorland I'll take a look at it now and try to reproduce

willdoran:
@jshorland Ah ha, ok, I see the problem. So we only show tasks if they have values on the detail page, if they have no values they would just be blank.

Is it preferable to show blank tasks or to only show them if they have values?

jshorland:
Show blank tasks is preferable, as they would still need to be marked as complete by signed in users.

willdoran:
Ok I'll change this tomorrow

PR #328, opened 2016-09-05
approach to building the pattern library that is friendlier to custom…

tuxpiper:
-
Allows the pattern library to be overridden by contents in vendor/platform-pattern-library . I've been struggling in custom deployments with this, and the latest changes have made it nearly impossible (for whatever obscure reason only npm knows)

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/328)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@ vault.txt
 *.swp
 *.svg
 server/www/img/icons/
+vendor/*
+
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -21,13 +21,25 @@ var gulp         = require('gulp'),
     jscs         = require('gulp-jscs'),
     dotenv       = require('dotenv'),
     Transifex    = require('transifex'),
-    runSeq       = require('run-sequence');
+    runSeq       = require('run-sequence'),
+    exec         = require('child_process').exec;
 
 // Grab env vars from .env file
 dotenv.load({silent: true});
 // load user defined options
 if (fs.existsSync('.gulpconfig.json')) {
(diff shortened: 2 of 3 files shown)


tuxpiper:
Closing for the moment. Going to look at bower before really suggesting going through with this approach

tuxpiper:
This is not such a big blocker as I meant it to be initially (npm actually does its thing), but consensus seems to be that, anyway, it would be nice for people doing customisations to have a custom path for overriding the PL

.gitignore
```diff
@@ -17,3 +17,5 @@ vault.txt
 *.swp
 *.svg
 server/www/img/icons/
+vendor/*
```
rjmackay: Could we be explicit about this being a customizations dir? ie call it `custom/` 
It doesn't serve the same role as `vendor` since thats already taken by `node_modules` but usually (ie. in PHP) vendor = installed packages.

gulpfile.js
```diff
@@ -106,6 +118,25 @@ function errorHandler(error) {
 }
 
 /**
+ * Task: `build-pl`
+ * Builds the bundled pattern library (if necessary)
+ */
+gulp.task('build-pl', function (done) {
+    if (buildPatternLibrary) {
+        exec('{ cd ' + patternLibrary + ' && npm install ;} 2>&1', function (error, stdout) {
```
rjmackay: Would it make sense to simply include the custom PL via a submodule? or do you thinking doing a separate npm install is useful?

tuxpiper:
Closing this for after webpack

PR #330, opened 2016-09-06
Further code reorg: Group into common, settings, main and move templates side by side 

rjmackay:
- Group code into common, settings, main
- Move templates into app alongside directives/controllers

Test these changes by:
- gulp test

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/330)


willdoran:
@rjmackay I don't see anything that breaks obviously since the changes are consistent. It should may be go in before the next sprint and then get QAed. Because so many things change everything needs to be tested BUT because the changes are fairly straightforward the chances of a break are much smaller.

rjmackay:
I'm hoping to do a switch to webpack or something that allows `require()`ing the templates.. which should make it much more obvious if I've got a path wrong somewhere.

PR #331, opened 2016-09-07
Improve csv after import guidance and messaging and experience

willdoran:
- This address the issue of unclear guiding of user through final part of csv import
- We add a final import screen that informs the user of the process and advises them of their options
- Updated and improved csv success slider to give greater clarity and information

Test these changes by:
- [ ] Begin the import process for a small csv, after you click Finish Import you should see a fourth screen telling the user that the import is in process, this screen should quickly update to tell the user that the import has complete and a slider should appear
- [ ] All buttons on the fourth screen should be clickable and should lead to the correct location expected based on the button clicked
- [ ] All buttons in the success slider should be clickable and lead to the expected location/result based on the button
- [ ] Attempt to import a very large csv, on the fourth screen, wait until the import has finished, the screen should update and a slider should appear
- [ ] Attempt to import a very large csv, once on the fourth screen click away to another view Map or Timeline for example, after several moments once the import has completed a slider should appear
- [ ] Attempt to import a very large csv, once on the fourth screen click import another csv, proceed to import another large csv - make sure this csv has a different name from the first, once you reach the fourth screen wait: if you have not already received a success slider for the first csv then you should, after that you should receive a success slider for the second csv once its import complete.
Notes: The progress bar is purely figurative it is not live and will not change

Fixes ushahidi/platform#1349

Ping @ushahidi/platform

…larified process

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/331)


app/settings/data-import/data-after-import.directive.js
```diff
@@ -0,0 +1,26 @@
...
+    $location,
+    Notify,
+    Features
+) {
+    return {
+        restrict: 'A',
+        link: function ($scope, $element, $attrs) {
+            $scope.importComplete = false;
+
+            $rootScope.$on('event:import:complete', function (event, args) {
```
jasonmule: Any way of avoiding the use of a global event here?

app/settings/data-import/data-mapper.directive.js
```diff
@@ -103,16 +105,24 @@ function (
...
-                                Notify.success('notify.data_import.csv_mappings_set', {processed: processed, errors: errors});
+                                Notify.importComplete(
+                                {
+                                    processed: processed,
+                                    errors: errors,
+                                    form_name: $scope.form.name,
+                                    filename: csv.filename
+                                });
+
+                                $rootScope.$emit('event:import:complete', {form: $scope.form, filename: csv.filename});
```
jasonmule: Can we avoid using `$rootScope.$emit` here?

app/common/notifications/notify.service.js
```diff
@@ -14,7 +14,8 @@ function Notify(_, $q, $rootScope, $translate, SliderService, ModalService) {
         confirm: confirm,
         confirmModal: confirmModal,
         confirmDelete: confirmDelete,
-        limit: limit
+        limit: limit,
+        importComplete: importComplete
```
rjmackay: I don't think this should life in the `notify` service. It isn't a generic notification.. it only relates to importing, so it should live in a service tied to the importer.

app/common/locales/en.json
```diff
@@ -760,6 +760,13 @@
         }
     },
     "data_import" :{
+        "see_imported_posts" : "See imported posts",
+        "import_another_csv" : "Import another CSV file",
+        "import_complete" : "Your import is complete",
+        "complete_info" : "The data from your CSV spreadsheet, <em>{{filename}}</em>, was successfully imported into your <a href=\"/settings/survey/{{form_id}}\">{{form_name}}</a> survey.",
+        "import_another" : "Import another CSV file",
+        "nearly_complete" : "Your import is nearly complete",
+        "finish_info" : "The data from your CSV spreadsheet is still being processed. You're welcome to leave this page or wait until the process is complete. Either way, we'll notify you when your data is ready.",
```
rjmackay: > You're welcome to leave this page

This isn't really true. It suggests if they close the browser window their import will still complete, but it won't.

willdoran: The import will at this stage still complete if they close the browser but they won't receive a notification. The text comes from 
https://github.com/ushahidi/platform-pattern-library/blob/architecture/assets/templates/layouts.settings-import-complete.hbs

What's a better phrasing? Navigate around the application, or navigate to another part of the application? If they stay inside the application they will receive a notification.

app/common/notifications/notify.service.js
willdoran: Hmmm, I can create a service that draws from notify.service.js but should it live next to the importer in settings? since it would be import.notify.service.js, we could later extend notify for specific elements that require special notifications.

app/common/locales/en.json
rjmackay: I guess thats true.. though I'm not 100% sure. ie. if the browser ditches the HTTP request to the server.. does the server abandon the process or just let it finish?
Assuming you're correct, we can probably leave it as is but just drop the `we'll notify you when your data is read` part.

rjmackay:
Review status: 0 of 10 files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

*[app/common/notifications/notify.service.js, line 18 \[r1\]](https://reviewable.io:443/reviews/ushahidi/platform-client/331#-KR4XUpMs-hzqIFjkdmr-r1-18:-KRC8tgnTxh62PqDtDat:b1xq08v) ([raw file](https://github.com/ushahidi/platform-client/blob/60796bf59cd068e66f90028e60dc028caca2c4e0/app/common/notifications/notify.service.js#L18)):*
<details><summary><i>Previously, willdoran (will) wrote…</i></summary>
> Hmmm, I can create a service that draws from notify.service.js but should it live next to the importer in settings? since it would be import.notify.service.js, we could later extend notify for specific elements that require special notifications.
</details>
It should definitely live alongside the importer code. I think this could just be `data-import.service.js` then you can use it for any other service code related to the imports. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ushahidi/platform-client/331)*

app/common/locales/en.json
tuxpiper: re: interrupting script execution when the underlying connection is closed. It depends, see: http://stackoverflow.com/questions/1280291/does-php-execution-stop-after-a-user-leaves-the-page

willdoran: So, angularjs $resource does not automatically abort requests, it can be forced to on either a timeout or a promise resolution. In this case, the API won't know the client hungup until it sends the response.

This seems to be a somewhat contentious issue as some people argue that promises should never be cancellable. Looking at the angularjs code it does appear that they only send an abort if a timeout is defined. At least, this is what I've found in the code and they have no documentation that suggests otherwise.

I would argue that as much as an async call is still a connection between two endpoints it is conceptually a separation where the response is not dependent on the requester not hanging up.

So not to prolong this - I'll commit and merge and this can be revisisted - but the "we will notify you bit" is I think important though perhaps it should be "If you remain in the app we will send you a notification to say the import has completed"

rjmackay: > angularjs $resource does not automatically abort requests

I wasn't suggesting that. However the browser will abort a request if the tab is closed, so then its a question of whether PHP-FPM+nginx will kill the PHP process handling the import.
TBH this add to the list of things we should be buffering behind a job queue instead.

tuxpiper: +1 for job queue . As a related experience, I've completed a not very big import of a CSV for 1000 ocean actions (~250 posts). I had to do two batches since it timed out in the middle (504 Gateway Timeout). php-fpm did abort the request. 

While this is moved to a queue, it may be a reason to revisit server timeout settings.

PR #334, opened 2016-09-14
Build with webpack

rjmackay:
- Load templates with require()
- Build JS, HTML and CSS with webpack
- Cachebusting via webpack
- Enable use of ES6 going forward
- Replace old gulpfile

Todo
- [x] Separate built CSS into own file
- [x] Load backend url from .env file
- [x] Figure out how to generate config.js on build
- [x] Restore transifex download on build/dist
- [x] Restore font loading
- [x] Fix test runner
- [x] Exclude moment locale files and load dynamically

Issues: 
- Transifex locales are never available for dev

Test these changes by:
- `gulp dev`
- `gulp dist`

Fixes ushahidi/platform#1320

Ping @ushahidi/platform

---

This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/334)


rjmackay:
@willdoran @jasonmule I'm curious what your thoughts are on using webpack.
Its seems good but its also an "all in" solution ie. we have to do most everything with webpack.

willdoran:
I'm not sure, what does it give us over Browserify? Are there any optimisations it facilitates? I'm a little concerned by how heavy it appears to be. If we were going to push in that direction and particularly gain ES6 should we look at [System.js](https://github.com/systemjs/systemjs) as well?

As a side note, I think someone posted a page speed test of an ushahidi deployment. I just ran it on 1000ocean.ushahidi.com and there are quite a few simple optimisations that would help speed up the page dramatically.

rjmackay:
I agree it is heavy.. that said I think our collection of custom gulp tasks was getting fairly messy to maintain too. This reduces most of that to a few config files. The primary difference here is approach: gulp+browserify can do all the same things, but we have to create+maintain gulp tasks for each of them, browserify is design to handle the entire app build (JS and everything else) and optimizations.
Trying to incorporate best practices with webpack felt easier.. but there was a heavy lift to get started. 

The dev experience with this change is much nicer.. but a lot of that is actually down to using BrowserSync as a dev server, which we could do with browserify.

Vs browserify:
- Less homebrew gulp tasks. 
- Handles compiling sass, css, svg sprites. Not just JS.
- Makes it fairly trivial to split built files into multiple files.
- Lazy JS loading .. though with angular this requires a little extra wrangling
- Handles auto injecting JS urls into the HTML. Primarily useful because we can also add a hash to the JS file name for cachebusting.
- Next version of webpack does treeshaking (automatically dropping unused modules from the final bundle)

Compared to system JS I think webpack has a much larger/ more mature eco system. If we were to look at other build tools beyond these, I would look at [RollUp](http://rollupjs.org/) which is focussed on performance. However webpack 2 brings many of the RollUp features anyway.

rjmackay:
This should be entirely working now, though it ended up huge :(

There are small changes to most of the test and files through out, but usually the same change everywhere.

One thing I haven't decided is if I should change the build dir from `dist` to `server/www`
`dist` makes more sense but `server/www` keeps backwards compatibility for existing deploys.

app/main/activity/activity.html
```diff
@@ -99,15 +99,15 @@ <h1 class="mode-context-title"><a href="/" ng-bind="nav.site.name"></a></h1>
 
     	<!-- unavailable view -->
         <post-view-unavailable ng-hide="isActivityAvailable" view="'nav.activity' | translate"></post-view-unavailable>
-        <!-- <img ng-hide="isActivityAvailable" src="/img/activity-page.png" style="opacity: 0.4; filter: blur(1px);" /> -->
+        <!-- <img ng-hide="isActivityAvailable" src/img/activity-page.png" style="opacity: 0.4; filter: blur(1px);" /> -->
```
willdoran: Thiis probably not worth changing but the =" is missing after src

willdoran:
There does seem to be an error on jshint

rjmackay:
This is ready to go, however I'd like to do a production deploy with embed etc before landing it.

rjmackay:
Without webpack:
screenshot 2016-11-08 13 05 23

With webpack:
screenshot 2016-11-08 13 05 16

rjmackay:
@jshorland I'd like us to do a broad QA pass after this lands.. its fairly wide reaching so need to double check we haven't broken things. Mustang QA checklist is probably a good start point?

rjmackay:
Started a QA checklist for this.. https://docs.google.com/document/d/1NqeLuVfT54VobZDdUYwCze6vd0ph7lbVYLultBr3IVI/edit
1 issue so far:
- If you load the timeline then switch to map, markers never render.

rjmackay:
@jshorland I started working my way through a QA checklist
https://docs.google.com/document/d/1NqeLuVfT54VobZDdUYwCze6vd0ph7lbVYLultBr3IVI/edit

I've fixed the map issues in https://github.com/ushahidi/platform-client/pull/416
Could you give this a look before I land it?

PR #341, opened 2016-09-21
Updating data import interface and functionality - next change valida…

willdoran:
- Changes data-import code to remove data retriever, data mapper and data configure
- Changes UX for csv import significantly
- Allows user to set status, map columns
- Updates post edit to provide a warning in the case that the post is in an invalid state

Test these changes by:
-

Fixes ushahidi/platform# .

Ping @ushahidi/platform

…tion when task is required and status is published

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/341)


app/common/notifications/notify.service.js
```diff
@@ -27,7 +27,7 @@ function Notify(_, $q, $rootScope, $translate, SliderService, ModalService) {
 
     function error(errorText, translateValues) {
         function showSlider(errorText) {
-            SliderService.openTemplate('<p>' + errorText + '</p>', 'warning', 'error');
```
rjmackay: What are we changing here and why?

app/main/posts/modify/post-editor.html
```diff
@@ -33,15 +33,15 @@ <h1 class="mode-context-title">{{post.form.name}}</h1>
 
           <div class="form-sheet">
               <div class="post-band" style="background-color: {{form.color}}"></div>
-              <div class="form-field title form-field-adapt init"
+              <div class="form-field title init required"
```
rjmackay: Are these changes part of the CSV work? or from somewhere else?

app/settings/categories/categories-edit.html
```diff
@@ -44,7 +44,7 @@ <h1 class="mode-context-title" ng-show="!category.id" translate>app.add_category
                        <input type="text" placeholder="{{'category.editor.name' | translate}}" ng-minlength="3" ng-maxlength="150" ng-model="category.tag" required>
                    </div>
 
-                    <div class="form-field form-field-adapt">
+                    <div class="form-field">
```
rjmackay: This seems unrelated?

app/main/posts/modify/post-value-edit.html
```diff
@@ -8,7 +8,6 @@
         'error': form['values_' + attribute.key].$invalid && form['values_' + attribute.key].$dirty,
         'success': ! form['values_' + attribute.key].$invalid && form['values_' + attribute.key].$dirty,
         'required': attribute.required,
-        'form-field-adapt': isText(attribute) || isTextarea(attribute),
```
rjmackay: unrelated?

app/settings/data-import/data-import.directive.js
```diff
@@ -1,42 +1,302 @@
...
+    ImportNotify,
+    Features,
+    _
 ) {
     return {
         restrict: 'A',
         link: function ($scope, $element, $attrs) {
-            $scope.formId;
+            $scope.selectedForm;
+            $scope.activeStep = 1;
```
rjmackay: Am I right in thinking all the steps are now a single directive?

app/settings/data-import/data-import.directive.js
```diff
@@ -1,42 +1,302 @@
...
+                    Notify.error('notify.data_import.form_missing');
+                    return;
+                }
                 var formData = new FormData();
                 formData.append('file', $scope.fileContainer.file);
-                formData.append('form_id', $scope.formId);
+                formData.append('form_id', $scope.selectedForm.id);
 
                 DataImportEndpoint.upload(formData)
                 .then(function (csv) {
```
rjmackay: I think moving this to a named function might make the logic much clearer ie.
```
DataImportEndpoint.upload(formData)
                  .then(loadStep2());
```

app/settings/data-import/data-import.directive.js
```diff
@@ -1,42 +1,302 @@
...
-                    $location.url('/settings/data-mapper/' + $scope.formId + '/' + csv.id);
+                    $scope.stepOneComplete = true;
+                    $scope.activeStep = 2;
+                    $scope.csv = csv;
+
+                    // Retrieve tasks and attributes
+                    $q.all([
+                        FormStageEndpoint.get({form_id: $scope.selectedForm.id}).$promise,
+                        FormAttributeEndpoint.get({form_id: $scope.selectedForm.id}).$promise
+                    ]).then(function (results) {
```
rjmackay: similar here
```
$q.all([
                        FormStageEndpoint.get({form_id: $scope.selectedForm.id}).$promise,
                        FormAttributeEndpoint.get({form_id: $scope.selectedForm.id}).$promise
                    ]).then(doSomething);
```

.. I'm not clear what `doSomething` actually is here.

app/settings/data-import/data-import.directive.js
```diff
@@ -1,42 +1,302 @@
...
+                            }
+                        });
+                    });
                 }, function (errorResponse) {
                     Notify.apiErrors(errorResponse);
                 });
             };
+
+            // Check for missing required fields and return the missing fields
+            $scope.checkRequiredFields = function (fields) {
```
rjmackay: Does this really need to be on `$scope`? it seems like its called from elsewhere in the directive code

app/settings/data-import/data-import.directive.js
```diff
@@ -1,42 +1,302 @@
...
+                        missing.push(
+                            $scope.required_fields_map[field]
+                        );
+                    });
+                }
+
+                return missing;
+            };
+
+            $scope.remapColumns = function () {
```
rjmackay: same here - is this called from the template? or could we move it off $scope?

app/settings/data-import/data-import.directive.js
```diff
@@ -1,42 +1,302 @@
 module.exports = [
```
rjmackay: This directive is now really hard to read as it has *all* the logic in it. Could you do a pass updating the coding style? ie. use named functions, only bind to scope when needed.

While an external service may not be essential, I think the complicated data munging steps might still be better living elsewhere on a DataImportService or similar. Though just the coding style change might be enough..

app/common/notifications/notify.service.js
willdoran: The error message was disappearing after the timeout, this seems problematic because by the time the user has half read and processed the issue it vanishes. It also means that they won't be able to grab error messages that they see and put them into issues.

app/main/posts/modify/post-editor.html
willdoran: They seem to be issues caused by the newer PL, I noticed them when testing the posts after import. I can pull them out into their own PR though.

app/settings/data-import/data-import.directive.js
willdoran: Ok, I'll do that now

app/settings/data-import/data-import.directive.js
willdoran: Yep

app/common/notifications/notify.service.js
rjmackay: Ok that makes sense.

app/common/directives/color-picker.js
```diff
@@ -29,7 +29,7 @@ module.exports = [
         return {
             restrict: 'E',
             replace: true,
-            templateUrl: 'templates/common/color-picker/color-picker.html',
+            templateUrl: 'templates/common/directives/color-picker.html',
```
rjmackay: Can you just push this fix directly onto develop?
(aside: I picked up a bunch of errors like this in the webpack branch because using require() means the build checks if paths actually exist)

app/main/posts/modify/post-editor.html
rjmackay: Ideally just make these fixes direct to develop I think. For now though we can just go ahead with this PR

app/settings/data-import/data-import.directive.js
rjmackay: Much better thanks!

app/common/directives/color-picker.js
willdoran: Yep, already done.

app/main/posts/modify/post-editor.html
willdoran: cool

PR #354, opened 2016-10-04
1387 embed

willdoran:
- Allow ushahidi map to be embed within an iframe
- Added embed section to general settings page, with example code for drop in, it includes the sites deployment url in the example. 
- The above needs a more detailed message and default height width values

Test these changes by:
- Create a local html page containing the embed code
- [ ] Load the page in your browser
- [ ] Check that pages and logos display accordingly with:
- [Map](http://preview.ushahidi.com/platform-pattern-library/f-embeds/assets/html/5_layouts/embed-map.html)
- [Post detail](http://preview.ushahidi.com/platform-pattern-library/f-embeds/assets/html/5_layouts/embed-post-detail.html)
- [Post edit](http://preview.ushahidi.com/platform-pattern-library/f-embeds/assets/html/5_layouts/embed-post-add.html)
- [ ] Ensure that viewing posts works
- [ ] Ensure that adding posts works
- [ ] Ensure that Ushahidi logo appears in correct location depending on size
- [ ] Ensure map zoom controls are visible
- [ ] Ensure that filtering by survey works

Fixes ushahidi/platform#1387

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/354)


rjmackay:
It looks like this accidentally covered some of ushahidi/platform#1388 and ushahidi/platform#1389 too

app/common/global/event-handlers.js
```diff
@@ -1,16 +1,25 @@
...
 ) {
     // Setup PL layout and switching function
     $rootScope.globalLayout = 'layout-a';
     $rootScope.setLayout = function (layout) {
         $rootScope.globalLayout = layout;
     };
-
+    // Set embed mode
+    $rootScope.embedUrl = Util.deploymentUrl();
+    $rootScope.globalEmbed = ($window.self !== $window.top) ? true : false;
```
rjmackay: I really dislike that we're adding more stuff to `$rootScope` .. 
rootScope is inherited by all child scopes, and then dirty checked in each digest cycle. ie. more things on rootScope, slower digest. It'd make sense for stuff like site name, if we inherited that everywhere. But we're mostly only checking `globalEmbed` within directives where we have to inject rootScope anyway.

We can keep this for now because I'm not sure what the right solution is.. but if you're got a better idea we should tend towards that over time.

app/main/posts/detail/detail.html
```diff
@@ -155,5 +155,6 @@ <h3 class="listing-heading" translate="app.tasks">Tasks</h3>
 
     <post-add-form post="post" ng-if="!post.form && post.allowed_privileges.indexOf('update') !== -1"></post-add-form>
     <post-messages post="post" ng-if="post.contact.id"></post-messages>
+    <ush-logo></ush-logo>
```
rjmackay: Does this just float on every view when embedded? Is it possible to just add it once to index.html?

app/common/directives/embed-control.directive.js
```diff
@@ -0,0 +1,18 @@
...
+function EmbedControlDirective() {
+    return {
+        restrict: 'A',
+        controller: EmbedControlController
+    };
+}
+
+EmbedControlController.$inject = ['$scope', '$element', '$attrs', '$rootScope', '_'];
+function EmbedControlController($scope, $element, $attrs, $rootScope, _) {
+    if ($rootScope.globalEmbed && !$attrs.embedOnly) {
```
rjmackay: It might be better to simply call this directive `embed-only` and always expect a true or false value.

app/common/directives/ush-logo.html
```diff
@@ -0,0 +1,118 @@
+<a embed-control embed-only=true href="https://www.ushahidi.com" class="ushahidi-bug" target="_blank">
+  <svg class="ushahidi-logo" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 273.5 266.5" xml:space="preserve">
```
rjmackay: We should already have the SVG for this in an include somewhere, we use it in the mode-bar

app/main/posts/detail/detail.html
willdoran: Unfortunately, it can't live at the top as the css is hierarchy dependent so it has to go here. I asked @brandonrosage about the ability to move it and he said that he can look into it but that it relates to a core approach(I may be mangling and completely misrepresenting what he meant).

app/common/directives/ush-logo.html
willdoran: ah we do, I'll swap it out

rjmackay:
@willdoran this all looks good. I had 1 last comment about moving the embed check off rootScope entirely though

PR #356, opened 2016-10-05
Recentre on popup

willdoran:
- Recentre map on marker on popupopen

Test these changes by:
- Ensure that map recentres when clicking on map marker

Fixes ushahidi/platform#1418

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/356)

### DIFF
--- a/app/common/services/maps.js
+++ b/app/common/services/maps.js
@@ -267,6 +267,11 @@ function (
 
                 this.map().then(function (map) {
                     map.addLayer(markers);
+                    map.on('popupopen', function (e) {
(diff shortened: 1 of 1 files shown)


willdoran:
@rjmackay the autopan is on by default but it doesn't seem to handle this, from my research this is the way most people recentre when there is a popup. Let me know what you prefer and I can adjust.

rjmackay:
Ok. I'll land this and send to Jess for QA / reviewing the behaviour

PR #374, opened 2016-10-19
1434 - image upload

willdoran:
- Adds new image upload system
- Changes upload flow

Test these changes by:
- [ ] On a post with an image field, add an image and click save
- [ ] Edit the above post to delete the image, you should be asked to confirm\

After deleting image the following should be tested:
  - [ ] Save post
  - [ ] Add a new image and save post

After creating a post with an image the following should be tested:
  - [ ] Replace current image and cancel
  - [ ] Replace current image and save

Caption:
  - [ ] Test saving an image without a caption then editing and adding one
  - [ ] Test editing a caption only
  - [ ] Test changing image and caption
  - [ ] Test editing caption then clicking cancel

Fixes ushahidi/platform#1434

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/374)


app/main/posts/modify/post-editor.directive.js
```diff
@@ -118,6 +121,14 @@ function PostEditorController(
 
             // Initialize values on post (helps avoid madness in the template)
             attributes.map(function (attr) {
+                // Create associated media entity
+                if (attr.input === 'upload') {
+                    var media = {};
+                    if ($scope.post.values[attr.key]) {
+                        media = $scope.post.values[attr.key][0];
+                    }
+                    $scope.medias[attr.key] = media;
```
rjmackay: What is in `media` at this point? just an ID? or an object?

This seem problematic because we're not keeping 2 copies of the media items around.. could we achieve the same thing just by grabbing a list of all the 'upload' type attributes?

app/main/posts/modify/post-media.directive.js
```diff
@@ -17,84 +19,44 @@ function (
...
-                        $scope.mediaId = response.data.id;
-
-                        // We found a file so change the state of parent form
-                        formCtrl[$scope.name].$setDirty();
-                    }, function (error) {
-                        Notify.apiErrors(error);
+            $scope.deleteMedia = function (mediaId) {
+                // Mark for deletion
+                Notify.confirmDelete('notify.notification.delete_confirm').then(function () {
+                    MediaEditService.deleteMedia(mediaId).then(function () {
```
rjmackay: Shouldn't this be queued until the post is saved?

app/main/posts/modify/post-media.directive.js
```diff
@@ -17,84 +19,44 @@ function (
...
+                return $scope.media.dataURI || $scope.media.id;
+            };
+
+            $scope.showDelete = function () {
+                return $scope.media.id;
+            };
 
             if ($scope.mediaId) {
                 MediaEndpoint.get({id: $scope.mediaId}).$promise.then(function (media) {
                     $scope.media = media;
```
rjmackay: I suspect this is likely to hit problems with the parent directive not picking this new object up..
You need to either 

1. pass the entire `medias` collection over the directive boundary, or...
2. make the post-edit directive a requirement for this one, pass its controller into this one and provide an interface there to add more media items

rjmackay:
Review status: 0 of 15 files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[app/main/posts/modify/media-edit.service.js, line 85 at r3](https://reviewable.io:443/reviews/ushahidi/platform-client/374#-KUPX-0-ocnu3f4liZt5:-KUPX-0-ocnu3f4liZt6:b-j5jd6l) ([raw file](https://github.com/ushahidi/platform-client/blob/06fb56881669404d7e111e42fb8a7ff6421af6eb/app/main/posts/modify/media-edit.service.js#L85)):*
> ```JavaScript
>                 }
> 
>                 $http.post(
> ```

I think if you return the promise here then you can just continue the chain like

```

this.deleteMedia(media.id)
.then(function () {
 
            //... 
 
            return $http.post(blah );
})
.then(function (response) {
                    media.id = response.data.id;
                    deferred.resolve(media);
})
.catch(function (error) {
           Notify.apiErrors(error);
                    deferred.reject(error);
});

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ushahidi/platform-client/374)*
<!-- Sent from Reviewable.io -->

File: app/main/posts/modify/post-media.directive.js (hunk 3 above)
Comment (willdoran) [inline app/main/posts/modify/post-media.directive.js]: So the flow is as follows:

1. Retrieve all media attributes and associated them with their key
2. For each media attribute create a media-value directive instances, pass in the media map reference
3. Update/construct object map to medias
4. On save, loop through medias items, taking appropriate action if changed

At the moment, it functions correctly, passing in the media map follows the same pattern as for the other post edit values. 

Could you elucidate why or when you think it'll break? I'm not sure I follow.

File: app/common/services/endpoints/MediaEndpoint.js [hunk 4]
```diff
@@ -19,7 +19,7 @@ function (
             }
         },
         update: {
-            method: 'POST'
```
willdoran: Nope

app/common/directives/file-upload.directive.js
```diff
@@ -4,18 +4,27 @@ function FileUpload() {
     return {
         restrict: 'E',
         templateUrl: 'templates/common/directives/file-upload.html',
+        replace: true,
```
willdoran: I think it's ok, I checked and apparently it's not being removed prior to 2.x

I would argue to keep this because it has the nice effect of allowing the id to fall directly through to the input. Which is neater than having to pass the id as a param but I won't fight particularly hard for it.

app/main/posts/modify/post-editor.directive.js
willdoran: No this is just an id. It will eventually be replaced by the full media object.

willdoran:
@rjmackay I have another question which I'm not sure we address well. 

I have 4 images on my post, I hit save, 1 and 2 and 3 are fine, but 4 is too big. At the moment, it fails and everything stops, and the user is warned. 

But this doesn't interact with the validation system really at all. If feels like the validation system doesn't really hang together all that well. It would be really nice to fix it in a sprint.

app/main/posts/modify/post-media.directive.js
rjmackay: hrm ok you make a good point. If we were to delay delete we'd probably have to include an 'undo' function and then scope creeps larger.

I don't like that this means a 'cancel' on post edit doesn't actually cancel everything because of this.. but I'm ok with leaving it for now.

rjmackay:
> I have 4 images on my post, I hit save, 1 and 2 and 3 are fine, but 4 is too big. At the moment, it fails and everything stops, and the user is warned.

I think ideally we just save the post + other images but warn the user 1 image couldn't be uploaded. Its not totally consistent with other views, but the alternative is a pretty terrible experience where I need to re-upload things.

willdoran:
Ok cool, I'll have to change the logic a little and may be the error message, so a fail on upload would still resolve the promise rather than reject but we'll show the error and go on with the save.

app/main/posts/modify/post-media.directive.js
rjmackay: Maybe I'm wrong.. I've never 100% figure out how angular handles this.. and going back to things I've read about angular scope and inheritance I can't find support material

My thinking was that angular passed a JS object reference across the directive boundary.. and updates to the JS object would propagate back to the parent because it was literally the same object, where actually doing `scope.media = ...` would overwrite the reference entirely. **but** I think I'm just wrong. Googling suggests isolate scope's two way binding is implemented with a watch..

TLDR: ignore me, please continue :)

app/main/posts/modify/post-editor.directive.js
rjmackay: For consistency, could we just populate it with an empty object and leave that to be populated by the media directive? as far as I can see the id is never used from here

PR #375, opened 2016-10-19
Fix markup on collection and saved search context menus

rjmackay:
- Fix markup on meta data
- Pull context menu and edit button out of metadata div
- Display actual visibility state, rather than always public

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/375)


willdoran:
LGTM

PR #381, opened 2016-10-25
Embed fixes

willdoran:
- Address embed feedback issues

Fixes ushahidi/platform#1458
Fixes ushahidi/platform#1459

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/381)

### DIFF
--- a/app/common/directives/embed-only.directive.js
+++ b/app/common/directives/embed-only.directive.js
@@ -11,8 +11,9 @@ function EmbedOnlyDirective() {
 EmbedOnlyController.$inject = ['$scope', '$element', '$attrs', '$rootScope', '_', '$window'];
 function EmbedOnlyController($scope, $element, $attrs, $rootScope, _, $window) {
     var globalEmbed = ($window.self !== $window.top) ? true : false;
-    if (globalEmbed) {
-        $rootScope.setLayout('layout-embed');
+    if (globalEmbed && $rootScope.globalLayout.indexOf('layout-embed') < 0) {
+        // If layout-embed is not set append it to the global layout class
+        $rootScope.setLayout($rootScope.globalLayout + ' layout-embed');
     }
     if (globalEmbed && ($attrs.embedOnly === 'false')) {
         $element.addClass('hidden');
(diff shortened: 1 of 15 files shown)


app/common/directives/embed-only.directive.js
```diff
@@ -11,8 +11,9 @@ function EmbedOnlyDirective() {
 EmbedOnlyController.$inject = ['$scope', '$element', '$attrs', '$rootScope', '_', '$window'];
 function EmbedOnlyController($scope, $element, $attrs, $rootScope, _, $window) {
     var globalEmbed = ($window.self !== $window.top) ? true : false;
-    if (globalEmbed) {
-        $rootScope.setLayout('layout-embed');
+    if (globalEmbed && $rootScope.globalLayout.indexOf('layout-embed') < 0) {
```
rjmackay: these seems fragile. Should we just be always adding the `layout-embed` class when we're in an embed? regardless of which view.

rjmackay: .. or actually if we just always use the `layout-class` directive rather than calling `setLayout` won't this take care of itself?

app/main/posts/views/mode-context-form-filter.directive.js
```diff
@@ -45,6 +48,13 @@ function ModeContextFormFilter($scope, FormEndpoint, PostEndpoint, $q, _, $rootS
         });
     }
 
+    function externalOpenDropDown() {
```
rjmackay: You shouldn't need to manually trigger a click handler. dropdown directive has an `is-open` property. Add that to the dropdown tag and then just set the variable true or false.

app/common/directives/embed-only.directive.js
willdoran: Yep we do and yeah that's actually a better solution, I'l change it

app/main/posts/views/mode-context-form-filter.directive.js
willdoran: ah cool, even better

rjmackay:
LGTM

PR #384, opened 2016-10-27
Adding share functionality for client

willdoran:
- Adds share menu for client, which includes, twitter facebook, embed, and export

Test these changes by:
- [ ] On the map or timeline views, click Share, the menu should open with 4 options
- [ ] Click share on Facebook, should allow you to share a link to the deployment on facebook
- [ ] Click share on Twitter, should allow you to share a link to the deployment on twitter
- [ ] Click share on embed, should show you the html code need for sharing
- [ ] Paste the above code into a html file on your local machine and test that the embed works
- [ ] Click share on Export, should allow you to export the current posts
- [ ] Set different filters then click on export the correct posts based on the filters should be exported only

Fixes ushahidi/platform#1465

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/384)

### DIFF
--- a/app/common/services/util.js
+++ b/app/common/services/util.js
@@ -1,12 +1,17 @@
 module.exports = [
     '_',
     'CONST',
+    '$window',
 function (
     _,
-    CONST
+    CONST,
+    $window
 ) {
 
     var Util = {
+        currentUrl: function () {
+            return $window.location.href;
+        },
         url: function (relative_url) {
             return CONST.BACKEND_URL + relative_url;
         },
--- a/app/main/posts/views/share/share-menu.html
+++ b/app/main/posts/views/share/share-menu.html
@@ -28,7 +28,7 @@
 										<div class="listing-item-image">
 												<img class="icon" src="../../img/twitter.png" alt="">
 										</div>
-										<h2 class="listing-item-title"><a ng-href="https://twitter.com/home?status={{shareUrl}}">Twitter</a></h2>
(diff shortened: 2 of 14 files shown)


app/common/services/util.js
```diff
@@ -1,12 +1,17 @@
...
+    '$window',
 function (
     _,
-    CONST
+    CONST,
+    $window
 ) {
 
     var Util = {
+        currentUrl: function () {
```
rjmackay: It looks like the `deploymentUrl` is going to end up wrong where its used in site settings too.. could you fix that up too?

app/main/posts/views/share/share-menu.html
```diff
@@ -0,0 +1,78 @@
...
+										</h2>
+								</div>
+						</div>
+
+						<div class="listing-item">
+								<div class="listing-item-primary">
+										<div class="listing-item-image">
+												<img class="icon" src="../../img/twitter.png" alt="">
+										</div>
+										<h2 class="listing-item-title"><a ng-href="https://twitter.com/home?status={{shareUrl}}">Twitter</a></h2>
```
rjmackay: Shouldn't this use shareUrlEncoded instead?

willdoran: Actually for twitter it appears to be ok unencoded, i'll test with encoded and see if they parse

PR #385, opened 2016-10-27
1462-collection and active filter fix

willdoran:
- Changed active filter array comparison check
- Add archived filter to collection default filters

Test these changes by:
- [ ] Create or import a set of posts with archived posts
- [ ] On viewing collection archived posts should be visible
- [ ] On all map or timeline views, when you click on a filter it should appear in the active filters on the left hand side in the mode context

Fixes ushahidi/platform#1462

Ping @ushahidi/platform

…r not appearing

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/385)

### DIFF
--- a/app/main/posts/collections/collections-controller.js
+++ b/app/main/posts/collections/collections-controller.js
@@ -39,11 +39,11 @@ module.exports = [
 
         // Extend filters, always adding the current collection id
         var extendFilters = function (filters) {
-            filters = angular.copy(filters, { set : [] });
+            filters = angular.copy(filters, { set : []});
             filters.set.push(collection.id);
-
             //Ensure that ALL posts are visible under collections
-            filters.status.push('archived');
+            // Set default collection status filters
+            filters.status = ['published', 'draft', 'archived'];
             return filters;
         };
 
--- a/app/main/posts/collections/collections-controller.js
+++ b/app/main/posts/collections/collections-controller.js
@@ -39,12 +39,13 @@ module.exports = [
 
         // Extend filters, always adding the current collection id
         var extendFilters = function (filters) {
-            filters = angular.copy(filters, { set : []});
+            //filters = angular.copy(filters, { set : []});
+            filters.set = [];
             filters.set.push(collection.id);
             //Ensure that ALL posts are visible under collections
             // Set default collection status filters
             if (_.has(filters, 'status')) {
-                filters.status.push('archived');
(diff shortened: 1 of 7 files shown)


app/main/posts/collections/collections-controller.js
```diff
@@ -39,8 +39,17 @@ module.exports = [
 
         // Extend filters, always adding the current collection id
         var extendFilters = function (filters) {
-            filters = angular.copy(filters, { set : [] });
+            //filters = angular.copy(filters, { set : []});
+            filters.set = [];
             filters.set.push(collection.id);
+            //Ensure that ALL posts are visible under collections
+            // Set default collection status filters
+            if (_.has(filters, 'status')) {
```
rjmackay: This is problematic because it will overwrite the status filter on every update, not just the first load.

Need to split this into 1. function to always add collection to filters 2. setting the initial filters to include archived.. maybe just use `PostFilters.setFilters({ status: ['archive', 'draft', 'published'] });` rather than `PostFilters.clearFilters()`

app/main/posts/views/filters/active-filters.directive.js
```diff
@@ -48,6 +48,8 @@ function ActiveFilters($translate, $filter, PostFilters, _, TagEndpoint, RoleEnd
             var activeFilters = angular.copy(filters);
             rawFilters = angular.copy(filters);
 
+            // Remove set filter as it is only relevant to collections and should be immutable in that view
+            delete activeFilters.set;
```
rjmackay: this is nicer than expected.. I forgot the collections filter doesn't exist in the dropdown any more :)

app/main/posts/views/post-filters.service.js
```diff
@@ -113,7 +113,7 @@ function PostFiltersService(_, FormEndpoint) {
                     return true;
                 }
                 // Is an array with all the same elements? (order doesn't matter)
-                if (_.isArray(defaults[key]) && _.difference(defaults[key], value).length === 0) {
+                if (_.isArray(defaults[key]) && _.difference(value, defaults[key]).length === 0) {
```
rjmackay: Regression test please

app/main/posts/views/filters/active-filters.directive.js
willdoran: yep

app/main/posts/views/post-filters.service.js
willdoran: Adding now

PR #397, opened 2016-11-02
Fix timezone issues with post date

rjmackay:
- Vastly simplify post-datetime-value directive
- Only save when date/time picker closes
- Just let datepicker handle setting the model
- Remove timezone from site settings
- Format post date time fields properly

Test these changes by:
- Add a new posts and check it gets current date/time as post date
- Edit an existing post and change the post date. Confirm the correct date + time (shown on hover) is saved
- Confirm order of posts in list is based on post date
- Add a datetime field to a post. Edit it and confirm it saved and display correctly

Fixes ushahidi/platform#1362

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/397)

### DIFF
--- a/app/main/posts/detail/post-value.directive.js
+++ b/app/main/posts/detail/post-value.directive.js
@@ -1,4 +1,4 @@
-module.exports = ['PostEndpoint', function (PostEndpoint) {
+module.exports = ['PostEndpoint', 'moment', function (PostEndpoint, moment) {
     return {
         restrict: 'E',
         replace: true,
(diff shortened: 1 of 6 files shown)


rjmackay:
@jshorland assuming the auto deploy works you should be able to test this before it lands.. the link should appear above in the timeline.

app/main/posts/detail/post-value.directive.js
```diff
@@ -18,6 +18,21 @@ module.exports = ['PostEndpoint', function (PostEndpoint) {
                     return PostEndpoint.get({ id : entry });
                 });
             }
+            if ($scope.attribute.type === 'datetime' && $scope.attribute.input === 'date') {
```
willdoran: Could this be a single outer if($scope.attribute.type==='datetime') 
And then split out ifs beneath it, I'm not sure if that would be clearer

rjmackay: Yup will do

PR #398, opened 2016-11-02
Hide 'filter by survey' button unless we're in an embed

rjmackay:
Fix ushahidi/platform#1472

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/398)


brandonrosage:
@rjmackay In the non-embed views, the `button-group` inside the toolbar _should not_ have the class "hide-until-medium." They should be visible at all screen sizes. That class is only intended for the embed view for Map & Timeline.

Otherwise, things look great.

PR #406, opened 2016-11-03
1452 video embed

willdoran:
- Adds video functionality

Test these changes by:
- [ ] Create a new survey
- [ ] Add to this survey a video embed field
- [ ] Save survey
- [ ] Add new post of the above survey type
- [ ] Set the field to junk and attempt to save
- [ ] Set field to youtube url of different types and save
- [ ] Set field to vimeo url of different types and save
- [ ] Attempt to view video on Post detail field
- [ ] Attempt to edit
- [ ] Change url and save

Refs ushahidi/platform#1452

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/406)


app/main/posts/detail/detail.html
```diff
@@ -81,10 +81,20 @@ <h2 class="form-label" translate>post.modify.form.categories</h2>
             {{category.tag}}
           </p>
         </div>
+        <div
+          ng-repeat="(key,value) in post.values"
```
rjmackay: This leaves us looping through `post.values` twice, and potentially showing thing out of order. We should at least reduce this to a single ng-repeat above both post-value and post-video-value.. or as suggested below move post-video-value inside post-value

app/main/posts/detail/post-video-value.directive.js
```diff
@@ -0,0 +1,31 @@
...
+];
+
+function PostVideoValueController(
+    $scope,
+    $sce
+) {
+    activate();
+
+    function activate() {
+        $scope.videoUrl = $sce.trustAsResourceUrl($scope.videoUrl[0]);
```
rjmackay: should we at least sanity check this URL here? ie. is it actually youtube?

app/main/posts/posts-module.js
```diff
@@ -11,13 +11,15 @@ angular.module('ushahidi.posts', [])
...
 .directive('postMediaValue', require('./detail/post-media-value.directive.js'))
+.directive('postVideoValue', require('./detail/post-video-value.directive.js'))
 .service('PostActionsService', require('./common/post-actions.service.js'))
 
 // Create / Edit Screens
 .service('PostEntity', require('./modify/post-entity.service.js'))
 .service('PostEditService', require('./modify/post-edit.service.js'))
 .service('MediaEditService', require('./modify/media-edit.service.js'))
 .directive('postMedia', require('./modify/post-media.directive.js'))
+.directive('postVideo', require('./modify/post-video.directive.js'))
```
rjmackay: I'd really like to improve the naming scheme on these. When we have PostVideo and PostVideoValue its not clear which is input or output. Could we have PostVideoInput/PostVideoEdit and PostVideoView?

app/main/posts/modify/post-video.directive.js
```diff
@@ -0,0 +1,92 @@
...
+    $sce,
+    Util
+) {
+    $scope.constructIframe = constructIframe;
+    // $scope.getVideoThumbnail = getVideoThumbnail;
+    activate();
+
+    function activate() {
+        // Is this dody? I dunno it seems dodgy
+        $scope.videoUrl = $sce.trustAsResourceUrl($scope.videoUrl);
```
rjmackay: I'm not clear what this does.. but it seems like we should sanity check before saying X is trusted?

app/main/posts/modify/post-video.directive.js
```diff
@@ -0,0 +1,92 @@
...
+    function constructIframe(videoUrl) {
+        parseVideo(videoUrl);
+
+        var preview = angular.element(document.getElementById($scope.previewId));
+
+        if ($scope.videoUrl === undefined) {
+            preview.empty();
+        } else {
+            preview.empty();
+            preview.html(createVideo()).wrap('<div class="video_embed-fluid" />');
```
rjmackay: why are we constructing HTML here instead of just using `ng-show` and setting variables in scope?

app/common/locales/en.json
```diff
@@ -572,10 +573,19 @@
             }
         },
         "media" : {
+            "enter_a" : "Enter a",
+            "youtube" : " Youtbe",
```
willdoran: done

app/common/locales/en.json
```diff
@@ -572,10 +573,19 @@
...
             "current_url" : "Current image",
             "add_caption" : "Add caption",
             "upload" : "Upload"
         },
+        "video" : {
```
willdoran: done

app/main/posts/detail/detail.html
```diff
@@ -81,10 +81,20 @@ <h2 class="form-label" translate>post.modify.form.categories</h2>
...
+          ng-if="isPostValue(key) && form_attributes[key] && showType(form_attributes[key].type)"
+          class="postcard-field"
+          >
 
+          <post-video-value
```
willdoran: done

app/main/posts/modify/post-video.directive.js
willdoran: Unfortunately, the iframe won't reload if we just update the src, it breaks, we have to remove it and readd.

app/main/posts/modify/post-video.directive.js
willdoran: So yeah, this is the real problem here, we have to trust the url. I will add a regex to check that it is a video url based on what we think a video url is but this is still potentially going to cause problems. 

We are assuming that the url in the iframe is going to be kosher. It would not be hard to trick our regex into allowing a url that looked right but was malicious. I'm not positive what to do about this. 

For the moment, I have prevented them from entering urls that don't match the format of youtube or vimeo urls. The fact they are gameable urls is sort of difficult to fix because we don't control vimeo or youtube and as such can't force their url format super strictly.

So if someone is able to insert a malicious url that passes our regex and loads in the iframe it could infect the machine of someone viewing the page. 

I'll think about more about sanitisation.

PR #408, opened 2016-11-08
Better handling when logging in and out

rjmackay:
- Stay on current page when logging in
- If access is denied, try to redirect back after login
- Stay on current page for log out, but reload

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/408)


jshorland:
Passes qa.

PR #410, opened 2016-11-08
Fix errors parsing domain for intercom

rjmackay:
- Use $location.host() instead
- Fix the data we send to intercom for `company` too

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/410)


rjmackay:
@jshorland FYI this should push 'Companies' into intercom properly. However this is still only going to happen for users as they log in. As we add a events we can improve this and push ALL users in to intercom.

PR #413, opened 2016-11-09
Make PostFilters aware of current view mode

rjmackay:
- Make post filters aware of viewing collection, search or all posts
- Reset filters when changing mode
- Maintain filters when switching map/timeline view but not mode

Refs ushahidi/platform#1089  ushahidi/platform#1200

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/413)


rjmackay:
@willdoran I'd be interested in your thoughts on this. We talked a bit about this a few weeks ago..
I'm not sure fond of mixing this logic in with the Post Filters.. but I wasn't sure where else to manage it really. Maybe state could live elsewhere, but then filters stay aware of it.

or perhaps we should handle less of this through routing in the first place, and rather combine all posts/collections/saved searches into a single controller with several directives and some managed state. I'm not sure.

This however does appear to work just fine

rjmackay:
One thing that is a bit messy: mode bar maintains some similar state for highlighting menu items so I might try to combine the too

rjmackay:
@Angamanga could you give this a final review + test?

PR #414, opened 2016-11-10
Replace posts pagination with 'load more'

rjmackay:
- Replace posts pagination with load more
- Fix broken </pagination> close tags
- Increase posts page size to 20
- Update post list after delete/status change without reloading
- Add 'loading' button when posts are loading

Refs ushahidi/platform#1441

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/414)


rjmackay:
> I wanted to find something to track but the only question I had was why use angular.forEach instead of underscore?

I generally prefer the angular function over underscore.. but not with any real logic either way. They do the same thing

willdoran:
Cool, I had been tending to use underscore because I thought it would ultimately make the code more portable but I think that's a fallacy.

rjmackay:
@jshorland this is ready for QA at http://qa.feature-load-more.ushahidilab.com/

PR #428, opened 2016-12-02
1412 role ui

Angamanga:
~~This pr is dependent on pr: https://github.com/ushahidi/platform/pull/1523~~

- Adds a directive that corresponds with the ui in http://preview.ushahidi.com/platform-pattern-library/develop/assets/html/5_layouts/settings-category-edit.html
- Makes the edit-categories-template use the directive-code for role-selection

Test these changes by:
- [ ] Click on settings->categories->choose a category
- [ ] Compare ui of selecting roles with http://preview.ushahidi.com/platform-pattern-library/develop/assets/html/5_layouts/settings-category-edit.html
- [ ] Test saving a category with different roles chosen
- [ ] Make sure the new roles are saved properly after saving

Fixes ushahidi/platform#1412

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/428)

### DIFF
--- a/app/common/directives/role-selector.directive.js
+++ b/app/common/directives/role-selector.directive.js
@@ -16,7 +16,17 @@ function RoleSelectorDirective() {
 RoleSelectorController.$inject = ['$scope', 'RoleEndpoint'];
 
 function RoleSelectorController($scope, RoleEndpoint) {
+    $scope.addAllRoles = addAllRoles;
+
+    // getting available roles from api
     RoleEndpoint.query().$promise.then(function (roles) {
         $scope.roles = roles;
     });
+
+    // adding all available roles to model if user clicks 'Everyone'
+    function addAllRoles() {
+        $scope.model.role = $scope.roles.map(function (role) {
+            return role.name;
+        });
+    }
 }
--- a/app/common/directives/role-selector.directive.js
+++ b/app/common/directives/role-selector.directive.js
@@ -1,0 +1,22 @@
+module.exports = RoleSelectorDirective;
+
+RoleSelectorDirective.$inject = [];
+
+function RoleSelectorDirective() {
+    return {
+        restrict: 'E',
+        scope: {
+            model: '=',
+            title: '='
+        },
+        controller: RoleSelectorController,
+        template: require('./role-selector.html')
+    };
+}
+RoleSelectorController.$inject = ['$scope', 'RoleEndpoint'];
+
+function RoleSelectorController($scope, RoleEndpoint) {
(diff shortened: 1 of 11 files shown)


app/common/directives/role-selector.directive.js
```diff
@@ -0,0 +1,32 @@
...
+        template: require('./role-selector.html')
+    };
+}
+RoleSelectorController.$inject = ['$scope', 'RoleEndpoint'];
+
+function RoleSelectorController($scope, RoleEndpoint) {
+    $scope.addAllRoles = addAllRoles;
+
+    // getting available roles from api
+    RoleEndpoint.query().$promise.then(function (roles) {
```
willdoran: can we place this part inside an activate function:

activate();

function activate() {
   RoleEndpoint.query(...
}

app/common/directives/role-selector.directive.js
```diff
@@ -0,0 +1,32 @@
...
+        $scope.roles = roles;
+    });
+
+    // adding all available roles to model if user clicks 'Everyone'
+    function addAllRoles() {
```
willdoran: Here, do we want to set the everyone_can_see property of the category? Was the original behaviour to add all the roles in an array if so we can probably keep that.

app/settings/categories/categories-edit.html
```diff
@@ -50,28 +50,7 @@ <h1 class="mode-context-title" ng-show="!category.id" translate>app.add_category
...
-                                    checklist-model="category.role"
-                                    checklist-value="role.name"
-                                >
-                                <label for="{{role.id}}">
-                                    {{role.display_name}}
-                                </label>
-                            </div>
-                        </div>
-                    </fieldset>
+                    <role-selector model="category" title="'Who can see and add to this category'"></role-selector>
```
willdoran: the title here will need to be translated, we can do that with
$scope.title = $translate.instant($scope.title);
in the directive

app/common/services/endpoints/tag.js
```diff
@@ -41,7 +41,9 @@ function (
 
     TagEndpoint.getFresh = function (params) {
         cache.remove(Util.apiUrl('/tags/' + params.id));
-        return TagEndpoint.get(params);
+        return TagEndpoint.get(params).$promise.then(function (result) {
```
Angamanga: @willdoran I had to make this change to get hold of the category.role-values in /settings/categories/edit.controller.js correctly. I don't think it affects something else, but not 100% sure?

willdoran: hmmm let's look at this together

PR #465, opened 2016-12-20
Visibility of categories

Angamanga:
- Wipes the category-cache when logging in and logging out
- Makes correct categories visible for the current user

Test these changes by:
- Create categories with different roles
- Create posts with different categories attached to them
- Make sure that: 
1. Only categories that should be visible to the current user is visible from the list when creating a post
2. Only categories that should be visible to the current user is visible when looking at a post, even if other categories with other roles is attached to the post.
3. Try logging in and out as different users and make sure the correct categories still are visible to each user.

Fixes ushahidi/platform#1505 (and this also fixes the issue left from QA in issue #1412 ) .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/465)

### DIFF
--- a/app/common/services/endpoints/tag.js
+++ b/app/common/services/endpoints/tag.js
@@ -10,10 +10,10 @@ function (
     CacheFactory
 ) {
     var cache;
-
     if (!(cache = CacheFactory.get('tagCache'))) {
         cache = new CacheFactory('tagCache');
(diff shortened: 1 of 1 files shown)


rjmackay:
Looks good to me

PR #574, opened 2017-02-28
Adding number of unmapped posts to map-view

Angamanga:
- Adds info about nb of unmapped posts in map-view
- Adds a link to time-line where unmapped posts are filtered
- Displays the unmapped posts in time-line when clicking on link

Test these changes by:
- Pull in a nb of tweets, 
- Check the left margin of map-view, a test with nb of unmapped post should be visible if more than 1 unmapped posts are present
- Click on link with nb of posts
- The link should send you to the time-line and only show the unmapped posts.

**Concerns:** 
1. Should we add the option of only showing unmapped posts to the filter-bar as well?
2. There is no easy way to unset the unmapped-filter right now, should we add that?

Fixes ushahidi/platform#1392 .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/574)


app/main/posts/views/post-view-list.directive.js
```diff
@@ -97,18 +97,37 @@ function PostListController(
         PostEndpoint.query(postQuery).$promise.then(function (postsResponse) {
             // Add posts to full set of posts
             // @todo figure out if we can store these more efficiently
-            Array.prototype.push.apply($scope.posts, postsResponse.results);
 
+            // TODO/question: feels like this logic should belong somewhere else in the application? But where?
```
Angamanga: I would like to put this somewhere else but I don't think I have entirely gotten how the filtering works so I don't know where? Any hints?

app/main/posts/views/mode-context-form-filter.directive.js
```diff
@@ -10,23 +10,29 @@ function ModeContextFormFilterDirective() {
...
+    $scope.getUnmapped = getUnmapped;
     activate();
 
     function activate() {
         // Load forms
         $scope.forms = FormEndpoint.query();
         var postCountRequest = PostEndpoint.stats({ group_by: 'form', status: 'all' });
-
-        $q.all([$scope.forms.$promise, postCountRequest.$promise]).then(function (responses) {
+        // todo: can I limit this to only returning 'unmapped'?
```
Angamanga: Here is an extra request for getting the number of unmapped posts. This requires an extra request and I only need the 'unmapped' attribute. Can I adjust the search to just return that number? Any hints? I don't need the whole collection at this point...

willdoran: I think we need to look at our requests overall, particularly, in light of the needs of mobile. You could add a separate request route to return just the counts. I'll take a look at this later today and try to come up with a better answer.

app/main/posts/views/mode-context-form-filter.directive.js
```diff
@@ -10,23 +10,29 @@ function ModeContextFormFilterDirective() {
...
+        // todo: How about the requests? Do I need to check auth first before searching for draft?
```
willdoran: You shouldn't need to, auth should be handled by the session. It should fail to return if the request doesn't bear the token, I think.

app/main/posts/views/post-view-list.directive.js
willdoran: Yep so, the filtering is complex, it should live here
https://github.com/ushahidi/platform-client/tree/develop/app/main/posts/views/filters

Angamanga:
@will have moved the filter-functionality to the api (pr https://github.com/ushahidi/platform/pull/1589) and a few other changes to the filter-directive.

app/main/posts/views/mode-context-form-filter.directive.js
```diff
@@ -66,7 +65,11 @@ function ModeContextFormFilter($scope, FormEndpoint, PostEndpoint, $q, _, $rootS
     }
 
     function goToUnmapped() {
-        $scope.PostFilters.setFilters({unmapped: ['unmapped']});
+        var filters = PostFilters.getDefaults();
+        filters.form.push('none');
+        filters.has_location = 'unmapped';
+        PostFilters.setFilters(filters);
+        // is there way of going to list and change the currentView without hardcoding the path?
```
Angamanga: Is there another way of going to the list without hardcoding the path?

willdoran: No annoyingly enough there isn't, we don't have states in the way rollcall does yet.

PR #647, opened 2017-05-12
Basic chunking using limit of 200 and iterating until all posts have

willdoran:
- Added basic chunking
- On completion of first request, we calculate how many additional request we should make for the given limit and if we need an additional request for the last set of posts if remainder is less than limit
- Changed to automatically add more posts
- Separate map clearing code

Test these changes by:
-

Fixes ushahidi/platform# .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/647)


app/main/posts/views/post-view-map.directive.js
```diff
@@ -85,6 +88,7 @@ function PostViewMap(PostEndpoint, Maps, _, PostFilters, L, $q, $rootScope, $com
                 return $scope.filters;
             }, function (newValue, oldValue) {
                 if (newValue !== oldValue) {
+                    clearData();
                     reloadMapPosts();
```
rjmackay: We need to make sure we show some kind of `loading` status in the UI while this happens now.

app/main/posts/views/post-view-map.directive.js
```diff
@@ -94,12 +98,35 @@ function PostViewMap(PostEndpoint, Maps, _, PostFilters, L, $q, $rootScope, $com
             loadPosts().then(addPostsToMap);
         }
 
-        function loadPosts(query) {
+        function loadPosts(query, offset) {
             query = query || PostFilters.getQueryParams($scope.filters);
-
+            query.has_location = 'mapped';
+            offset = offset || 0;
+            var limit = 200,
```
rjmackay: Could we hoist this to the top of the directive? ie. so its easy to find and tweak the magic number later when we need to.

app/main/posts/views/post-view-map.directive.js
```diff
@@ -94,12 +98,35 @@ function PostViewMap(PostEndpoint, Maps, _, PostFilters, L, $q, $rootScope, $com
...
+                    var remainder = (posts.total - limit) % limit;
+
+                    while (evenChunks > 0) {
+                        evenChunks -= 1;
+                        offset += limit;
+                        loadPosts(query, offset).then(addPostsToMap);
+                    }
+
+                    if (remainder > 0) {
+                        offset += remainder;
```
rjmackay: There's no need to use an exact limit. You can just add 1 to the number of chunks if there is any remainder

app/main/posts/views/post-view-map.directive.js
```diff
@@ -94,12 +98,35 @@ function PostViewMap(PostEndpoint, Maps, _, PostFilters, L, $q, $rootScope, $com
...
+                offset: offset
+            });
             $scope.isLoading = true;
-            return PostEndpoint.geojson(query).$promise.then(function (posts) {
+            return PostEndpoint.geojson(conditions).$promise.then(function (posts) {
                 $scope.isLoading = false;
+
+                // Retrieve remaining posts
+                if (offset === 0 && posts.total > limit) {
+                    var evenChunks = Math.floor((posts.total - limit) / limit);
```
rjmackay: can we rename this to something clearer? ie. numberOfChunks?

willdoran: 👍

app/main/posts/views/post-view-map.directive.js
willdoran: 👍

PR #654, opened 2017-05-25
displaying categories in a post and adding new categories in a post

Angamanga:
- Displays correct categories when adding / editing a post
- Makes it possible to add a category from a post

Testing checklist:
- [x] Add a new post to a field with categories
- [x] Make sure the categories connected to that form is displayed
- [x] Add a new category through clicking on 'Add category' under the category-list, make sure it saves and is displayed as an option.
- [x] The above should also work for editing a post. 

Link to QA-deployment: http://qa.category-add-and-filter.ushahidilab.com
Fixes ushahidi/platform#1397 .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/654)

### DIFF
--- a/app/main/posts/modify/post-editor.directive.js
+++ b/app/main/posts/modify/post-editor.directive.js
@@ -142,9 +142,13 @@ function PostEditorController(
                 }
                 if (attr.input === 'tags') {
                     // adding category-objects attribute-options
-                    attr.options = _.map(attr.options, function (category) {
+                    attr.options = _.chain(attr.options)
+                        .map(function (category) {
                             return _.findWhere(categories, {id: category});
-                        });
+                        })
+                        .filter()
+                        .value();
+
                     // adding category-objects to children
                     _.each(attr.options, function (category) {
                         if (category.children.length > 0) {
--- a/app/main/posts/modify/post-editor.directive.js
+++ b/app/main/posts/modify/post-editor.directive.js
@@ -114,7 +114,7 @@ function PostEditorController(
             var attrs = _.chain(results[1])
                 .sortBy('priority')
                 .value();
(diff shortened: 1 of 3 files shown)


app/main/posts/modify/post-editor.directive.js
```diff
@@ -140,6 +140,25 @@ function PostEditorController(
...
+                            return _.findWhere(categories, {id: category});
+                        });
+                    // adding category-objects to children
+                    _.each(attr.options, function (category) {
+                        if (category.children.length > 0) {
+                            category.children = _.chain(category.children)
+                                .map(function (child) {
+                                    return _.findWhere(attr.options, {id: child.id});
+                                })
+                                .filter(function (child) {
```
rjmackay: FYI just _.filter() with no arguments remove false/undefined values.

app/settings/surveys/survey-editor.directive.js
```diff
@@ -604,26 +600,16 @@ function SurveyEditorController(
...
-        });
-        return tags;
-    }
-
     function saveAttributes() {
         var calls = [];
         _.each($scope.survey.tasks, function (task) {
             _.each(task.attributes, function (attribute) {
+                // removing faulty category-ids
+                if (attribute.input === 'tags') {
```
rjmackay: This should check `attribute.type === 'tags'`

rjmackay: Not sure I understand *why* we're doing this but it seems safe enough

Angamanga: Yes, it is a precaution to not end up with weird values in options.

PR #656, opened 2017-05-25
Fixed bug that prevented delete all for selected categories

willdoran:
- Fixed bug where category was being added to select list rather than category id

Testing checklist:
- [ ] On Category list view, select a category, then when the toolbar actions appear, click Select All, then click delete, confirm deletion, all categories should be deleted
- [ ] Having created categories with parents, select the child category, only the child category should be selected
- [ ] Delete the child category only the child category should be deleted
- [ ] Having created categories with parents, select the parent category, only the parent category should be selected
- [ ] Select a category, then hit select all, then unselect one of the categories, the number of selected categories should decrease

Fixes ushahidi/platform# .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/656)

### DIFF
--- a/app/settings/categories/categories.controller.js
+++ b/app/settings/categories/categories.controller.js
@@ -33,6 +33,7 @@ function (
 
     $scope.refreshView = function () {
         TagEndpoint.queryFresh().$promise.then(function (tags) {
+            $scope.allCategories = tags;
             $scope.categories = _.map(_.where(tags, { parent_id: null }), function (tag) {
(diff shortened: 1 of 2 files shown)


rjmackay:
So if you 
1. Go to categories list
2. Select a category 
3. Click select all
4. Deselect a child category and watch the selected counter go UP

rjmackay:
- Deleting a child category, deletes its parent too

PR #659, opened 2017-05-26
Fix category minor issues

willdoran:
- Fixes Confirm delete message
- Fixes selecting parent as itself a parent

Testing checklist:
- [ ] On a category's edit page, click delete the confirm delete message should be: 
`Deleting this category will remove it from all existing posts. This action cannot be undone.`
- [ ] When creating/editing a category, only parents/categories without children should be selectable as a parent
- [ ] When editing a parent/category without children it should not itself be selectable as its own parent
- [ ]  Create a category
- [ ] Edit a category
- [ ] Children/parent-relationships:
  - [ ] A parent cannot be a child
  - [ ] A child cannot be a parent
  - [ ] Dropdown-box with parents should not be visible for parents
- [ ] Delete a category inside the edit-page
- [ ] Delete a single category from the category-settings-list
- [ ] Bulk delete categories from the category-settings-list
- [ ] Delete a parent, child should be set as a upper-level category
- [ ] Add a category-field to a survey:
  - [ ] Correct categories are checked when editing the survey
  - [ ] Displayed categories when adding a post are the same as the ones chosen for the survey
- [ ] Edit a category-field, displayed categories when adding the post should be changed to the new, edited categories.
- [ ] Add a category on the fly, the new category should be:
  - [ ] checked
  - [ ] added to the category-field when adding a new post
  - [ ] added to the list of categories when editing the actual survey
- [ ] Delete a category-field from a survey
  - [ ] The field should not be visible when adding a post to the survey
  - [ ] Data from the field in previous posts should be deleted 
- [ ] Filter posts, correct posts should be visible
- [ ] CSV-import, correct categories should be added to the posts.
Fixes ushahidi/platform# .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/659)

### DIFF
--- a/app/settings/categories/categories-edit.html
+++ b/app/settings/categories/categories-edit.html
@@ -76,7 +76,7 @@
                             type="radio"
                             name="parent_label"
                             id="parent_label-nothing"
-                            value="null"
+                            ng-value="null"
                             ng-checked="!category.parent"
                             ng-model="category.parent_id"
                             />
@@ -123,5 +123,3 @@
         </div>
     </main>
 </div>
-
-
--- a/app/settings/categories/categories-edit.html
+++ b/app/settings/categories/categories-edit.html
@@ -56,6 +56,7 @@
               <fieldset
                 class="custom-fieldset init"
                 dropdown auto-close="outsideClick"
+                ng-show="!isParent"
               >
                 <legend
                   class="dropdown-trigger init active"
--- a/app/settings/categories/categories-edit.html
+++ b/app/settings/categories/categories-edit.html
@@ -79,6 +79,7 @@
                             ng-value="null"
                             ng-checked="!category.parent"
                             ng-model="category.parent_id"
+                            ng-click="deselectParent()"
                             />
                           Nothing
                       </label>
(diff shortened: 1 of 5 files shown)


app/settings/categories/categories-edit.html
```diff
@@ -76,9 +77,10 @@ <h1 class="mode-context-title" ng-show="!category.id" translate>app.add_category
                             type="radio"
                             name="parent_label"
                             id="parent_label-nothing"
-                            value="null"
+                            ng-value="null"
                             ng-checked="!category.parent"
                             ng-model="category.parent_id"
+                            ng-click="deselectParent()"
```
rjmackay: This should really be using checklist-model to avoid the messiness of so many ng- directive trying to work together

app/settings/categories/edit.controller.js
```diff
@@ -61,8 +62,13 @@ function (
         if (tag.length === 0) {
```
rjmackay: We could actually just check `$scope.category.children` to enable/disable this

app/settings/categories/categories-edit.html
willdoran: Users has the same issue BUT roles doesn't so you could just take from roles.

PR #660, opened 2017-05-29
Deselecting children when deselecting a parent

Angamanga:
- When a parent is deselected in the mode-context-bar, the children is also deselected

Testing checklist:
- [x] Add a post which contains a child-tag (this is for testing-purposes)
- [x] Select the parent in mode-context-bar
  - [x] The post should be visible
- [x] Deselect the parent in the mode-context-bar
  - [x] The child tag should also be deselected and the post should not appear.

QA can be done here: http://qa.filtering-categories-fix.ushahidilab.com

Fixes ushahidi/platform#1397 .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/660)

### DIFF
--- a/app/main/posts/views/mode-context-form-filter.directive.js
+++ b/app/main/posts/views/mode-context-form-filter.directive.js
@@ -90,7 +90,6 @@ function ModeContextFormFilter($scope, FormEndpoint, PostEndpoint, TagEndpoint, 
     function selectParent(parent, formId) {
         // If we've just selected the tag
         if (_.contains($scope.filters.tags, parent.id)) {
-            console.log('selected');
             // ... then select its children too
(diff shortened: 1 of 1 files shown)


app/main/posts/views/mode-context-form-filter.directive.js
```diff
@@ -90,6 +90,7 @@ function ModeContextFormFilter($scope, FormEndpoint, PostEndpoint, TagEndpoint,
     function selectParent(parent, formId) {
         // If we've just selected the tag
         if (_.contains($scope.filters.tags, parent.id)) {
+            console.log('selected');
```
willdoran: should drop this

PR #670, opened 2017-06-19
1759 styling categories: changing margins

crcommons:
-Removes 50px margin from child category div
-Adds 50px margin directly to input box and a tag
-Result: keeps trash bin icon aligned with other bins

Testing checklist:
- [ ]

Fixes ushahidi/platform#1759 .

Ping @ushahidi/platform

### DIFF
--- a/app/settings/categories/categories.html
+++ b/app/settings/categories/categories.html
@@ -1,5 +1,6 @@
+
 <div>
-    <layout-class layout="b"></layout-class>
+	<layout-class layout="b"></layout-class>
     <div class="mode-context init" dropdown>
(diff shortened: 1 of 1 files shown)


middle8media:
@crcommons I am curious ... is there a specific reason that we are using inline styles here as opposed to fixing it in the platform PL?

@rjmackay

rjmackay:
@middle8media no. We should move it into the PL, but I need to pair up with @crcommons and give an intro to the PL. Thanks for catching this though.

rjmackay:
Lets close this and create a new PR with the pattern library update

PR #677, opened 2017-06-27
Fixes custom field description not saving

ryanchristo:
- adds attribute instructions model to description input

Testing checklist:
- [x] When I change the description for a custom field and save, the changes are saved

Fixes ushahidi/platform#1787.

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/modify/post-value-edit.html
+++ b/app/main/posts/modify/post-value-edit.html
@@ -16,6 +16,9 @@
       <label for="values[{{attribute.key}}][0]">
           {{attribute.label}}
       </label>
+      <!-- Attribute Instructions -->
+      <p>{{attribute.instructions}}</p>
+
       <!-- attributes which use the form-field structure -->
       <div ng-repeat="(key, value) in post.values[attribute.key] track by key">
           <div ng-switch="attribute.input">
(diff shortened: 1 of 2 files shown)


Angamanga:
@ryanchristo Thank you so much for your submission! @willdoran or @rjmackay This looks good to me, could you double-check if it is the instructions-column in the form_attributes tables that should be used here, or the description-column in the form_stages? Also, I do not find anywhere in the pattern-library where this description is used. Should it be only in the attributes-edit-field or should it be displayed when adding/viewing posts too? Not sure how the discussions went when designing this?

willdoran:
@Angamanga I'm not completely sure, I'll have to look at the data model.

rjmackay:
This looks good to me. However we should make sure the `instructions` field is actually displayed on the survey form too!

@ryanchristo do you think you could add that as well? I think it need to use a pattern similar to "Form field with instructions" shown here: http://preview.ushahidi.com/platform-pattern-library/develop/assets/html/2_fragments/#pl-pattern-forms
in the `post-value-edit` directive. That really just amounts to a `<p>{{ attribute.instructions }}</p>` after the label, before the input

ryanchristo:
@rjmackay Post form updated with attribute instructions.

ryanchristo:
@rjmackay Attribute instructions added.

PR #678, opened 2017-06-27
Apply parent category permissions to children on save

ryanchristo:
- Applies the parent category permissions to each child when the parent category is saved.
- Applies the new parent category permissions to the child category if the parent is changed.
- Adds a message to child categories that says permissions are managed by the parent.

Testing checklist:
- [x] When I change permissions to a category that has child categories, each child category is updated with the same permissions as the parent.
- [x] When I edit a child category and change the parent category, the child category is updated with the same permissions as the parent.
- [x] When I edit a child category, I am not able to change the permissions and instead I see a message that says the permissions are managed by the parent.

Fixes ushahidi/platform#1789 / ushahidi/platform#1769.

Ping @ushahidi/platform


Angamanga:
@ryanchristo Thank you for your submission, this looks good to me and works in the following way: If a parents role is changed, the childrens role will be updated to the same as the parents. @jshorland @aoduor Could you please double-check this: What should we do if a childrens role is changed? Should we then change the parent too? Or remove the option to save roles on children in the UI completely?

ryanchristo:
@Angamanga @jshorland @aoduor I updated the pull request with a preview of mentioned changes in ushahidi/platform#1789. Let me know if you would like me to adjust the message or anything else.

Angamanga:
@ryanchristo Thank you again for your submission! These changes makes sense from a code-perspective, @jshorland, @aoduor Is this solution ok for you?: screen shot 2017-06-28 at 8 13 53 am

aoduor:
I think this works.. Just confirming that, if you changed a child category and updated it to be a parent category, that the permissions section would re-appear?

PS: @ryanchristo you rock socks :)

ryanchristo:
@aoduor Thanks Angela! And to answer your question, yes, if you change a child category to be a parent category, the permissions section re-appears. If you change your mind, you can change it back too. You can switch back and forth as long as the category is originally a child category.

ryanchristo:
@Angamanga I started updating the edit category directive to improve the way it was handling the promises on save (which was an issue brought up in ushahidi/platform-client#680 for handling the saves when creating and editing a survey), then realized I needed to update the create category directive with the same code. I ended up combining the create and edit category directives instead, which is the same setup for creating and editing a survey and makes the saving actions consistent whether creating a new category or saving an existing one.

If I took this in the wrong direction, please let me know. I can revert back and copy the save actions. The main goal of the most recent updates was to ensure all promises have been processed before displaying the success message and redirecting.

ryanchristo:
Looking into the failed Travis CI build. Will need to update the tests to take into account promises or fix the structure of the promises in the directive.

PR #679, opened 2017-06-28
Removed auto-focus on posts so default view is always used

ryanchristo:
- removes auto-focus on posts so default view is always used

Testing checklist:
- [x] When I add posts to a new deployment, the map does not automatically zoom in on the posts, instead the map always uses the default map configuration settings.

Fixes ushahidi/platform#1736.

Ping @ushahidi/platform


rjmackay:
This is pretty significant behaviour change, and while it solves the edge case logged in #1736 it possibly creates further issues elsewhere, lets see if we can get a decision about what the solution should be on #1736 before landing any changes.

ryanchristo:
As discussed in ushahidi/platform#1736, I agree and think ushahidi/platform#1339 provides a solution in which this request would no longer be necessary.

PR #681, opened 2017-07-04
Label and input ids were conflicting for task configure tab

willdoran:
- Set not conflicting ids for Task Configure Tab inputs and labels

Testing checklist:
- [x] Create a new survey
- [x] Add one task to the survey, click on the 'Configure' tab for the task
- [x] Toggle the 'Required' and 'Show this task to everyone when published' and ensure they turn on and off correctly
- [x] Add a second task to the survey, click on the 'Configure' tab for the task
- [x] Toggle the 'Required' and 'Show this task to everyone when published' and ensure they turn on and off correctly
- [x] Save the Survey
- [ ] Set the 'Required' toggle for the first task, confirm this works as expected by attempting to create a Post of this type
- [ ] Set the 'Required' toggle for the second task, confirm this works as expected by attempting to create a Post of this type
- [x] Set the 'Show this task to everyone when published' toggle for the second task, confirm this works as expected by attempting to create a Post of this type and publish it
- [x] Set the 'Show this task to everyone when published' toggle for the second task, confirm this works as expected by attempting to create a Post of this type and publish it

Fixes ushahidi/platform#1846

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/modify/post-tabs.directive.js
+++ b/app/main/posts/modify/post-tabs.directive.js
@@ -4,7 +4,6 @@ PostVerticalTabs.$inject = [];
 function PostVerticalTabs() {
     return {
         restrict: 'E',
-        replace: true,
         scope: {
             form: '=',
             post: '=',
@@ -12,8 +11,8 @@ function PostVerticalTabs() {
             attributes: '=',
             visibleStage: '='
         },
-        controller: PostVerticalTabsController,
-        template: require('./post-tabs.html')
+        template: require('./post-tabs.html'),
+        controller: PostVerticalTabsController
     };
 }
 
@@ -37,6 +36,7 @@ function PostVerticalTabsController(
     $scope.stageIsComplete = stageIsComplete;
     $scope.toggleStageCompletion = toggleStageCompletion;
 
+
     activate();
 
     function activate() {
@@ -48,11 +48,10 @@ function PostVerticalTabsController(
     }
 
     function stageIsComplete(stageId) {
-        return _.includes($scope.post.completed_stages, stageId);
+        return $scope.post.completed_stages.indexOf(stageId) > -1;
     }
 
     function toggleStageCompletion(stageId) {
-
         stageId = parseInt(stageId);
         if (_.includes($scope.post.completed_stages, stageId)) {
             $scope.post.completed_stages = _.without($scope.post.completed_stages, stageId);
--- a/app/settings/surveys/survey-editor.html
+++ b/app/settings/surveys/survey-editor.html
@@ -333,18 +333,18 @@
           <div class="form-field switch">
             <label translate="survey.required">Required</label>
             <p data-fieldgroup-target="require-review-message" ng-show="task.required" translate="survey.required_desc">Require this section be completed before a post can be visible to the public.</p>
-            <div class="toggle-switch">
-                <input class="tgl" id="task-required" type="checkbox" ng-model="task.required">
-                <label class="tgl-btn" for="task-required"></label>
+            <div class="toggle-switch" ng-attr-id="{{ 'toggle-switch-required-' + task.id }}">
+                <input class="tgl" ng-attr-id="{{ 'task-required-' + task.id }}" type="checkbox" ng-model="task.required">
+                <label class="tgl-btn" for="task-required-{{task.id}}" ng-attr-id="{{ 'task-required-label-' + task.id }}"></label>
             </div>
         </div>
           <div class="form-field switch checked">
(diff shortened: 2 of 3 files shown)


app/settings/surveys/survey-editor.html
```diff
@@ -333,18 +333,18 @@ <h2 class="listing-item-title">
...
+                <input class="tgl" ng-attr-id="{{ 'task-required-' + task.id }}" type="checkbox" ng-model="task.required">
+                <label class="tgl-btn" for="task-required-{{task.id}}" ng-attr-id="{{ 'task-required-label-' + task.id }}"></label>
             </div>
         </div>
           <div class="form-field switch checked">
               <label translate="survey.show_this_task_to">Show this task to everyone when published</label>
               <p data-fieldgroup-target="show_section-message" class="active" ng-show="task.show_when_published" translate="survey.show_this_task_to_desc">When a survey response is published, data from this section will be displayed.</p>
-              <div class="toggle-switch">
-                  <input class="tgl init" id="show_section" type="checkbox" data-fieldgroup-toggle="show_section-message" ng-model="task.show_when_published">
+              <div class="toggle-switch" ng-attr-id="{{ 'toggle-switch-show-' + task.id }}">
```
rjmackay: Looks good. Not sure why we need IDs on the div and label though?

rjmackay:
Looks good. However when I set a task as 'required before publish' then try to mark it complete & publish, I can't publish. I keep getting `Stage "Translation" is required before publishing` despite marking it complete.

Also - why is this against `master` not `develop`?

rjmackay:
So investigating a little more. If I mark a stage complete, it just never saves that whether I try to publish or not. Thats probably completely unrelated to this PR though.

willdoran:
It's against master because it's a patch for an issue in production, it'll have to be back ported into develop. Should it be against production instead?

app/settings/surveys/survey-editor.html
willdoran: We don't technically, I'll pop them off.

willdoran:
Interesting, I'll look at the publishing issue

willdoran:
I see the bug I'll fix it with this PR too

rjmackay:
> It's against master because it's a patch for an issue in production, it'll have to be back ported into develop. Should it be against production instead?

Nope thats perfect. Just wanted to double check and be explicit that we're hotfixing.

app/main/posts/modify/post-tabs.directive.js
```diff
@@ -4,16 +4,15 @@ PostVerticalTabs.$inject = [];
 function PostVerticalTabs() {
     return {
         restrict: 'E',
-        replace: true,
```
rjmackay: Why did you drop this?

app/main/posts/modify/post-tabs.directive.js
```diff
@@ -48,11 +48,10 @@ function PostVerticalTabsController(
     }
 
     function stageIsComplete(stageId) {
-        return _.includes($scope.post.completed_stages, stageId);
+        return $scope.post.completed_stages.indexOf(stageId) > -1;
```
rjmackay: This doesn't fix the bug. It still both fails to save the stage as completed, and fails post validation

rjmackay:
Here's a video of the failures: http://quick.as/almCd1wY

app/main/posts/modify/post-tabs.directive.js
willdoran: The directive was failing to load scope for the page, the replace: true was the cause. The reason the completion toggle was not working was because it was not calling through to the function.

app/main/posts/modify/post-tabs.directive.js
willdoran: That's odd but work for me.

willdoran: This wasn't intended to fix the bug itself, this fixed a secondary bug that appeared once the directive scope loaded. The stageIsComplete function was caught in a loop.

willdoran:
@rjmackay hmmm, the issue was that I removed the check for stage completion and didn't put it back, I'm not sure why but I must have committed it by mistake. It was super weird because it was working in my local and I couldn't reproduce your issue until I was looking back through the commit and noticed I removed it :/ 

So it works correctly, it was correctly publishing but the edit page was not updating the toggle.

rjmackay:
> It was super weird because it was working in my local and I couldn't reproduce your issue 

Well ... this just makes it more obvious how good it is to QA during review I guess.

rjmackay:
Merged. @willdoran lets review develop on Monday and check there isn't anything else we should be pushing into a production release too

PR #700, opened 2017-07-31
Refactor endpoints to use a generic endpoint class [WIP]

rjmackay:
- Create a generic endpoint class
- [ ] Convert existing endpoints to use this
- [ ] Add events to endpoint class

Testing checklist:
- [ ]

Refs ushahidi/platform#1922

Ping @ushahidi/platform

### DIFF
--- a/app/common/common-module.js
+++ b/app/common/common-module.js
@@ -24,6 +24,7 @@ angular.module('ushahidi.common', [
 .directive('ushSlider', require('./notifications/slider.directive.js'))
 
 // API Endpoint wrappers
+.factory('EndpointFactory', require('./services/endpoints/endpointFactory.js'))
 .service('ConfigEndpoint', require('./services/endpoints/config.js'))
 .service('UserEndpoint', require('./services/endpoints/user-endpoint.js'))
 .service('FormEndpoint', require('./services/endpoints/form.js'))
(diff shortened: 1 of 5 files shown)


rjmackay:
@willdoran @Angamanga if you get a chance I'd appreciate early thoughts on this.
The goal is that this would replace almost all endpoints with a simple `EndpointFactory('tags', 'tags/:id'...)` which would handle all caching etc.

Angamanga:
@rjmackay This is neat for many endpoints, but what happens to those who have their own specific functions that is modifying data for that specific endpoint? For example post-endpoint: https://github.com/ushahidi/platform-client/blob/develop/app/common/services/endpoints/post-endpoint.js#L54 And also the login/logout-events that is listened for in for example tags-endpoint: https://github.com/ushahidi/platform-client/blob/develop/app/common/services/endpoints/tag.js#L70 Would they still have their own endpoints or can we extend from this factory in some way?

rjmackay:
> but what happens to those who have their own specific functions that is modifying data for that specific endpoint? For example post-endpoint: https://github.com/ushahidi/platform-client/blob/develop/app/common/services/endpoints/post-endpoint.js#L54 

Ideally you'll be able to pass these in as part of the config. I'll have to see how to handle caching with those though.

> And also the login/logout-events that is listened for in for example tags-endpoint: https://github.com/ushahidi/platform-client/blob/develop/app/common/services/endpoints/tag.js#L70 

I think we could either continue to handle that on a case by case basis, *or* simply make all endpoints refresh on login/logout.

rjmackay:
I'm going to abandon this on the assumption we'll replace endpoints entirely during react migration.
@crcommons just double checking thats your plan eventually?

crcommons:
@rjmackay yes, I'll add it as a specific issue.

PR #702, opened 2017-08-02
Remove race conditions in TranslationService.getLanguage()

rjmackay:
- Make sure we wait for all requests to return before resolving
  the getLanguage promise. Otherwise its not clear if site config
  or user config takes precedence.
- Also accept site config as a parameter

Testing checklist:
- [x] Switch languages and check it works as expected
- [x] Switch languages, and reload the page... does language reset?
- [x] Configure conflicting languages in site and user config... does user config take precedence?

Ping @ushahidi/platform

### DIFF
--- a/app/common/services/translation.service.js
+++ b/app/common/services/translation.service.js
@@ -42,20 +42,31 @@ function (
         });
     };
 
-    var getLanguage = function () {
+    var getLanguage = function (config) {
         return $q(function (resolve, reject) {
                 if (Session.getSessionDataEntry('language')) {
                     resolve(Session.getSessionDataEntry('language'));
+                    return;
                 }
+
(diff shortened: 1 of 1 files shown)


rjmackay:
@jshorland this needs a test

app/common/services/translation.service.js
```diff
@@ -42,20 +42,31 @@ function (
         });
     };
 
-    var getLanguage = function () {
+    var getLanguage = function (config) {
```
Angamanga: All works fine but I not sure I understand this config. I cannot find anywhere getLanguage is called with a config?

rjmackay: It was there so I could call get language within the config endpoint. Only happened within some custom Kuwait code.

PR #705, opened 2017-08-04
Uchaguzi 2017

willdoran:
- Adds logging for Change Tracking in Uchaguzi
- Adds ability to lock Posts
- Adds ability for Admins to break locks
- Add ability to mark Tasks as internal use only

Testing checklist:
- Locking Posts:
  - [ ] Edit a Post, note the id of the Post from the browser URL bar, save this you'll need it later
  - [ ] In a separate tab go to the same Post's detail page
    - [ ] There should be no edit pencil
    - [ ] There should be no edit option in the menu on the right hand side of the post
    - [ ] There should be no delete option in the menu on the right hand side of the post
  - [ ] Logout 
    - [ ] Login as a User of Member Role
    - [ ] Go to the detail page for the Post you previously entered edit mode for
    - [ ] In the browser url input, append /edit to the end of the url and hit enter
    - [ ] You should see an error slider telling you the Post is locked
  - [ ] Logout
    - [ ] Login as an admin
    - [ ] visit the detail page
    - [ ] append /edit to the end of the url and hit enter
    - [ ] you should be given the option to break the lock
    - [ ] click ok and you should be taken to the edit page
  - [ ] Logout
    - [ ] Login as a member user
    - [ ] Edit a Post
    - [ ] In a separate tab, visit the detail page for the same post
    - [ ] In the original tab, click save
    - [ ] Back to the second tab, refresh
        - [ ] You should now see edit options on the detail page

- Tasks visibility control
  - [ ] Add a new Task to a survey or edit an existing survey task, click the configure tab
    - [ ] Set the internal use only option
    - [ ] Save and open the survey again
    - [ ] Ensure the option is still set
  - [ ] Login in as a member user
     - [ ] Ensure that when adding a Post for Survey that has an internal use only Task, that the Task is not visible
  - [ ] Logout    
     - [ ] Login in as admin or use with manage post permission
     - [ ] Edit the new post, ensure that the user can see the task, edit and save it.

Fixes ushahidi/platform# .

Ping @ushahidi/platform


rjmackay:
Closing this PR since it has ALL the changes. We can reopen PRs for individual features.

PR #708, opened 2017-08-06
#1862  - UX changes to add post title

wanjar:
Removing title class so that the form element that holds the title text
can fall in line with other elements in the form.


wanjar:
[clabot:check]

wanjar:
@kamaulynder here it is.

willdoran:
@wanjar Can you remake this PR against the branch uchaguzi-2017, please

PR #710, opened 2017-08-07
1862 Layout Changes to the Basic Post Title Field

wanjar:
Improved the layout of the post title in the add post page. Achieved by removing the title class from the element that holds the title text. Removing the title class removes associated 0 margins and aligns the title element with other elements on the form.

Refs ushahidi/platform#1862


wanjar:
@willdoran, @kamaulynder, here is the PR. Result of edit:
screenshot 5

rjmackay:
@wanjar Looks solid to me. I agree that the `title` style isn't really working well for users, merging this in!

PR #715, opened 2017-08-07
remove requirement to obtain token before sending request

tuxpiper:
- the client no longer eagerly obtains a token for interacting with the API. As per ushahidi/platform#1982 , several endpoints are available now to receive GET operations without requiring access token
- as some sort of transition strategy (in case some necessary API endpoints used by guests still require a token), if the request fails with 400 or 401 and a token is not available, the client then resorts to obtaining a token.

Ping @ushahidi/platform


rjmackay:
@willdoran I think this is solid enough to land on develop. But I'd like another set of eyes.

rowasc:
screen shot 2017-08-09 at 11 20 53 pm

`Branches: cache-api in platform & cache-api in platform-client`
1. TRY to log in , but use invalid credentials. 
2.  get a 400 error. 
3. app crashes with this problem (it keeps going into  `if (!ongoingRequest) {` in line 53)


Checking the code but I wanted to post in the meantime to try and get this validated with someone else..

rowasc:
Since login is a scenario where you would need to let this fail and go through the normal flow.. I believe the solution to this issue would be something like 
`if (config.url.indexOf('oauth/token') !== -1) {
            config.ignorable = true;
        }`
I think (line ~76, function request(config) { of the interceptor)

tuxpiper:
thanks @rowasc for catching that! I've pushed a fix that marks oauth/token requests as ignorable in case there are errors.

rowasc:
@willdoran  I think it's failing because the backend isn't deployed. There's a matching PR in platform

willdoran:
The tests for client should be independent of the API.

willdoran:
@tuxpiper I think this is ready to merge

tuxpiper:
Related to ushahidi/platform#1982 . @willdoran Since it was originally put together pretty rushedly, I think it wouldn't hurt to give it some nice focused look. I vote to move back to dev backlog.

rjmackay:
@tuxpiper I'm going to disagree: this should land ASAP because we'll need it for 2chaguzi and it improves speed of other deployments too.

rowasc:
This is looking good. I've poked around a bit trying to break it, and also tried to reproduce the error I had found initially, besides using it all day in general for testing this + working. 

PASSES.  It's so good to have fewer irrelevant errors while using the site.

PR #716, opened 2017-08-08
ushahidi/platform#1988 support google places API for geocoding

tuxpiper:
- uses google places API for name geocoding if window.googleMapsApiKey is defined

Fixes ushahidi/platform#1988 .

Ping @ushahidi/platform

### DIFF
--- a/app/common/services/geocoding.js
+++ b/app/common/services/geocoding.js
@@ -19,9 +19,35 @@ function (
                 }
             );
 
+    var google_places_search = {
+        query: function (location_name) {
+            var deferred = $q.defer();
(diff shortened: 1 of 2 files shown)


rjmackay:
@willdoran @Angamanga I think with a little clean up this could go into core, though we should default to OSM still.

willdoran:
I would totally agree, we need to add a settings field to allow them to enter an api key and to control which mapping setup they use.

rjmackay:
I'd actually err to keeping this as an envvar for configuration. Its not something users should need to know about, but something we can use for enterprise deployments when needed.

I'll test this doesn't break anything when disabled then merge if its fine.

PR #720, opened 2017-08-09
Maintain saved search or collection state even when navigating away

rjmackay:
Refs ushahidi/platform#1960

Ping @ushahidi/platform


rjmackay:
UI for this needs some work before it goes back to core. Its card of hard to tell what the current state is (ie. which saved search I'm viewing) and how to get back to the default view

rjmackay:
Merged into Uchaguzi, but might need some polish before going into develop

rjmackay:
@willdoran @Angamanga this needs a review. But also a sanity check on the behaviour... does it make sense as a user?

PR #726, opened 2017-08-16
Displaying correct number of posts in survey-filter

Angamanga:
- Displays the number of posts in the survey-filter-bar based on current filtering
- Excludes archived posts in the total count except when filtering for archived posts (imported from a uchaguzi-commit)

Testing checklist:
- [x] Go to views/map or views/list
- [x] Check the numbers in the survey-filter-bar
- [x] Choose a filter or saved search
- [x] The number in the survey-filter-bar should update according to the new search

Fixes ushahidi/platform#1997 and ushahidi/platform#1967  .

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/views/mode-context-form-filter.directive.js
+++ b/app/main/posts/views/mode-context-form-filter.directive.js
@@ -26,12 +26,16 @@ function ModeContextFormFilter($scope, FormEndpoint, PostEndpoint, TagEndpoint, 
     $scope.getUnmapped = getUnmapped;
     $scope.changeForms = changeForms;
     activate();
+    $scope.$watch('filters', function () {
+            getPostStats();
+        }, true);
 
     function activate() {
         // Load forms
         $scope.forms = FormEndpoint.queryFresh();
         $scope.tags = TagEndpoint.queryFresh();
         var postCountRequest = PostEndpoint.stats({ group_by: 'form', status:  ['draft', 'published'], include_unmapped: true });
+
         $q.all([$scope.forms.$promise, postCountRequest.$promise, $scope.tags.$promise]).then(function (responses) {
             if (!responses[1] || !responses[1].totals || !responses[1].totals[0]) {
                 return;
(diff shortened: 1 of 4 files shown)


app/main/posts/views/mode-context-form-filter.directive.js
```diff
@@ -26,12 +26,16 @@ function ModeContextFormFilter($scope, FormEndpoint, PostEndpoint, TagEndpoint,
...
+    $scope.$watch('filters', function () {
+            getPostStats();
+        }, true);
 
     function activate() {
         // Load forms
         $scope.forms = FormEndpoint.queryFresh();
         $scope.tags = TagEndpoint.queryFresh();
-        var postCountRequest = PostEndpoint.stats({ group_by: 'form', status: 'all', include_unmapped: true });
+        var postCountRequest = PostEndpoint.stats({ group_by: 'form', status:  ['draft', 'published'], include_unmapped: true });
```
rjmackay: Shouldn't this initial request use the initial filters?

app/main/posts/views/mode-context-form-filter.directive.js
```diff
@@ -86,6 +90,28 @@ function ModeContextFormFilter($scope, FormEndpoint, PostEndpoint, TagEndpoint,
             }
         });
     }
+    function getPostStats(query) {
+        query = query || PostFilters.getQueryParams(PostFilters.getFilters());
+        var queryParams = _.extend({}, query, {
+            'group_by': 'form',
+            include_unmapped: true
+        });
+        // we want stats for all forms, not just the ones visible right now
```
rjmackay: I'd suggest we drop the tags filter too. It seems like it'd get confusing since categories show in the sidebar underneath forms.

app/main/posts/views/mode-context-form-filter.directive.js
rjmackay: Or could you make this simply use `getPostStats` so we're not duplicating code?

app/main/posts/views/mode-context-form-filter.directive.js
Angamanga: @rjmackay I made some changes and dropped the tags-filter

app/main/posts/views/mode-context-form-filter.directive.js
Angamanga: Fixed, am using getPostStats for everything now.

app/main/posts/views/mode-context-form-filter.directive.js
```diff
@@ -27,11 +27,25 @@ function ModeContextFormFilter($scope, FormEndpoint, PostEndpoint, TagEndpoint,
     $scope.changeForms = changeForms;
     activate();
 
+    $scope.$watch('filters', function () {
+            getPostStats(PostFilters.getFilters()).$promise.then(function (results) {
```
rjmackay: Seems like the contents of `then` could move to function ie. `updateCounts(stats)` which would be called here and below in `activate`.. I'll poke at this and see if I can push this into todays release

rjmackay:
QA result:
- Updating counts beside forms worked perfectly when messing with the status filter
- However because we removed the form and tags filters from the status request, the unmapped count doesn't change when we change the form or category filters :(

Angamanga:
@rjmackay I think the unmapped count has a separate bug. When I filter on something, the unmapped stat from the api returns a negative value.

Angamanga:
@rjmackay I gone through QA once more. The post-number in each survey is correct. The unmapped value is not. And it is behaving a bit weird since it doesn't count the unknown posts in the map-view, but when we go to the timeline to look at the unmapped posts, it does. I added a separate issue for the unmapped problem (https://github.com/ushahidi/platform/issues/2031) so this could be merged now.

PR #728, opened 2017-08-23
Task visibility

willdoran:
This pull request makes the following changes:
-

Testing checklist:
- [ ]

Fixes ushahidi/platform#1945

Ping @ushahidi/platform

PR #731, opened 2017-08-23
Visual feedback to tell users which tasks and fields are public

willdoran:
- Show lock next to private survey fields
- Show message at the top of public tasks

Testing checklist:
- [x] Submit to a survey with public tasks, check for info message
- [x] Submit to a survey with private attribute, check for lock

Fixes ushahidi/platform#1979

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/modify/post-tabs.html
+++ b/app/main/posts/modify/post-tabs.html
@@ -2,6 +2,9 @@
     <header class="form-sheet-summary">
         <h1 class="form-sheet-title section-title">{{stage.label}}</h1>
     </header>
+    <div ng-show="stage.show_when_published" class="alert">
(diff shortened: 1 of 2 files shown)


app/main/posts/modify/post-tabs.html
```diff
@@ -2,6 +2,14 @@
     <header class="form-sheet-summary">
         <h1 class="form-sheet-title section-title">{{stage.label}}</h1>
     </header>
+    <div ng-show="stage.task_is_internal_only" class="alert">
+        <p>This task is marked for internal use only. It is not seen by users when they create a post.</p>
```
rjmackay: Needs to be translatable

rjmackay:
This should probably be part of #728

PR #733, opened 2017-08-24
Adding name instead of id to categories in tasks

Angamanga:
- Displaying name instead of Id to categories in tasks

Testing checklist:
- [ ] Add a post to a survey with categories in a task and save the post. The categories selected should be displayed as names, not as ids when looking at the saved post.

Fixes ushahidi/platform#2022 .

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/detail/detail.html
+++ b/app/main/posts/detail/detail.html
@@ -56,11 +56,12 @@
 
             <div ng-if="form_attributes[key].input === 'tags'">
                 <h2 class="form-label">{{form_attributes[key].label}}</h2>
+                <p>
                     <svg class="iconic" style="margin-right:4px;">
                         <use xlink:href="/img/iconic-sprite.svg#tag"></use>
                     </svg>
                     {{formatTags(post.values[key])}}
-                    </span>
+                </p>
             </div>
         </div>
         <div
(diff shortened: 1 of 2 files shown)


app/main/posts/detail/detail.html
```diff
@@ -116,15 +117,25 @@ <h3 class="listing-heading" translate="app.tasks">Tasks</h3>
           </div>
           <div
             ng-repeat="(key,value) in post.values"
-            ng-if="form_attributes[key].type !== 'media' && !isPostValue(key) && form_attributes[key] && showType(form_attributes[key].type)"
             ng-show="form_attributes[key].form_stage_id === visibleTask"
             >
             <post-value
+              ng-if="form_attributes[key].type !== 'media' && !isPostValue(key) && form_attributes[key] && showType(form_attributes[key].type)&& form_attributes[key].input !== 'tags'"
```
rjmackay: Can we move this if condition to a function on the controller?
In actually looks like `showType` might already be doing that and the other conditions should get merged in?

app/main/posts/detail/detail.html
```diff
@@ -116,15 +117,25 @@ <h3 class="listing-heading" translate="app.tasks">Tasks</h3>
...
               key="key"
               value="value"
               attribute="form_attributes[key]"
               type="'standard'">
             </post-value>
+             <div class="listing-item-primary" ng-if="form_attributes[key].input === 'tags'">
```
rjmackay: I wonder if this should actually be part of the `post-value` directive instead?

rjmackay:
@Angamanga check out https://github.com/ushahidi/platform-client/pull/732 ... I think it may be fixing the same issue?

Angamanga:
@rjmackay Ohnoe, yes it does. Closing this...

PR #734, opened 2017-08-29
Added ability to select all categories and expand categories list

Angamanga:
- creates and installs a reusable categories selector
- adds 'select all' option to category checklist when creating a post
- adds 'select all' option to category checklist when creating a survey
- adds ability to expand categories list when creating a post
- adds ability to expand categories list when creating a survey
- hides 'add category' link when add category form is visible

Testing checklist:

- [x]  When I click 'More' I can see the full list of available categories.
- [x]  When I click "Select All", all the categories in the list of available categories are selected.
- [x] When I manually select all the categories, "Select All" is automatically checked.
- [x] When I select a child category and the parent category is not selected, it selects the parent category automatically.
- [x] When I select a child category the parent-category-checkbox is disabled in a selected state
- [x] When I deselect all child categories, the parent-category-checkbox becomes enabled again.

Fixes ushahidi/platform#1800.
Fixes ushahidi/platform#1801.

Ping @ushahidi/platform


Angamanga:
I have a few remaining style-issues that needs to be resolved:
The overflow-toggle only works with <fieldset> or .form-fieldgroup and that creates a border on the left side of the category-section: 
screen shot 2017-08-29 at 18 26 23
When I try to remove the border temporary with inline styles, the positioning of the checkboxes and the heading above the categories gets weird:

screen shot 2017-08-29 at 18 24 57

@middle8media @rjmackay Does this require a change in the pattern-library or can I change it in the way I use the over-flow-toggle/custom-field types? The over-flow-toggle is the one used on other places where there are checkboxes.

app/settings/surveys/attribute-editor.html
```diff
@@ -31,35 +31,12 @@
...
-            <div
-              class="form-field checkbox"
-              ng-if="category.children"
-              ng-repeat="child in category.children">
-                <label>
-                <input
-                  type="checkbox"
-                  checklist-model="editAttribute.options"
-                  checklist-value="child.id"
-                  ng-click="selectChild(child)"
```
rjmackay: Can we remove the `selectChild` function from the directive now?

app/main/posts/modify/post-value-edit.html
```diff
@@ -201,49 +201,17 @@
...
-          ng-if="option.parent_id === null"
-          >
-        <label>
-          <input
-                type="checkbox"
-                checklist-model="post.values[attribute.key]"
-                name="values_{{attribute.id}}_{{option.id}}"
-                checklist-value="option.id"
-                value="option.id"
-                ng-click="selectParent(option, attribute.key)"
```
rjmackay: Can we remove this from the post-value-edit directive now?

app/common/directives/category-selector.directive.js
```diff
@@ -0,0 +1,118 @@
...
+        // filter out child categories posing as available parent
+        $scope.available = _.filter($scope.available, function (category) {
+            return category.parent_id === null;
+        });
+
+        flattenCategories();
+        $scope.changeCategories();
+    }
+
+    function flattenCategories() {
```
rjmackay: Isn't this basically undoing the work done in SurveyEditor in loadAvailableCategories? and similarly in post-editor...

app/common/directives/category-selector.directive.js
```diff
@@ -0,0 +1,118 @@
+module.exports = CategorySelectorDirective;
+
+CategorySelectorDirective.$inject = [];
+
+function CategorySelectorDirective() {
+    return {
+        restrict: 'E',
+        scope: {
+            available: '=',
```
rjmackay: I suspect we could move some of the category loading and parent/child matching code out of other directives and in here too.

Rather than sending full categories into this directive, pass in an (optional) array of ids. Then load all categories from the TagsEndpoint and filter then in this directive. Right now we do this in the post-editor and its kinda messy.

Maybe thats too big a refactor though? Up to you...

app/settings/surveys/attribute-editor.html
Angamanga: fixed

app/common/directives/category-selector.directive.js
Angamanga: Fixed, but it needs a second look now...

app/main/posts/modify/post-editor.directive.js
```diff
@@ -155,18 +154,6 @@ function PostEditorController(
                         })
                         .filter()
                         .value();
-
-                    // adding category-objects to children
-                    _.each(attr.options, function (category) {
```
rjmackay: I think this also had the effect of filters the tags to only those specified in attr.options. Have we lost that now? or does it still happen elsewhere?

app/settings/surveys/attribute-editor.html
```diff
@@ -31,35 +31,12 @@
...
-                />
-                {{child.tag}}
-                </label>
-              </div>
-        </div>
-      </div>
+        <label translate>Which categories should be available</label>
+        <category-selector
+            available="availableCategories"
```
rjmackay: This is now *all* tags right?

app/main/posts/modify/post-value-edit.html
```diff
@@ -201,49 +201,17 @@
...
-    </div>
-
-      <add-category
-        form-id="post.form.id"
-        attribute="attribute"
-        post-value="post.values[attribute.key]"
-      ></add-category>
+    <div ng-switch-when="tags">
+        <category-selector
+            available="attribute.options"
```
rjmackay: Isn't this now only the ID of each category? not the full object? I don't see anywhere where you load the full tag inside category-selector... or did I miss it?

app/main/posts/modify/post-editor.directive.js
Angamanga: It happens in post-value-edit-directive instead.

app/settings/surveys/attribute-editor.html
Angamanga: yes, availableCategories is ALL tags and selected = the ones in attribute.options

app/main/posts/modify/post-value-edit.html
Angamanga: its happening when loading the tags in the post-editor

PR #737, opened 2017-09-01
[WIP] 1677: Added a 'disable caption' option

rowasc:
issue https://github.com/ushahidi/platform/issues/1677 
### This PR makes it possible to not have the caption field in the post creation form. 
screen shot 2017-09-01 at 6 18 11 pm

### This pull request makes the following changes:
- Added a 'disable caption' option in the UI so that admins can choose to have/not have captions in their media.
- Added captionDisabled field in config array to toggle on/off the input for media caption field
- UI/DATA:  hides captions when the user disables the caption field. (Jess's feedback) . 

### TODO tasks
* [x] more testing instructions should be added.
* [ ] check tests for this part of the app, if any. => Note : I don't think the e2e suite is running. I will discuss the implementation with Robbie today and decide on tests. Probably some unit tests for the captionDisabled field feature would be the best way to go.
* [ ] go through notes, get help or feedback => I will wait on Robbie's feedback before I move any further 
* [x] UI: add a note for users telling them that they can re-enable the captions at any time by turning them back on. (Jess's feedback) . 
* [x] UI/DATA:  hide captions when the user disables the caption field. (Jess's feedback) . 

### Notes
* Should I rename the captionDisabled field to "captionEnabled" ? It always feels a bit weird to have a name like `fieldnameDisabled`, but I don't know what the best way to go would be here. Is captionEnabled even an improvement as a name for this config option?  
     * * IMPORTANT:* If we go for captionEnabled we will have to migrate and by default set it up as true, otherwise it would break existing deployments. That should be a very simple migration but I'm noting it here so I don't forget. 

* About the `FIXME` note inside the file attribute-editor.directive.js , line 15: I tried some ideas to solve this from the backend to ensure the data was consistent and not need this line, but the backend seems to be forcing any empty values for that field into an empty array because of the *json flag in the getDefinition and the 'backendoptionJsonAlwaysArray' option on the  `transformJson` method of the data transformer. Do we have some way to make sure the backend transformer creates an object literal instead of an array in that field? 

### Testing checklist:

- [ ] When I add an image field to a survey, I have the option to disable its **caption** . 
    -    Add an image field 
    -    Use the toggle UI to disable the caption for that image field.
    -    Save the field
    -    save the survey 
    -    try to add a post for that survey type.
          -  [ ] you will see an image field with no caption field. 
- [ ] When I **edit** a survey's image field and disable its caption, I can no longer add a caption for that field when I add new posts.
    -   Edit a survey, and edit a pre-existing image field.
    -   Use the toggle UI to disable the image field caption
    -    Save the field
    -    save the survey 
    -  [ ]   try to add a post for that survey type.
          - [ ]  you will see an image field with no caption. 

- [ ] When I disable an image's caption and there's already data for it, I stop seeing the caption in pre-existing posts.
    -    Edit a survey, and edit a pre-existing image field.  That field should have data already preloaded for the images and the images' captions.
    -    Use the toggle UI to **disable** the caption field.
    -    Save the field.
    -    Save the survey 
    -    Select a post (from the same survey) that has an image _with a caption_, on the field you just disabled the caption for.
          - [ ] you will see an image but not a caption

- [ ] After disabling an image's caption for image fields with pre-existing data, I can decide to re-enable the caption field and see the caption data again.
    -    Edit a survey, and edit a pre-existing image field where the caption was disabled. That field should have data already preloaded for the images and the images' captions.
    -    Use the toggle UI to **enable** the caption field.
    -    Save the field.
    -    Save the survey 
    -    Select a post (from the same survey) that has an image _with a caption_, on the field you just disabled the caption for.
          - [ ] you will see an image AND the caption that the image had before it was disabled.
          
Fixes ushahidi/platform#1677 .

Ping @ushahidi/platform

### DIFF
--- a/app/common/locales/en.json
+++ b/app/common/locales/en.json
@@ -199,6 +199,7 @@
         "title" : "Title",
         "field_allowed_relation_survey": "Allowed surveys",
         "disable_caption_field": "Disable caption field",
+        "disable_caption_field_explanation": "This will disable all captions for this field, but it won't delete them from the database. You will be able to re-enable them later.",
         "description" : "Description",
         "task_description" : "Task description",
         "post_fields" : "Post fields",
--- a/app/common/locales/en.json
+++ b/app/common/locales/en.json
@@ -198,6 +198,7 @@
         "add_task" : "Add Task",
         "title" : "Title",
         "field_allowed_relation_survey": "Allowed surveys",
+        "disable_caption_field": "Disable caption field",
         "description" : "Description",
         "task_description" : "Task description",
         "post_fields" : "Post fields",
--- a/app/main/posts/detail/detail.html
+++ b/app/main/posts/detail/detail.html
@@ -51,6 +51,7 @@
 
           <post-media-value
             ng-if="form_attributes[key].type === 'media'"
+            media-caption-disabled="form_attributes[key].config.captionDisabled"
             label="{{form_attributes[key].label}}"
             media-id="value"></post-media-value>
         </div>
--- a/app/main/posts/detail/post-media-value.directive.js
+++ b/app/main/posts/detail/post-media-value.directive.js
@@ -4,7 +4,8 @@ module.exports = ['MediaEndpoint', '_', function (MediaEndpoint, _) {
         replace: true,
         scope: {
             mediaId: '=',
-            label: '@'
+            label: '@',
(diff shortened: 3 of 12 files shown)


rowasc:
Closing because codeship is not running for forks. See new PR at https://github.com/ushahidi/platform-client/pull/740

PR #740, opened 2017-09-04
1677 caption field optional

rowasc:
issue https://github.com/ushahidi/platform/issues/1677 
### This PR makes it possible to not have the caption field in the post creation form. 
screen shot 2017-09-01 at 6 18 11 pm

### This pull request makes the following changes:
- Added a 'disable caption' option in the UI so that admins can choose to have/not have captions in their media.
- Added captionDisabled field in config array to toggle on/off the input for media caption field
- UI/DATA:  hides captions when the user disables the caption field. (Jess's feedback) . 

### TODO tasks
* [x] more testing instructions should be added.
* [x] go through notes, get help or feedback => I will wait on Robbie's feedback before I move any further 
* [x] UI: add a note for users telling them that they can re-enable the captions at any time by turning them back on. (Jess's feedback) . 
* [x] UI/DATA:  hide captions when the user disables the caption field. (Jess's feedback) . 

### Notes
* Should I rename the captionDisabled field to "captionEnabled" ? It always feels a bit weird to have a name like `fieldnameDisabled`, but I don't know what the best way to go would be here. Is captionEnabled even an improvement as a name for this config option?  
     * * IMPORTANT:* If we go for captionEnabled we will have to migrate and by default set it up as true, otherwise it would break existing deployments. That should be a very simple migration but I'm noting it here so I don't forget. 

* About the `FIXME` note inside the file attribute-editor.directive.js , line 15: I tried some ideas to solve this from the backend to ensure the data was consistent and not need this line, but the backend seems to be forcing any empty values for that field into an empty array because of the *json flag in the getDefinition and the 'backendoptionJsonAlwaysArray' option on the  `transformJson` method of the data transformer. Do we have some way to make sure the backend transformer creates an object literal instead of an array in that field? 

### Testing checklist:

- [x] When I add an image field to a survey, I have the option to enable its **caption** . 
    -    Add an image field 
    -    Use the toggle UI to enable the caption for that image field.
    -    Save the field
    -    save the survey 
    -    try to add a post for that survey type.
          -  [x] you will see an image field with a caption field. 
- [x] When I **edit** a survey's image field and disable its caption, I can no longer add a caption for that field when I add new posts.
    -   Edit a survey, and edit a pre-existing image field.
    -   Use the toggle UI to disable the image field caption
    -    Save the field
    -    save the survey 
    -  [x]   try to add a post for that survey type.
          - [x]  you will see an image field with no caption. 

- [x] When I disable an image's caption and there's already data for it, I stop seeing the caption in pre-existing posts.
    -    Edit a survey, and edit a pre-existing image field.  That field should have data already preloaded for the images and the images' captions.
    -    Use the toggle UI to **disable** the caption field.
    -    Save the field.
    -    Save the survey 
    -    Select a post (from the same survey) that has an image _with a caption_, on the field you just disabled the caption for.
          - [x] you will see an image but not a caption

- [x] After disabling an image's caption for image fields with pre-existing data, I can decide to re-enable the caption field and see the caption data again.
    -    Edit a survey, and edit a pre-existing image field where the caption was disabled. That field should have data already preloaded for the images and the images' captions.
    -    Use the toggle UI to **enable** the caption field.
    -    Save the field.
    -    Save the survey 
    -    Select a post (from the same survey) that has an image _with a caption_, on the field you just disabled the caption for.
          - [x] you will see an image AND the caption that the image had before it was disabled.
          
Fixes ushahidi/platform#1677 .

Ping @ushahidi/platform


rjmackay:
> Should I rename the captionDisabled field to "captionEnabled" ? It always feels a bit weird to have a name like fieldnameDisabled, but I don't know what the best way to go would be here. Is captionEnabled even an improvement as a name for this config option?

I'd go for something like `hasCaption`. It gives more literate code like ie `if (attribute.config.hasCaption)`

rowasc:
@rjmackay ok yes,I like that better.  I'll refactor this and send the platform API PR with a migration then.

rowasc:
@rjmackay this could use a code review if you have some time. Pushing the backend in a few minutes.
Thanks

rowasc:
Ref https://github.com/ushahidi/platform/pull/2039

app/main/posts/detail/detail.html
```diff
@@ -48,9 +48,9 @@ <h1 class="form-sheet-title survey-title">
           ng-if="isPostValue(key) && form_attributes[key] && showType(form_attributes[key].type)"
           class="postcard-field"
           >
-
           <post-media-value
             ng-if="form_attributes[key].type === 'media'"
+            media-has-caption="form_attributes[key].config.hasCaption"
```
rjmackay: Does this fail weirdly if `config.hasCaption` doesn't exist?

app/settings/surveys/attribute-editor.html
```diff
@@ -79,6 +79,22 @@
              <label class="tgl-btn" for="switchnewattribute"></label>
           </div>
        </div>
+
+       <!-- start: media upload toggle to make caption an optional field-->
+       <div class="form-field switch" ng-show="canDisableCaption()">
+         <label translate="survey.disable_caption_field">Disable Caption field</label>
+         <p data-fieldgroup-target="field-caption-disabled-warning" class="init active" translate="survey.disable_caption_field_explanation" ng-show="!editAttribute.config.hasCaption">
+           <svg class="iconic">
```
rjmackay: Does the SVG here show up? I think it'll get wiped out by the translate substitution

app/common/locales/en.json
```diff
@@ -198,6 +198,8 @@
         "add_task" : "Add Task",
         "title" : "Title",
         "field_allowed_relation_survey": "Allowed surveys",
+        "disable_caption_field": "Disable caption field",
```
rjmackay: I'd reverse this, similar to the actual config value ie. `Show image caption` or `Collect image caption`

rowasc: Ah. OK yes, that makes sense.

app/main/posts/detail/detail.html
rowasc: nope. The only reason it could throw an error would be if config was NULL for the media field, which isn't the default behavior for the config field so it really isn't worth cluttering the code with a ternary in there. Also, IF at some point we had messed up and the config could be NULL in some databases, our migration solves that.

app/settings/surveys/attribute-editor.html
rowasc: No it doesn't and we didn't need the SVG there. That was a bad copy paste by me... I was copying the style from the required field and messed up.

app/common/locales/en.json
rowasc: mmmm... I'm asking Jess about this. I'm not sure I like the explanation bit when it's reversed (and I don't know what to change it for)

app/settings/surveys/attribute-editor.html
rowasc: NOte: (I'll send this fix over after Jess's feedback)

rowasc:
@rjmackay updated after  yours and Jess's feedback.

rjmackay:
Feedback from testing:
- Everything works well!
- I'd think we should default to including a caption (ie. hasCaption: true) on new image fields

rowasc:
>I'd think we should default to including a caption (ie. hasCaption: true) on new image fields
ok yes, I'll add that.

rowasc:
@rjmackay done.

PR #744, opened 2017-09-06
Adding datasources to mode-context-bar

Angamanga:
- Adding datasources to the mode-context bar
- Removing the unknown data-bucket
- Updating the survey-filter count based on source-filter

*Testing checklist:*
**as logged out:**
- [ ] When I look at the map or timeline, buckets for each datasource that have posts should appear

**as logged in:**
- [ ] When I look at the map or timeline, buckets for all enabled datasources should appear, even if the have no posts yet.
- [ ] All posts from a source, both published and assigned to a form as well as unpublished (previously in 'unknown'-bucket) should be visible
**when clicking on the dropdown for each datasource in the mode-context-bar**:
- [ ] An edit-button should be visible, when clicking on it, it takes you to the datasource-settings-page
- [ ] An option to show only posts not assigned to a survey, when clicking, it filters for that data-source and form.id = 'none'

**as logged in or logged out:**
- [ ] When choosing 'show-only', only posts from this datasource appear
- [ ] When choosing 'hide', all posts from this datasource is hidden

Fixes ushahidi/platform#1646 .

Ping @ushahidi/platform


Angamanga:
@rjmackay  I removed the datasources from the filter-module, it felt inconsistent since the forms and categories are not there.

jshorland:
@Angamanga to answer your questions:
1. Do we want all posts in the source-buckets, published, unpublished and assigned to forms/not assigned to forms? Or only unpublished that are not assigned to forms (the ones that previously went into the unknown-bucket)? I prefer the first option, but still unsure. What is least confusing?
- I think having ALL posts in the source buckets makes most sense, and then people can drill down as they wish (AKA the normal filter behavior)
2. I have completely removed the 'unknown'-bucket. I think that makes sense, but.. if we put all posts in the source-buckets, do we want to keep the unknown-bucket too? to make it easier to immediately see how many posts there are to administrate?
- Is it possible to keep the unknown bucket and then within that break it down by source? this would be in addition to the source buckets that include ALL posts from that source, and would give people a quick link to posts that they need to structure. 
3. Do we want the numbers of posts in the survey-filter-numbers to be affected of which datasource that is chosen? I removed it now so the number doesn't change even if I choose show-only posts with from a specific datasource (similar to the catagories).
- I believe we have an outstanding issue to actually change the number based on the category that is selected (which, in my opinion makes sense for that number to update as you filter). I'm not sure if it makes most sense to change this now so the number updates, or do it all at once alongside categories (different behaviors between category filtered numbers and data source filtered numbers might confuse the user). Thoughts?

rjmackay:
> I believe we have an outstanding issue to actually change the number based on the category that is selected (which, in my opinion makes sense for that number to update as you filter). I'm not sure if it makes most sense to change this now so the number updates, or do it all at once alongside categories (different behaviors between category filtered numbers and data source filtered numbers might confuse the user). Thoughts?

We fixed the issue where totals didn't update *at all* based on other features, but we excluded categories. Because they're children of the survey, I didn't think it made sense that the total for a survey would change when the category filter changes. However now I'm less sure.

Its confusing because the totals are visible even for a survey/source thats excluded from the current results.

Opened https://github.com/ushahidi/platform/issues/2047 to resolve that.

rjmackay:
> Is it possible to keep the unknown bucket and then within that break it down by source? this would be in addition to the source buckets that include ALL posts from that source, and would give people a quick link to posts that they need to structure.

@jshorland I'm concerned we're going to end up making users figure out how the data model works, so they can find their incoming SMS. As an admin/manager I just want a clear filter for 'Incoming SMS, Incoming Tweets etc'... 
If we don't break unknown down by source they have to "select Unknown + some data source".
If we do break unknown down by data source... then I have 2 places I can filter by SMS and its potentially confusing/not clear how they interact.

Perhaps we should just have a toggle when filtering by source for "Only show posts not assigned to a survey yet" ... rather than forcing users to understand what Unknown means?

willdoran:
@Angamanga Let me know when the final tests are in and I can review and run through your QA list then we can move it to testing.

rjmackay:
@Angamanga do you think you can resolve this before Uchaguzi starts? @jshorland this might be waiting on your feedback?

Angamanga:
@rjmackay Starting this again today...

So a summary of the replies above (i.e this is what I will be doing, @jshorland please have a look...): 
1. All posts go into the buckets, not just the incoming, unstructered ones
2. I skip the unknown-bucket and add an option/toggle to "show only posts not assigned to a survey yet" for each datasource
3. I'll update the numbers in the survey-buckets to reflect the choosen datasource now, and we'll revisit this when working with the filters/numbers in the issues: https://github.com/ushahidi/platform/issues/2047 and https://github.com/ushahidi/platform/issues/2035

Angamanga:
@jshorland The sms-provider was disabled in qa and now the sms-bucket is visible 💃. A question, when choosing to display only messages not assigned to a survey, should I always be moved to the timeline or only when no posts have a location?

PR #748, opened 2017-09-08
1816 rtl css switch [Do not merge]

rjmackay:
- Make RTL switching work by using namespaces CSS
- Disable ExtractText Plugin for CSS, means JS is bundled with CSS
- Clean up language switcher code
- Relies on https://github.com/ushahidi/platform-pattern-library/pull/146

Testing checklist:
- [ ]

Fixes ushahidi/platform#1816

Ping @ushahidi/platform

### DIFF
--- a/app/common/services/translation.service.js
+++ b/app/common/services/translation.service.js
@@ -28,7 +28,17 @@ function (
                 Languages.then(function (languages) {
                     let language = languages.find(l => l.code === langKey);
 
-                    $rootScope.rtlEnabled = language.rtl;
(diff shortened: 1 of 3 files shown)


rjmackay:
@Angamanga I'm not really happy with this, because it means the CSS is no longer split into a separate file but I couldn't make webpacks ExtractText plugin work with 2 separate CSS files.
I think I might need to update webpack and then try again, but this works right now

rjmackay:
Looks like this works in dev mode, but fails in a production build. Digging further

PR #751, opened 2017-09-13
Hotfix rtl switch

rowasc:
- Makes platform-client compatible with platform-pattern-library rc20 

Testing checklist:
- [ ] Open the client with pattern library rc15. Navigate to the map. 
- [ ] Open the client with pattern library rc20. Navigate to the map. 
- [ ] Both versions should work and load the styles. 
- [ ] Both versions should show the loading bar and not an error while loading 

Screenshot of what happens if you don't have this fix but apply Rc20 

screen shot 2017-09-13 at 1 15 48 pm

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/package.json
+++ b/package.json
@@ -116,7 +116,7 @@
     "ngGeolocation": "rjmackay/ngGeolocation",
     "nvd3": "^1.8.4",
     "underscore": "^1.7.0",
-    "ushahidi-platform-pattern-library": "3.7.1-rc.15"
+    "ushahidi-platform-pattern-library": "^3.7.1-rc.20"
   },
   "engines": {
     "node": ">=4.0"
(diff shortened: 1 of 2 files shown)


rowasc:
cc : @Angamanga @crcommons @middle8media this should fix the problem we were having until the real feature with RTL is merged to platform-client at least.

package.json
```diff
@@ -116,7 +116,7 @@
     "ngGeolocation": "rjmackay/ngGeolocation",
     "nvd3": "^1.8.4",
     "underscore": "^1.7.0",
-    "ushahidi-platform-pattern-library": "3.7.1-rc.15"
+    "ushahidi-platform-pattern-library": "^3.7.1-rc.20"
```
rjmackay: This should be fixed to a specific version, not "^version". The client tightly tied to exact PL versions

rowasc:
@middle8media you might want to bump this up to rc22 later. Merging this fix in the meantime, so we are using the PL with the namespacing and catching issues.

PR #752, opened 2017-09-13
2044: users should be able to export all posts, not only published posts

rowasc:
- Add all statuses by default when users export posts

Testing checklist:
- [ ] As an administrator or other user with privileges:
     - [ ] Click on Share to open the share toolbar
     - [ ] Click on Export to CSV, confirm. 
     - [ ] When the CSV is downloaded, you should see all posts, not only the published posts 

- [ ] As a non logged in user
     - [ ] Click on Share to open the share toolbar
     - [ ] Click on Export to CSV, confirm. 
     - [ ] When the CSV is downloaded, you should see ONLY published posts included in the CSV

Fixes ushahidi/platform#2044 .

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/views/share/post-export.directive.js
+++ b/app/main/posts/views/share/post-export.directive.js
@@ -19,7 +19,8 @@ PostExportController.$inject = [
     'ConfigEndpoint',
     'Notify',
     '$q',
-    'PostFilters'
+    'PostFilters',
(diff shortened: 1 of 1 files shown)


rowasc:
Closed this PR because I don't think it's the right approach. I'll be sending  a new one soon

PR #753, opened 2017-09-14
2045 + 2044 - Fix export filters

rowasc:
## This pull request makes the following changes:
- Adds missing filter params to the post share directive and share menu directive. Post export was expecting a filter that was not being sent.
- very basic refactor to make this more testable 
- same basic tests added for filter values and the directive itself

## Testing checklist:
- As an administrator, I can export all posts. 
     - [ ] Login as an administrator or another user with privileges:
     - [ ] Open the timeline view
     - [ ] Click on Share to open the share toolbar
     - [ ] Click on Export to CSV, confirm. 
     - [ ] When the CSV is downloaded, you should see all the posts that you see in the app, not only the published posts.

- [ ] As an anonymous user, I can export only posts that I am allowed to see.
     - [ ] Access the app as an anonymous user
     - [ ] Open the timeline view
     - [ ] Click on Share to open the share toolbar
     - [ ] Click on Export to CSV, confirm. 
     - [ ] When the CSV is downloaded, you should see ONLY published posts included in the CSV.

- [ ] As any type of user, I can only export posts according to the filters applied in the UI.
     - [ ] Access the app as any user
     - [ ] Open the timeline view
     - [ ] Apply filters (examples)
         - [ ] Open the filter modal and apply the "Under review" filter. 
         - [ ] Select only 1 survey type in the left panel
     - [ ] Click on Share to open the share toolbar
     - [ ] Click on Export to CSV, confirm. 
     - [ ] When the CSV is downloaded, you should see only the posts that you selected. 

Fixes ushahidi/platform#2044.

Fixes ushahidi/platform#2045.

Ping @ushahidi/platform


rowasc:
@rjmackay  The root cause of not having filters working at all was the same for 2044 and 2045. There's an additional issue I'm working on fixing that seems to come form the unmapped posts, but I'm not so sure it's particular to exports.
Anyway, just some context on why this is 2 tickets in 1 PR ...

rowasc:
OK actually I think this should be OK to merge. Just realized that the totals in the sidebar don't match the total posts exported (which matches the filters), or even the amount of posts in the database. 

I think that's a different problem entirely. Will look into reporting this as a separate ticket with more details. Posting here for context. 

screen shot 2017-09-13 at 11 39 15 pm
screen shot 2017-09-13 at 11 38 24 pm

rowasc:
@tuxpiper adding you since you were following up on the issue for Diego as well. Any thoughts? (on the issue described above, or on the PR itself :) )
Thanks

rjmackay:
> Ok... I just realized that the totals in the sidebar don't match the total posts exported (which matches the filters), but the posts exported match the number of posts that I have in the database.

I think thats probably an unrelated bug. I thought we fixed it on develop but maybe not completely.
Is it sending the same filters in the HTTP request to the API?

rowasc:
The only parts that are different are related to sorting, format and limits. 
**In the posts request:** 
limit:20
offset:0
order:desc
orderby:post_date
**In the csv export request:** 
format:csv
In the example I just tested, there are 18 posts looking directly at the database that should be returned. 
When I run the export, I get 18 posts.
When I look at the timeline filters, they indicate that I should see 11. **I see 9**
When I look at the result of the posts request, it says I should see 9. **I see 9.**
http://192.168.33.110/api/v3/posts?form%5B%5D=1&form%5B%5D=5&form%5B%5D=2&has_location=all&limit=20&offset=0&order=desc&orderby=post_date&status%5B%5D=published&status%5B%5D=draft

rowasc:
Tests are mostly around making sure there's default filters sso that we don't run into issues with export not sending all required data as it used to be for 2044, and that the directive and main functions are being called.

rowasc:
Conflcits fixed. 
@willdoran

willdoran:
Reviewing now

app/main/posts/views/share/post-export.directive.js
```diff
@@ -28,54 +29,85 @@ function PostExportController(
...
+        var site = ConfigEndpoint.get({ id: 'site' }).$promise;
+        var query = getQuery();
+        var exportQuery = PostEndpoint.export(query);
+        requestExport(site, query, exportQuery);
+    }
+
+    function requestExport(site, query, exportQuery) {
+        $q.all([site, exportQuery]).then(function (response) {
+            showCSVResults(response, query.format);
+        }).finally(function (response) {
```
willdoran: Can we add a failure state that use Notify.errors to update the user for a fail? Since this is pretty standard everywhere we should be able to just pass the error response directly to the notify function on failure.

app/main/posts/views/share/post-export.directive.js
```diff
@@ -28,54 +29,85 @@ function PostExportController(
...
+            format: format
+        });
+        return query;
+    }
 
-                var filename = response[0].name + '-' + (new Date()).toISOString().substring(0, 10) + '.' + format,
-                    data = response[1].data;
+    function showCSVResults(response, format) {
+        // Save export data to file
+        var filename = response[0].name + '-' + (new Date()).toISOString().substring(0, 10) + '.' + format,
```
willdoran: In the future, we should have the API do this piece of work.

app/main/posts/views/share/post-export.directive.js
```diff
@@ -28,54 +29,85 @@ function PostExportController(
...
 ) {
     $scope.loading = false;
-    $scope.exportPosts = exportPosts;
-
-    function exportPosts() {
+    this.getQuery = getQuery;
+    this.prepareExport = prepareExport;
+    this.showCSVResults = showCSVResults;
+    this.requestExport = requestExport;
+    $scope.exportPostsConfirmation = function () {
```
willdoran: Based on the existing pattern this should become 

this.exportPostsConfirmation = exportPostsConfirmation;
function exportPostsConfirmation () {

though it might be handy to have a quick way to discern functions that are used by the template, function used only by the controller and functions used by both.

I'm kind of conflicted but for the sake of consistency changing it might be better. What do you think?

app/main/posts/views/share/post-export.directive.js
```diff
@@ -28,54 +29,85 @@ function PostExportController(
...
+        // Prepare filters for export
+        var query = angular.extend({}, PostFilters.getQueryParams($scope.filters), {
+            format: format
+        });
+        return query;
+    }
 
-                var filename = response[0].name + '-' + (new Date()).toISOString().substring(0, 10) + '.' + format,
-                    data = response[1].data;
+    function showCSVResults(response, format) {
```
willdoran: This is not a block to merging but it would be good to add a test covering this function.

app/main/posts/views/share/post-export.directive.js
```diff
@@ -28,54 +29,85 @@ function PostExportController(
...
+    };
+    function prepareExport() {
+        loadingStatus(true);
+        var site = ConfigEndpoint.get({ id: 'site' }).$promise;
+        var query = getQuery();
+        var exportQuery = PostEndpoint.export(query);
+        requestExport(site, query, exportQuery);
+    }
+
+    function requestExport(site, query, exportQuery) {
```
willdoran: This is not a block to merging but it would be good to add a test covering this function.

app/main/posts/views/share/post-export.directive.js
rowasc: I disagree that this would need to become this.exportPostsConfirmation , but agree that maybe we just want to be consistent here. And in that scenario, I would go with using $scope.exportPostsConfirmation instead of this.exportPostsConfirmation .

app/main/posts/views/share/post-export.directive.js
rowasc: ok

rowasc: @crcommons I'll also implement the feedback from Jess directly here if you don't mind .

rowasc: https://github.com/ushahidi/platform-client/pull/755

app/main/posts/views/share/post-export.directive.js
rowasc: While I agree with the sentiment, and I'd always choose to have higher coverage, I believe with *unit tests*, all we can test here in requestExport is how a  mock of $q responds to an error or success response, since there is no logic whatsoever in the callbacks?. 
Any help with how to cover this function correctly would be appreciated. 

How I see it:
We could add tests separately for showCSVResults, as you mentioned, and for loadingStatus to check that $scope.loading is correct and that notifyProgress or notify were called, which would be more significative and give us more insights on how things are working. This is a good example of why we would want to have good integration tests in addition to unit tests. :/ .

app/main/posts/views/share/post-export.directive.js
rowasc: agree

app/main/posts/views/share/post-export.directive.js
rowasc: agree, do you want to open a ticket for that?

willdoran: Yep, I'll capture it.

app/main/posts/views/share/post-export.directive.js
willdoran: I think you can use this guy
https://github.com/ushahidi/platform-client/blob/8f8ad88a3a92eb107f4da32de03d1104354a7ace/app/main/posts/common/post-actions.service.js#L25

app/main/posts/views/share/post-export.directive.js
willdoran: Agree

jshorland:
Notes from QA test:

Functionality is great and passes QA. 

I'm not sure if it has to do with this particular fix or not, but there's a problem with the CSV export in progress message. The text's line spacing is off. It should also have a period at the end. 


screen shot 2017-09-28 at 13 22 27

jshorland:
@rowasc one more thing -- it doesn't seem to export Twitter handles, sms numbers, or email addresses as author (which I'm assuming would be user_id in the export) even though I'm an admin with full privileges. 

Can you look into this? cc: @Shadrock

rowasc:
Yea, I can look into this two things. 
1. The CSV message is probably on me (most likely I broke it  while merging 👎 ), I'll send over a  PR soon. 
2. The address , twitter handles and sms numbers are not part of what I handled  in the scope of this ticket (this was about filters and the draft/published/etc statuses ), but I'll look into it ASAP, since I've been doing the CSV stuff 👍  I'll open a separate ticket now for this point in particular so we stay organized :)
Thanks Jess for all the feedback!

jshorland:
wonderful! thanks @rowasc

Shadrock:
Thanks @jshorland. I don't know if this will help @rowasc but here's a bit more detail on what I'm seeing as a user.

I've just exported a .csv from the deployment, while being logged in as a user. Prior to download, I unchecked all filters. When looking at the .csv I see no Twitter handles and very few Twitter profile links. It appears that some records contain a URL to the Tweet itself; these are in column S or "media link." Ideally, the Twitter user's handle would be populated into a column for contact information *without* that user needing to be logged into the platform.

PR #754, opened 2017-09-14
toggling between noui and ui on embedded map

Angamanga:
- adds a specific route for embedded map with no ui
- makes map-view directive toggle between classes `map-only` and `full-size` depending on path

Testing checklist:
- [ ] go to /map/noui
- [ ] a map without ui-elements should be visible
- [ ] try embed the map with the route /noui (you can test it here: http://jsfiddle.net/angamanga/Lqe6as41/), only the map without ui-elements should be visible

Fixes ushahidi/platform# .

Ping @ushahidi/platform


Angamanga:
@tuxpiper Makes sense! I fix that now.

PR #755, opened 2017-09-19
1565 export feedback

crcommons:
-Adds a slider/pop-up in the "Share" modal when user selects export to CSV to improve UI feedback
-Adds progress/loading bar within slider/pop-up when export is in progress
-Adds "complete" slider/pop-up when export is complete. 

Testing checklist:
Select "Share" -> "Export to CSV" -> "OK"
- [ ] Slider should appear with "Export in progress..." message and loading bar at the top
- [ ] Slider should appear with "Your CSV export is complete." message once download has completed

Fixes ushahidi/platform#1565

Ping @ushahidi/platform

### DIFF
--- a/app/common/notifications/notify.service.js
+++ b/app/common/notifications/notify.service.js
@@ -6,6 +6,7 @@ Notify.$inject = ['_', '$q', '$rootScope', '$translate', 'SliderService', 'Modal
 function Notify(_, $q, $rootScope, $translate, SliderService, ModalService) {
     return {
         notify: notify,
+        notifyProgress: notifyProgress,
         error: error,
         errors: errors,
         errorsPretranslated: errorsPretranslated,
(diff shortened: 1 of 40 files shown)


app/common/notifications/notify.service.js
```diff
@@ -27,6 +28,14 @@ function Notify(_, $q, $rootScope, $translate, SliderService, ModalService) {
         $translate(message, translateValues).then(showSlider, showSlider);
     }
 
+    function notifyExport(message, translateValues) {
```
willdoran: Could this become more generic, something like notifyProgress, if the only diff between it and notify is the setting of the loading flag?

willdoran: Or it could be notifyLoading if using the loading flag or notifyProgress and change loading to inProgress

crcommons: Done!

jshorland:
@crcommons QA feedback: 
- can we add additional clarifying text to the "export in progress.." message that says "You can navigate away from this window, we'll let you know when your download is complete." I wasn't sure if I could close out of that share window for fear of stopping the download (but i tested and realized i could)
- When I close out of the share window while download is still in progress, the download finishes, receive the success message slider, see that the file has appeared in my browser downloads, and then hit the "share" button again, I see the "in progress" text again, which shouldn't be there. Screenshot attached. This only happens when I close out of the share window when download is still in progress.
screen shot 2017-09-21 at 6 46 08 pm

Otherwise, looks great!

PR #757, opened 2017-09-20
2078 - saving the user_Id when creating a saved search or collection. 

rowasc:
This pull request makes the following changes:
- Adds user_id to the saved search 

Testing checklist:
- [ ] Create a saved search 
- Verify that database table `sets`  saving the user_id

Fixes ushahidi/platform#2078 .

Ping @ushahidi/platform

PR #758, opened 2017-09-20
53 List only my own saved searches and Featured saved searches

rowasc:
- Saved search should only be listed to its owner or if Featured
- Removed unused users array

Requirements covered here: 
- The default saved search listing should show 'featured' saved searches and my own saved searches, but not everyone's saved searches
- I can view someone else's saved search (but I don't get a list of them, they'd need to send me the URL)

Testing checklist:
- I can see featured saved searches as any user
    - [ ] Set a saved search as featured (need to be logged in as admin) 
    - [ ] Logout. 
    - [ ] Open the saved search modal. You should see only featured saved searches.

- I can see my only my own saved searches and featured searches
    - [ ] Login
    - [ ] Create a saved search
    - [ ] Open the saved search modal. You should see only your saved searches and any featured saved search 

Fixes ushahidi/platform-fix-our-environment#53.

Ping @ushahidi/platform


rowasc:
@rjmackay  https://github.com/ushahidi/platform-client/pull/757 should be merged before this ticket since this ticket depends on it.

app/main/posts/savedsearches/saved-search-modal.directive.js
```diff
@@ -33,18 +31,9 @@ function SavedSearchModalController($scope, $element, $attrs, $rootScope, $locat
...
-                // if this search has a user, and its not the currentUser
-                if (_.isObject(search.user) && search.user.id !== _.result($rootScope.currentUser, 'userId')) {
-                    // Load the user (if we haven't already)
-                    if (!users[search.user.id]) {
-                        users[search.user.id] = UserEndpoint.getFresh({id: search.user.id});
-                    }
-                    // Save user info onto the search itself
-                    search.user = users[search.user.id];
-                }
+            $scope.searches = _.filter(searches, function (search) {
```
rjmackay: This looks ok for now. I'm a little concerned it might not scale with lots of searches ie. because we'll get many many results from the DB. 

Thats probably a problem to save later though!

rowasc: Yup, this won't scale. The original approach from the backend was better, or if that was too much, at least adding a filter that supports OR, which I don't think we have.  I tried sending the request with the filters  in the query string, but it didn't work out because it was looking for featured AND my own posts. We can look into it when you get back I think.

rowasc: > The original approach from the backend was better, 
Actually, that's not right :) I have no idea if it was better and it probably wasn't. The option of the OR filter sounds better in retrospective and aligned with the needs in this ticket
@willdoran you helped me in this ticket originally. What do you think? I'll create an issue based on this so we don't forget about it.

PR #759, opened 2017-09-21
[WIP] propagating locale change to post-metadata directives

kinstelli:
This PR is mainly for discussion with @willdoran, but it makes the following changes:
-  fixes a small issue with dependencies on language-settings
- makes a (possibly hacky) fix for updating a directive on language change 
- explicitly updates moment's locale when the language is changed

(adds some logs to console, to be removed )

This does *not* fix an issue where moment.js's translation files aren't matching up with Transifex's language code names.

Related to ushahidi/platform#2065.

### DIFF
--- a/app/common/directives/language-switch.directive.js
+++ b/app/common/directives/language-switch.directive.js
@@ -12,8 +12,8 @@ function LanguageSwitchDirective() {
         template: require('./language-switch.html')
     };
 }
-LanguageSwitchController.$inject = ['$scope', 'Languages', 'TranslationService'];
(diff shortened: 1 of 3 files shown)


kinstelli:
This works somewhat, but calling getPosts() on a language update isn't the right approach, since it seems to duplicates the list in some cases.

rjmackay:
So, if you look at #748 we're forcing a reload of the current angular route on language changes. That actually has the side effect of avoiding this issue altogether.

If we want to try to make moment strings in the UI update on language change, we might be better to use a angular-moment plugin of some sort instead (as @kinstelli suggested in slack).

PR #761, opened 2017-09-23
ushahidi/platform#2086 make http://... resources scheme relative (//...)

tuxpiper:
Makes "http://" resources to be referenced using scheme relative URLs ("//")

### DIFF
--- a/app/common/services/maps.js
+++ b/app/common/services/maps.js
@@ -24,7 +24,7 @@ function Maps(ConfigEndpoint, L, _, CONST) {
             },
             hOSM: {
                 name: 'Humanitarian',
-                url: 'http://{s}.tile.openstreetmap.fr/hot/{z}/{x}/{y}.png',
+                url: '//{s}.tile.openstreetmap.fr/hot/{z}/{x}/{y}.png',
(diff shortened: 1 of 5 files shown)


rowasc:
- [ ] //{s}.tile.openstreetmap.fr/hot/{z}/{x}/{y}.png
- [ ] lorempixel.com/400/300
https://www.gravatar.com/avatar/%7B%7B%20post.user.gravatar%20%7C%7C%20'00000000000000000000000000000000'%20%7D%7D?d=retro&s=40 
- Gravatar post metadata
- Gravatar post toolbox
- Gravatar post messages

rowasc:
1. the html where lorempixel is doesn't look like its being used. 
2. there are 2 other places where gravatar is used, not relative (but https)
app/common/directives/mode-bar/mode-bar.html
and 
 app/settings/users/users.html.

tuxpiper:
Rebasing to master for a hotfix . It looks like this could be breaking geolocation in chrome + iOS .

tuxpiper:
Discarding in favour of #762 (hotfix)

PR #763, opened 2017-09-25
Use a cached location if available ushahidi/platform#2087

rjmackay:
- Use a cached location if available
- Update leaflet.locate control too

Testing checklist:
- Load / reload the site (from the map or timeline view)
- Click on the + sign to add a post to a form that has a location field
- Click the "Use your current location" button that appears when clicking on the location search bar. The location is detected and displayed in the map.
- Go back to the map view by clicking on its icon in the left sidebar
- Click on the + sign to add a post again (to the same form)
- Click the "Use your current location" button again

Fixes ushahidi/platform#2087

Ping @ushahidi/platform


rowasc:
Also: tested it locally, works fine.

PR #768, opened 2017-09-29
795 - Uses the class assignment directly to play nice with Leaflet

Angamanga:
- Makes class-assignment for the map-view not conflict with leaflets class-assignments. 

Testing checklist:
- [ ] Go to map-view as not logged in
- [ ] Log in
- [ ] Map should still be visible without reload

Fixes ushahidi/platform#795 .

Ping @ushahidi/platform


rowasc:
@Angamanga please when merging this PR, choose squash and merge so GitHub gives the chance to change the commit message to something better.  Something like the new title "795 - Uses the class assignment directly to play nice with Leaflet"  I think would be more appropriate.

I used a throwaway message and I wouldn't want that message to go into the mainstream branches, it's confusing for the next person that sees it. (*100% my bad, obviously) 


cc @willdoran

PR #778, opened 2017-10-05
Split screen post-detail

Angamanga:
- Exchange timeline to data in mode-bar
- Adds the split pane-view in /views/data
- If you click on a post, the post-card should animate and appear in the right-hand-pane

Testing checklist:
- [ ] Look at the mode-bar, there should be a link to /view/data instead of /view/timeline
- [ ] When going to /view/data, the split view-pane should appear and load all posts
- [ ] When clicking on a post, it should animate and the details appear in the right hand pane (currently below all posts though...)
- [ ] Post-details is updated with new information (Survey-title and description if existing)

Fixes ushahidi/platform#1896 (part of it .

Ping @ushahidi/platform


app/main/posts/detail/post-detail-data.directive.js
```diff
@@ -0,0 +1,253 @@
...
+    'FormStageEndpoint',
+    'FormEndpoint',
+    'Maps',
+    '_',
+    'Notify',
+    'moment',
+    'PostSurveyService'
+    ];
+
+function PostDetailDataController(
```
willdoran: Like you pointed out we should look at refactoring this into a shared controller or moving logic elsewhere. But for now it's fine. Can you make an issue to revisit it next week?

Angamanga: https://github.com/ushahidi/platform/issues/2116

PR #781, opened 2017-10-05
Adding basic Task Visibility UI

willdoran:
- Adding new Task Visibility control, to ensure that certain tasks are visibile only to logged in user's with required permissions

Testing checklist:
- [x] In edit mode for a survey, ensure that for a task there there is an option "Task is only for internal use" on the configure section

For the following tests you will need the companion PR for API
- [x] Ensure that survey saves when "Task is only for internal use"
- [x] Ensure that option remains selected on page reload
- [x] Ensure that when an anonymous user tries to create a new survey, where the survey has a task marked internal only, they do not see that task or any of its fields
- [x] Ensure that when a user without Mange Post permission tries to create a new survey, where the survey has a task marked internal only, they do not see that task or any of its fields
- [x] Ensure that when a user with Manage Posts permission tries to create a new survey, where the survey has a task marked internal only, they do see that task and of its fields
- [x] Ensure that when a user with Manage Posts permission tries to edit a new survey, where the survey has a task marked internal only, they do see that task and its fields

Fixes ushahidi/platform#2024

Ping @ushahidi/platform


rowasc:
Looks good. Testing now.

PR #788, opened 2017-10-09
2091 - bufixing

rowasc:

- I can switch between views (Data & Map) without losing my active filter options
    - 2091: Saved search filter remains active when switching between map and data views
    - 2091: post order options filter remains active when switching between map and data views
    - 2091: asc desc order filter remains active when switching between map and data views
    - 2091: unlocked on top remains active when switching between map and data views

- I can write a search query witouth it generating a .get request for each key change while I type. When I click the "Search all " button in the dropdown the search query is enabled once. 
- Adds some basic tests for dropdown directive
- Moves filtertransformers out of active filters directive
- Adds basic tests for active clean filters (diff only)

Testing checklist:

 Add filters (one of each)
- [x] When you switch between map and data views, you should have all your filters still enabled.
- [x] When you write a query in the searchbar, you should only see the loading bar change after clicking the Search all button .

Pending for a different PR: Categories and Surveys cannot be filtered together, there is some logic that replaces them for some reason. I need to look into that and find out if it's an intended feature or a bug

Fixes ushahidi/platform# .

Ping @ushahidi/platform#2091

### DIFF
--- a/app/common/directives/filter-system/filter-searchbar.html
+++ b/app/common/directives/filter-system/filter-searchbar.html
@@ -14,7 +14,7 @@
 
     <div class="searchbar-results dropdown-menu" ng-class="{active: (searchResultsVisible && form.q.$viewValue.length > 0) }">
         <div class="form-field">
-            <button type="submit" class="button-plain">
+            <button class="button-plain" ng-click="enableQuery()">
                 <span class="button-label">
                     <span translate-values="{ entity: entityType }" translate>
                         toolbar.searchbar.search_all_entities_for
--- a/app/common/directives/filter-system/filter-searchbar.js
+++ b/app/common/directives/filter-system/filter-searchbar.js
@@ -4,8 +4,8 @@
  * and selection of appropriate sub directive
  */
 
-module.exports = ['$timeout',
-function ($timeout) {
+module.exports = ['$timeout', 'PostFilters',
+function ($timeout, PostFilters) {
     var link =
         function (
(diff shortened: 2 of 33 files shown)


rowasc:
@willdoran @Angamanga if either one of you has some time for a PR, I would really appreciate it. Thanks!

willdoran:
@rowasc Reviewing now

Angamanga:
Filters works fine, I really like that they are divided further, much cleaner! One thing, not sure if it is only on my local machine, but the dropdown doesn't close after I choose 'apply filters' or 'cancel'

Angamanga:
When I choose a saved-search, it goes back to the timeline, is that the correct behaviour? Or should I end up on data-view? Has it to do with the editing of saved-searches?

rowasc:
>Filters works fine, I really like that they are divided further, much cleaner! One thing, not sure if it is only on my local machine, but the dropdown doesn't close after I choose 'apply filters' or 'cancel'
Apply filters and clear filters now close the dropdown. 
> When I choose a saved-search, it goes back to the timeline, is that the correct behaviour? Or should I end up on data-view? Has it to do with the editing of saved-searches?
Yep, that would be ok I think. 

Thanks! @Angamanga

rowasc:
@willdoran I'm not sure why, but the codeship is stuck there..

willdoran:
@rowasc not sure, you could try restarting the codeship build from inside codeship, @tuxpiper any idea why?

PR #789, opened 2017-10-09
2092 load new posts

crcommons:
- Adds functionality to get new posts every 30 seconds and makes them available for user to see by clicking a button. Button still needs to be styled in PL and in-line styling removed.

Testing checklist:
- [x] Open two browsers. In one, navigate to `timeline`. In the other, add a new post. In the first browser, wait ~30 seconds to see button appear at the top of the DOM indicating you have new posts available. Click button and confirm that new posts are added to top of the list.
- [x] Open two browsers. In one, navigate to `data view`. In the other, add a new post. In the first browser, wait ~30 seconds to see button appear at the top of the DOM indicating you have new posts available. Click button and confirm that new posts are added to top of the list.

- [x] Do this again and add two+ posts; confirm that pluralized text shows correctly (See X new posts vs. See 1 new post).

Fixes ushahidi/platform#2092

Ping @ushahidi/platform


app/main/posts/views/post-view-list.directive.js
```diff
@@ -123,14 +134,7 @@ function PostListController(
...
 
-            // Merge grouped posts into existing groups
-            angular.forEach(groupPosts(postsResponse.results), function (posts, group) {
-                if (angular.isArray($scope.groupedPosts[group])) {
-                    Array.prototype.push.apply($scope.groupedPosts[group], posts);
-                } else {
-                    $scope.groupedPosts[group] = posts;
-                }
-            });
+            groupPosts(postsResponse.results);
```
rowasc: neat refactor.

app/main/posts/views/post-view-list.directive.js
```diff
@@ -242,4 +258,40 @@ function PostListController(
         // Clear selected posts
         $scope.selectedPosts.splice(0);
     }
+
+    function getNewPosts() {
```
rowasc: can this be moved to a service?

app/main/posts/views/post-view-list.html
```diff
@@ -1,6 +1,8 @@
-<div class="main-col">
+<div class="main-col" id='top'>
```
rowasc: is this an id we need because of the PL?

rowasc: if it's not absolutely necessary that this is named 'top', I think a prefixed id like post-view-list-top would be better. The reason is that ids being unique, you want to minimize opportunities for collusion in IDs.

app/main/posts/views/post-view-list.html
```diff
@@ -1,6 +1,8 @@
...
 
-    <div class="listing timeline" ng-class="{ 'toolbar-active' : selectedPosts.length > 0 }">
+    <button class="button-gamma button-flat" ng-click="addNewestPosts()" ng-show="newPostsCount > 0" style="position:fixed; z-index: 10000;">See {{newPostsCount}} new post<span ng-hide="newPostsCount > 1">s</span>.</button>
```
rowasc: this should use translations

app/main/posts/views/post-view-list.html
crcommons: Good point! Done.

app/main/posts/views/post-view-list.html
crcommons: Done

app/main/posts/views/post-view-data.directive.js
```diff
@@ -139,4 +154,40 @@ function PostViewDataController(
...
+    function getNewPosts() {
+        var existingFilters = PostFilters.getQueryParams($scope.filters);
+        var filterDate = moment(existingFilters.date_before).format('MMM Do YY');
+        var now = moment().format('MMM Do YY');
+        if (filterDate >= now) {
+            var mostRecentPostDate = $scope.recentPosts[0] ? $scope.recentPosts[0].post_date : $scope.posts[0].post_date;
+            existingFilters.date_after = mostRecentPostDate;
+            var query = existingFilters;
+            var postQuery = _.extend({}, query, {
+                order: $scope.order,
```
rowasc: @crcommons please use $scope.filters.orderby and $scope.filters.order since those are the ones applied by the user.

app/main/posts/views/post-view-list.directive.js
```diff
@@ -242,4 +258,40 @@ function PostListController(
...
+        var existingFilters = PostFilters.getQueryParams($scope.filters);
+        var filterDate = moment(existingFilters.date_before).format('MMM Do YY');
+        var now = moment().format('MMM Do YY');
+        if (filterDate >= now) {
+            var mostRecentPostDate = $scope.recentPosts[0] ? $scope.recentPosts[0].post_date : $scope.posts[0].post_date;
+            existingFilters.date_after = mostRecentPostDate;
+            var query = existingFilters;
+            var postQuery = _.extend({}, query, {
```
rowasc: @crcommons please use $scope.filters.orderby and $scope.filters.order since those are the ones applied by the user.

rowasc:
@crcommons looks great. Just a tiny comment about filters that I missed yesterday. 
After you solve that + fix conflicts I will test it and approve

rowasc:
- Tested out OK in data view. 
I'd approve it since it works in data view just fine and it looks good,  but I can't complete testing in the list view because it's failing (the view, not the feature). I just confirmed that this is not failing due to issues in this PR, it's something in develop messing things up. 
cc @willdoran what do you think ?

rowasc:
We will look into the issue in the morning and continue from there. No need to merge tonight until we figure things out.

rowasc:
It also sends the same "date after" in all requests. :(

PR #805, opened 2017-10-11
Layout usage improvements

rowasc:
- Improvements to layout in saved searches.

Testing checklist:
- [x] Open a saved search with the default view "data".  Verify that you can see the view data with it. 
- [x] Click on the map view and see your saved search in the map.  Verify that you can see the view map with it. 
- [x] Check that the normal map and data modes are still working
- [x] When clicking return to all posts, it should show the map . (I think?) @Angamanga ? I left collectionView in the scope to set whatever the right view would be. Empty for now and loads the default view instead which is map. 
- [x] Open a collection with the default view "data".  Verify that you can see the view data with it. 
- [x] Click on the map view and see your collection in the map.  Verify that you can see the view map with it. 
- [x] Check that the normal map and data modes are still working
- [ ] When clicking return to all posts, it should show the map . (I think?) @Angamanga ? I left collectionView in the scope to set whatever the right view would be. Empty for now and loads the default view instead which is map. 

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/common/services/view-helper.js
+++ b/app/common/services/view-helper.js
@@ -11,6 +11,10 @@ function (
         {
             name: 'list',
             display_name: $translate.instant('views.list')
+        },
+        {
+            name: 'data',
+            display_name: $translate.instant('views.data')
         }
     ];
 
--- a/app/config.js
+++ b/app/config.js
@@ -5,3 +5,7 @@
 // 	backendUrl : "https://ushahidi-platform-api-release.herokuapp.com",
 // 	mapboxApiKey: ""
 // };
+window.ushahidi = {
+    platform_websocket_redis_adapter_url: '127.0.0.1',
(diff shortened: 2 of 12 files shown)


rowasc:
@Angamanga Please take a look if you can

Angamanga:
> When clicking return to all posts, it should show the map . (I think?) @Angamanga ? I left collectionView in the scope to set whatever the right view would be. Empty for now and loads the default view instead which is map.
I think it actually should show the same view you started on, but just remove the saved search. Which from now always would be in map-mode since data-view doesn't show mode-context-bar so yes, it should show the map

Angamanga:
In collections, when clicking 'return to all posts' it actually goes back to the default-view. So when I was in map, it returned to the data-view with all the posts.

PR #811, opened 2017-10-12
Enabled bulk actions select

rowasc:
- Remove broadcast for bulkactions checkboxes. That is not needed since we have 2 way data binding available there. 
- Small modifications to make the lock and checkboxes appear

Testing checklist:
- [x] I can see the checkboxes when I click bulk actions
- [x] If there is a lock enabled in the post, I see the lock.
- [x] If there is a lock enabled in the post and bulk actions is enabled, I see the lock and the checkbox
- [x] I can select one or many posts with their checkboxes. 
- [x] I can see the bulk toolbar changes the total selected items counter when I check/uncheck posts

TODO Fix checkbox + lock styles

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/views/card.html
+++ b/app/main/posts/views/card.html
@@ -1,16 +1,13 @@
 <article ng-class="['postcard', {selected: inFocus}]" ng-click="clickAction(post)">
     <div class="post-band" ng-style="{backgroundColor: post.form.color}"></div>
-
+    <div class="listing-item-select">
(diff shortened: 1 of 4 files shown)


rowasc:
@crcommons @willdoran first one who gets to it please leave a comment so there's no duplication of work :)  feel free to reassign.
Thanks

crcommons:
Reviewing now

PR #812, opened 2017-10-12
Loading actions in Data View

crcommons:
- Adds loading bar and progress bar to data view when posts are loading
- removes bulk actions text when posts are loading
- gives message when there are no posts or no posts return from filtering

Testing checklist:
- [x] Go to Data View, confirm that loading bar and progress bar appear while posts load
- [x] Confirm "BULK ACTIONS" and "X of X reports" are not visible while posts are loading
- [x] filter posts so that it will not return any (i.e. set filter dates to sometime last year), confirm that "Your search didn't match any posts" appears

Fixes ushahidi/platform# .

Ping @ushahidi/platform


rowasc:
@crcommons I can see the x of x reports and the bulk actions while it's loading

PR #817, opened 2017-10-12
Implementing toolbar and add posts button changes

crcommons:
-implements styling changes for toolbar dropdown and "See New Posts" button. Updates pattern library version.

Testing checklist:
In Data View:

- [x] Change class of "See New Posts" button to ng-show from ng-hide (or wait until new posts come in. Confirm it looks nice. 
- [x] Should be below "Bulk Actions"
- [x] should be fixed so that it stays at top when scrolling (known bug appears once you scroll to bottom. Already filed)

- [x] Open bulk actions. confirm that 'X' button is in top right corner and that it stays in position as you resize browser
- [x] confirm alignment of buttons and text in toolbar stays nice as you resize browser.

Fixes ushahidi/platform# .

Ping @ushahidi/platform


rowasc:
Looks great. One tiny thing, not even sure if it is a bug.When I scroll, the "x of x posts"  message is no longer there, and there's a white space instead. Is that by design/intentional? 

screen shot 2017-10-13 at 5 14 56 pm
screen shot 2017-10-13 at 5 14 53 pm

crcommons:
@rowasc That's the intended design. The header with X of X reports and Bulk actions aren't meant to be fixed/sticky. so as you scroll down, you will see the button and margins but not those other things.

app/main/posts/common/collection-toggle/collection-toggle-button.html
```diff
@@ -2,5 +2,5 @@
     <svg class="iconic">
         <use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="/img/iconic-sprite.svg#grid-three-up"></use>
     </svg>
-    <span class="button-label" translate="set.add_to_collection">Add to collection</span>
+    <span class="button-label hidden" translate="set.add_to_collection">Add to collection</span>
```
rowasc: I'm not sure about this. Why is this being hidden ?

rowasc:
@crcommons ok awesome then. Left a question because I don't really understand that bit

app/main/posts/common/collection-toggle/collection-toggle-button.html
crcommons: Seth had me add it for accessibility reasons.

PR #819, opened 2017-10-13
Warns you with the default browser warning if you have a post that wasn't saved and you are leaving the page

rowasc:
- Warns you with the default browser warning if you have a post that wasn't saved and you are leaving the page

Testing checklist:
Test 1: 
- [ ] Edit a post in the data mode, don't save
- [ ] Reload the page. You should see a warning telling you that you have unsaved changes. 

Test 2: 
- [x] Edit a post in the data mode, don't save
- [x] Hit the browser back button. You should see a warning telling you that you have unsaved changes. 

Fixes ushahidi/platform# .

Ping @ushahidi/platform


rowasc:
On why we can't use nicer modal here. TLDR: if we can truly prevent a user from leaving by using the browser back button or reloading,  it means all sites can, which means you would have a hard time leaving many sites with bad intentions. The smarter explanation is in this SO post :)
https://stackoverflow.com/a/41125558

On why we CAN prevent navigation in other situations: We can use it in internal navigation because we are not preventing them from leaving our site, we are merely preventing them from navigation inside the site. 

@aoduor @jshorland would you take a look at this and let me know if this is something we want to add? I think it adds value, so I'm leaving it here for review.

app/main/posts/common/post-lock.service.js
```diff
@@ -49,7 +49,7 @@ function PostLockService(
         // is different from the User who owns the lock
         if (post.lock) {
             if ($rootScope.currentUser) {
-                return $rootScope.currentUser.userId !== parseInt(post.lock.user.id);
+                return $rootScope.currentUser.userId !== post.lock.user ? parseInt(post.lock.user.id) : 0;
```
willdoran: good catch

app/main/posts/modify/post-data-editor.directive.js
```diff
@@ -101,6 +101,11 @@ function PostDataEditorController(
             $scope.cancel();
         }
     });
+    window.onbeforeunload = function (e) {
+        if ($scope.postForm && $scope.postForm.$dirty) {
+            return 'Are you sure you want to leave this page?  You will lose any unsaved data.';
```
willdoran: @Angamanga create a leavePost() function to handle this, ideally we need to call through to that
1. For consistency
2. For translation of the message.

https://github.com/ushahidi/platform-client/blob/37395583ddec3e845b855342edbc4a2f62f7d4f7/app/main/posts/modify/post-data-editor.directive.js#L97

rowasc:
@crcommons thanks for the feedback. A few comments: 
- the unlock feature bug is not related `at all` to this ticket. The only change I made related to the lock is to make sure we don't keep getting undefined errors from an undefined user object, which screws up the view completely when it happens (if we let it happen, it shows raw angular variables instead of the form labels and values) 
- `When I edit a post and then reload the page, it still reloads without the warning appearing first` : are you changing the post / Are you sure you are testing with a dirty form?(one that has had a change). If so, I can't reproduce it. Can we do a quick hangouts? 

- I will fix the translate text, thanks

rowasc:
@crcommons something else about the lock: safari goes insane when testing this (and some other lock related features)  because it's not on a secure connection and we are using sockets for the locks. I forgot to mention that  :)

rowasc:
@willdoran It looks like it's not taking the message at all. Like, even the hardcoded string. :/ Not sure why. It's supposed to take the string you return. I'll read up and get back to it later in the day since this is not so high priority.

app/main/posts/modify/post-data-editor.directive.js
```diff
@@ -101,6 +101,11 @@ function PostDataEditorController(
             $scope.cancel();
         }
     });
+    window.onbeforeunload = function (e) {
```
rjmackay: This should probably us angular's $window

rowasc:
Closing, this is just polluting the repo at this point. Downloaded a diff and posted it in https://github.com/ushahidi/platform/issues/3684 in case someone wants to work on it

PR #818, opened 2017-10-12
Delete saved searches modal

rowasc:
This pull request makes the following changes:
- Adds a modal to let users delete their saved searches

Testing checklist:
- [x] Click on the edit button next to the saved search select list (in the filters)
- [x] When the modal opens, select one or more saved searches. 
- [x] Click "Delete" . 
- [x] The modal should close and the saved searches should be deleted. 

Fixes ushahidi/platform#2145.

Ping @ushahidi/platform#2145

PR #820, opened 2017-10-13
Saved search loading bars

rowasc:
- Adds a loading bar in saved searches to make sure the user understands that things are still loading in both filters and lists

pre-requirement: https://github.com/ushahidi/platform-client/pull/818 (because HTML structure changes so I wanted to make sure it was compatible)

Testing checklist:
Test 1: 
- [x] Open the saved search list (star icon) . 
- [x] Validate that you see a "loading" bar. 
- [x] If there are no results, you should see a  "No results found" message
- [x] If there are results, you should the normal list of results
Test 2: 
- [x] Open the filters dropdown list .
- [x] Validate that you see a "loading saved searches" bar where the saved search usually lives
- [x] If there are results, you should the normal list of results in the select list 

Fixes ushahidi/platform# .

Ping @ushahidi/platform


kinstelli:
Looks good. And it's always good to see more loading bars.

PR #821, opened 2017-10-13
Removing list view

rowasc:
- Completely remove the list view.

Testing checklist:
- [x] Every usage of list view is now replaced by 'data' view. 
- [x] Links everywhere point to data instead of list 
- [x] If the user attempts to access /views/list we take them to /views/data
- [x] If the user attempts to access /collections/:id/list we take them to /collections/:id/data
- [x] If the user attempts to access /savedsearches/:id/list we take them to /savedsearches/:id/data
Related backend PR: https://github.com/ushahidi/platform/pull/2149
Fixes ushahidi/platform# .

Ping @ushahidi/platform


willdoran:
@rowasc is there an associated API PR? Savedsearches and Collections have a "default" view that the user can select. This view is stored on the Set table
https://github.com/ushahidi/platform/blob/develop/src/Core/Entity/Set.php#L23

We need a simple migration to update existing Sets where view===list change it to data

There is also a view_options field but I'm not aware of it being used in Client at all presently. But it might be worth double checking that.

rowasc:
Add a redirect from list -> data = DONE
Migration: Adding migration in a different PR now. 
view_options : I can't find any usages, except for a couple view_options keys in collections.json that do nothing basically.

rowasc:
Adding a redirect for list view in collections and savedsearches. It would be super annoying otherwise if someone bookmarked it.

rowasc:
@crcommons @willdoran I think I addressed all comments. Let me know if I'm missing something or need to change something else. Thanks so much for the feedback. I had totally missed the migration and redirects

PR #825, opened 2017-10-16
Prevent anon users from seeing lock info

willdoran:
- Adds check to ensure user is logged in before showing a lock message

Testing checklist:
- [ ] As a logged in user lock a post that is published
- [ ] From a separate browser while not logged in ensure that you can not see the lock icon on the Post Card or the lock message on the detail view

Fixes ushahidi/platform#2159

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/common/post-lock.service.js
+++ b/app/main/posts/common/post-lock.service.js
@@ -45,8 +45,14 @@ function PostLockService(
     }
 
     function isPostLockedForCurrentUser(post) {
-        // We only wish to show the lock message when the Current User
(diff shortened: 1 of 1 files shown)


crcommons:
@willdoran This should also prevent regular members who do not have edit privileges from seeing the lock info as well. Appears to be working correctly for anonymous users though.

willdoran:
@crcommons groovy, will change now

willdoran:
Ok this is fixed in API PR the lock is only returned if the user has permission to see it. The check on client side is just additional.

PR #827, opened 2017-10-16
2147 - update the url of the post when clicking on the data post item

rowasc:
- 2147 - update the url of the post when clicking on the data post item. Do not reload as they click. Link to a post detail page if they copy paste the url.

Testing checklist:
Happy path: 
- [x] Open the data view
- [x] Select any post. 
- [x] Check the URL changes to the one of the detail view of the selected post
- [x] Check that the URL works if you copy paste it and enter it into the browser (should take you to the detail view of the post)
Things to check : 
- [x] Bulk actions still works
- [x] Clicking on the checkboxes for bulk actions does NOT change the URL. 
- [x] You can navigate the data view correctly, use the load more button, and still get "new posts" button when there are new posts. 

### TODO: 
- [x] Confirm with Angela that this is what we want to merge . 
- [x] Fix tests

Fixes ushahidi/platform#2147 .

Ping @ushahidi/platform


aoduor:
confirmed @rowasc ;)

app/main/posts/common/post-actions.directive.js
```diff
@@ -75,7 +77,7 @@ function PostActionsDirective(
             } else {
                 $scope.selectedPost = {post: $scope.post};
             }
-            if ($location.path().indexOf('data') === -1) {
+            if ($routeParams.view !== 'data') {
```
Angamanga: Using indexOf covers saved-searches and collections too. If we only check for view and select a post in saved-search or collection-mode, the post opens in post-detail-view

Angamanga:
For the failing tests, routeParams in tests seems to be really messy, if you want to pair we can give it a try, have been messing around with it too.

app/main/posts/common/post-actions.directive.js
rowasc: yes, that makes sense. I'll look into it. Thanks Anna

rowasc:
@Angamanga OK so this should make sense now. Please take a look. Thanks

PR #829, opened 2017-10-16
Edit mode static object with editing=false sent in map

rowasc:
- sends a dummy editmode=false to avoid the share button from hiding in the toolbar while being used in the map view

Testing checklist:
- [x] Check that you see the Share button in the map mode
- [x] Check that you see the Share button in the data mode
- [x] Check that you don't see the share button in the data mode while editing. 
- [x] Check that you don't see the share button in a post's  detail page. 

Fixes ushahidi/platform# . 

Ping @ushahidi/platform


rowasc:
@aoduor I can't find the ticket for this. Is it in GitHub? Thanks

aoduor:
@Romina-Suarez now it is. https://github.com/ushahidi/platform/issues/2169

PR #838, opened 2017-10-17
2172: Clearing search text does not cause search to refresh

rowasc:
- enable the reactive filters momentarily when 'q' is cleared  so that the search query updates correctly

Testing checklist:
Quick tests concerning possible regressions:
- [x] Change filters in data mode and see that they do not get applied until you apply them 
- [ ] Change query filter (search query) in map mode and click "Search all for ", and see it get applied. 
- [ ] Change query filter (search query) in data mode and then click "Apply filters" to see it get applied. 
- [ ] Map mode: Change query filter (search query) without having opened the filters drop-down.  Click on "Search all for" button to see the posts list update.
- [ ] Map mode: Make sure you do not see the map update the query filter as you type. 
- [ ] Data mode: Change query filter (search query) without having opened the filters drop-down.  Click on "Search all for" button to see the posts list update.

Test that it fixes what it says it fixes: 
- [ ] Data Mode: Write a query filter. Click on "x" clear button of the query filter, see the filters update.
- [ ] Map Mode: Write a query filter. Click on "x" clear button of the query filter, see the filters update.

Fixes ushahidi/platform#2172 .

Ping @ushahidi/platform


rowasc:
@crcommons was working on this too, and she is sending in a new fix for the tiny x in the search bar too

willdoran:
@rowasc It looks good, but I haven't tried it locally yet.

app/common/directives/filter-system/filter-searchbar.js
```diff
@@ -27,12 +27,17 @@ function ($timeout, PostFilters) {
...
+            $scope.removeFilter = function (filterKey, value) {
+                PostFilters.clearFilter('q', '');
+            };
         };
     return {
         restrict: 'E',
         replace: true,
         template: require('./filter-searchbar.html'),
         scope: {
+            filters: '=',
```
rjmackay: Do we need this? I don't see it used anywhere.

app/common/directives/filter-system/filter-searchbar.js
```diff
@@ -27,12 +27,17 @@ function ($timeout, PostFilters) {
                 PostFilters.qEnabled = true;
                 PostFilters.reactiveFilters = 'enabled';
             };
+
+            $scope.removeFilter = function (filterKey, value) {
```
rjmackay: This doesn't need params right? it always clears `q`

app/common/directives/filter-system/filter-searchbar.html
```diff
@@ -6,12 +6,16 @@
             <input name="q" type="search" autocomplete="off"
             placeholder="{{ 'toolbar.searchbar.search_entity' | translate:{ entity: placeholderText } }}"
             ng-model="model" ng-blur="hideSearchResults()" ng-focus="showSearchResults()" ng-keyup="showSearchResults()">
-            <svg class="iconic">
+            <svg class="iconic" ng-show="!form.q.$viewValue.length">
               <use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="/img/iconic-sprite.svg#magnifying-glass"></use>
             </svg>
+            <svg class="iconic" ng-click="removeFilter(key, value)" ng-show="form.q.$viewValue.length">
```
rjmackay: are `key` and `value` defined?

app/main/posts/views/post-filters.service.js
```diff
@@ -65,6 +65,15 @@ function PostFiltersService(_, FormEndpoint, TagEndpoint, $q) {
 
     function clearFilter(filterKey, value) {
         filterState[filterKey] = getDefaults()[filterKey];
+        /**
+         * Since q is a special type of filter that gets applied on the click
+         * of a button separate from the input control, and since it should be automatically
+         * enabled when you clear it, we are adding this
+         */
+        if (filterKey === 'q') {
```
rjmackay: I don't really know what reactiveFilters are, but this looks sane :)

rjmackay:
I'm seeing a few things failing when testing on http://qa.2172-clear-queryfilter.ushahidilab.com ...

- Change filters in data mode and see that they do not get applied until you apply them **sort & filter dropdown doesn't open at all**
- Map mode: Make sure you do not see the map update the query filter as you type. **definitely searching as a type**

rowasc:
thanks @rjmackay . 


Answering here. 

-- Change filters in data mode and see that they do not get applied until you apply them sort & filter dropdown doesn't open at all => OK this was a thing with on our merge order. Sorry about that. Basically we needed a PL version for another ticket that got merged before this one, and that messed up this one (in the drop-down  styles PR) .

--- Map mode: Make sure you do not see the map update the query filter as you type. definitely searching as a type => OK I think I just realized what you found here. It's not searching for results but it is updating the stats while you type, which is wrong. Thanks for that. I will look into it now.  


Also yes, your CR is accurate, the thing with clearFilters was me being distracted apparently. No need to have those params/filter var, and the naming is also not good so I changed it  (locally, will push)

rjmackay:
>  Change filters in data mode and see that they do not get applied until you apply them

When I change filters in the dropdown on the map, should they apply immediately? Currently they do.

rowasc:
>When I change filters in the dropdown on the map, should they apply immediately? Currently they do.
Yes, in the map that's the expected behavior.

rowasc:
Ok the dropdown styles we sent  earlier today messed up some things: 
- when clearing filters (x button next to bug icon) the dropdown closes. It should remain open

rowasc:
Running checklist again now , fixed removeFilter (x) button closing dropdown

app/main/posts/views/mode-context-form-filter.directive.js
```diff
@@ -27,9 +27,38 @@ function ModeContextFormFilter($scope, FormEndpoint, PostEndpoint, TagEndpoint,
     $scope.unknown = [];
 
     activate();
-
-    $scope.$watch('filters', function () {
-        getPostStats(PostFilters.getFilters()).$promise.then(updateCounts);
+    /**
+     * TODO this needs refactoring about refactoring the  qenabled/reactivefilters situation
```
rjmackay: I think this is ok for now, but its definitely fairly fragile.

rjmackay:
> When I change filters in the dropdown on the map, should they apply immediately? Currently they do.

This behaviour is weird because there's still an 'Apply filters' button, but now it does nothing?

rowasc:
I agree, It's weird. We are still discussing some of this stuff (filters came up today again)  because in general some things about how they work are not super intuitive.

rowasc:
I  think we could hide the apply filters button in the map so it's less confusing, keep the clear filters button ? 
I'll get an issue started for discussing that in particular.

app/main/posts/views/mode-context-form-filter.directive.js
rowasc: Agree.

PR #840, opened 2017-10-17
Error locking on Edit Page

willdoran:
- Divergence between two version of editor is causing problems, they need to be reintegrated

Testing checklist:
- [ ] On Edit Page for an existing Post ensure that the Lock request succeeds in getting a lock.

Fixes ushahidi/platform# .

Ping @ushahidi/platform


rowasc:
I can't get a lock to show up, but I do see PUT requests for locks going out. Not sure what's going on. I'll check again  later and see if I missed something . On the bright side, the console errors that were happening are gone .

willdoran:
@rowasc sorry yes it is the put that should be happening!

PR #849, opened 2017-10-19
Unhide map when it has data

kinstelli:
- Sets $scope.hideMap = false on reload. This was staying true when each post was selected.

Testing checklist:
- [x] Select a post with a map, verify that the map is shown
- [x] Select a post without a map, verify that the map is hidden
- [x] Select another post with a map, verify that the map is shown 

Fixes ushahidi/platform#2184.

Ping @ushahidi/platform


willdoran:
@kinstelli The first two work for me but the thirds steps doesn't. 
- When I click on a Post with Map then without then with the Map is not shown. It seems once I click on a Post without a Map the Map is hidden there after.

willdoran:
Passes tests!

PR #870, opened 2017-10-23
2210 ASC order should not allow the see new posts button to appear

rowasc:
- do not get new posts when sorting by ASC Order

### Testing checklist:

Test normal behavior: 
- [x] Browser 1: load data view. Don't use sorting options
- [x] Browser 2: create and publish a new post
- [x] Browser 1 : wait ˜ 30 seconds and see the "see new posts"button appear.

Test new edge case: 
- [x] Browser 1: load data view. Use "oldest first" as your sort order
- [ ] Browser 2: create and publish a new post
- [x] Browser 1 : wait >30 seconds and don't see the "See new posts" button appear

Fixes ushahidi/platform#2210 .

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/views/post-view-data.directive.js
+++ b/app/main/posts/views/post-view-data.directive.js
@@ -86,7 +86,7 @@ function PostViewDataController(
     /**
      * setting "now" time as utc for new posts filter
      */
-    var newPostsAfter = moment.utc().format();
+    var newPostsAfter = moment().utc();
     $scope.savingPost = {saving: false};
     activate();
     function activate() {
@@ -208,6 +208,11 @@ function PostViewDataController(
             // @todo figure out if we can store these more efficiently
             Array.prototype.push.apply($scope.posts, postsResponse.results);
 
+            // Use the most recent post date as the date to search for new posts since
+            if ($scope.posts.length > 0 && $scope.posts[0].created) {
+                newPostsAfter = moment($scope.posts[0].created).utc().add(1, 's');
+            }
+
             // Merge grouped posts into existing groups
             groupPosts(postsResponse.results);
 
@@ -320,7 +325,7 @@ function PostViewDataController(
         $scope.selectedPosts = [];
         recentPosts = [];
         $scope.newPostsCount = 0;
-        newPostsAfter = moment.utc().format();
+        newPostsAfter = moment().utc();
     }
 
     function loadMore() {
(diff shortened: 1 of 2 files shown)


rowasc:
Passed test locally (also it's easier to test if you use 3 browsers, but I didn't want to overcomplicate the checklist ha) 

@crcommons LMK

crcommons:
When you are in normal data mode, add a new post and wait until the "See new posts" button appears. Then select sort by: oldest first. The button remains :(

rowasc:
ah . Checking if I understand: 
- load data view, no sort order. 
- add post, see new posts buttons appear
- then without "removing " the button by clicking on it, change order? And that should hide the button ?

crcommons:
Right, exactly.

rowasc:
ugh :( ok , thanks !!!!

rowasc:
@crcommons please review.

app/main/posts/views/post-view-data.directive.js
```diff
@@ -350,8 +350,7 @@ function PostViewDataController(
     function getNewPosts() {
         var existingFilters = PostFilters.getQueryParams($scope.filters);
         var filterDate = moment(existingFilters.date_before).utc().format();
-
-        if (filterDate >= newPostsAfter) {
+        if (filterDate >= newPostsAfter && existingFilters.order !== 'asc') {
```
rjmackay: It might be a little clearer if this was split into its own check at the start of the fn.
```
if (existingFilters.order !== 'asc') {
    $scope.newPostsCount = 0;
}
```

rjmackay:
Tested: 
- [x] Normal behaviour works
- [x] Sorting by asc works as expected
- [ ] Still need to test switching sort order after 'load new' is visible

I found another weird case:
- Load the data view
- Switch to sort: ASC
- Add a new post P1 (in other browser)
- Wait a while
- Switch to sort: desc. List loads, including the new post (P1)
- Add another post P2 (in other browser)
- Wait a while. 'See 2 new posts' appears
- Clicking on 'See 2 new posts' add 2 new posts adds P1 and P2 to the post list. **P1 is now in the list twice.**

rjmackay:
Tested this too:

- load data view, no sort order.
- add post, see new posts buttons appear
- then without "removing " the button by clicking on it, change order

Works but the 'See 1 new post' button is hidden before I click 'Apply filters'.

rjmackay:
@rowasc I think this works now. Could you or @Angamanga retest this and review my changes?

Angamanga:
Tested locally and it now works:
- [ ] normal behaviour as described in first checklist
- [ ] switching to ask when button is visible makes the button disappear
- [ ] Only unloaded posts are visible after changing order from asc to desc and then posting another post are visible

jshorland:
This passes QA in sandbox.uchaguzi.or.ke

PR #871, opened 2017-10-23
2203 : check if lock exists before we try to remove it

rowasc:
- lets the user cancel on unstructured posts

Testing checklist:

#### Testing with no changes: 
- [x] Open an unstructured post (one with  no survey attached)
- [x] Don't make any changes
- [x] Try to cancel. It should show the post detail view (right side in data mode) 

#### Testing with no changes: 
- [x] Open an unstructured post (one with  no survey attached)
- [x] Change the title or some other field without saving. 
- [x] Try to cancel. It should show "Are you sure" modal 

Fixes ushahidi/platform#2203 .

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/modify/post-data-editor.directive.js
+++ b/app/main/posts/modify/post-data-editor.directive.js
@@ -312,13 +312,21 @@ function PostDataEditorController(
     function cancel(url) {
         $scope.isLoading.state = false;
         $scope.savingPost.saving = false;
-        PostLockEndpoint.unlock({
-            id: $scope.post.lock.id,
-            post_id: $scope.post.id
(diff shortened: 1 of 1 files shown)


rowasc:
@willdoran this looks like a patch that should not exist, but I'm not sure about unstructured posts so I went ahead and did it in case it's actually expected. LMK if it's reasonable  /approved or if we need to debug deeper into locks. Thanks

rjmackay:
Tested it out on the QA deployment. Works for me

willdoran:
@rowasc I'll take a look on the API. Not sure why unstructured posts would cause issues with locks immediately so I'll test it out.

PR #872, opened 2017-10-23
Don't reinit the map unless post id changed

rjmackay:
- Don't reinit the map unless post id changed
- Don't watch postId till the map is initialized

Testing checklist:
- [x] Load data view and check for leaflet errors
- [x] Switch to map view - check for leaflet errors

Fixes ushahidi/platform#2189

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/detail/map.directive.js
+++ b/app/main/posts/detail/map.directive.js
@@ -15,14 +15,17 @@ function PostDetailMap(PostEndpoint, Maps, _, PostFilters, L, $q, $rootScope, $c
         var map;
         $scope.hideMap = false;
 
-        $scope.$watch('postId', function (postId) {
-            if (map) {
-                map.remove();
-            }
-            activate();
-        });
-
         activate();
+
+        $scope.$watch('postId', function (postId, oldPostId) {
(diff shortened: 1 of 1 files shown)


rjmackay:
@kinstelli I haven't tested this yet - this is just the change I made on our call.

kinstelli:
I had just been trying this. This part does work, but from testing, it seems that the explicit removal of a map in $scope.$on('$destroy') isn't necessary and also causes issues.

rjmackay:
> I had just been trying this. This part does work, but from testing, it seems that the explicit removal of a map in $scope.$on('$destroy') isn't necessary and also causes issues.

Odd. That should really only happen when switching away from the data view entirely and was really just an attempt to prevent a memory leak. What issues does it cause? Does it get called multiple times? or triggered at the same time as the watcher?

kinstelli:
I agree it's strange, but leaving it in causes the 'Error: Map container is being reused by another instance' when switching from Data view to somewhere else, such as the Map view. 
I was worried about leaks too, but I looked at this in a memory profiler, and I couldn't create a leak, (or I'm not using the profiler right.)

PR #873, opened 2017-10-24
2213 pagination users

rowasc:
- add pagination for users
-  remove pagination html where not needed
- commented out pagination in categories while we clarify the requirement there since it was not working and It's not obvious why it would exist

Testing checklist:
- [ ] Load the users (settings/user) screen.  
- [ ] Load the users (settings/user) screen. If there are more than 10 users, there is pagination 
- [ ] Load the users (settings/user) screen. If there are 10 users or less, there is no pagination 
Nothing new, but worth mentioning: 
- [ ] Load the categories page. There is no pagination
- [ ] Load the activity page. There is no pagination
- [ ] Load the roles page. There is no pagination

Fixes ushahidi/platform#2213.

Ping @ushahidi/platform

### DIFF
--- a/app/settings/users/users.html
+++ b/app/settings/users/users.html
@@ -1,6 +1,5 @@
 <div>
     <div class="mode-context init" dropdown>
-
         <header class="mode-context-header">
             <ol class="breadcrumbs">
                 <li><a href="/" ng-controller="navigation as nav">{{nav.site.name}}</a></li>
@@ -90,7 +89,7 @@
                     </button>
                 </listing-toolbar>
              </div>
-            <nav class="pagination" ng-if="showPagination">
+            <nav class="pagination" ng-show="showPagination">
                 <ul uib-pagination ng-model="currentPage" items-per-page="itemsPerPage" total-items="totalItems" ng-change="pageChanged()" max-size="5" boundary-links="false" rotate="false"></ul>
             </nav>
         </div>
--- a/app/settings/users/users.html
+++ b/app/settings/users/users.html
@@ -90,9 +90,9 @@
                     </button>
                 </listing-toolbar>
              </div>
-
-           <uib-pagination ng-model="currentPage" items-per-page="itemsPerPage" total-items="totalItems" ng-change="pageChanged()" max-size="5" boundary-links="false" rotate="false"></uib-pagination>
-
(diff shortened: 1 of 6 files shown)


app/settings/users/users.html
```diff
@@ -90,9 +89,9 @@ <h2 class="listing-item-title"><a ng-href="/settings/users/{{user.id}}">{{user.r
                     </button>
                 </listing-toolbar>
              </div>
-
-           <uib-pagination ng-model="currentPage" items-per-page="itemsPerPage" total-items="totalItems" ng-change="pageChanged()" max-size="5" boundary-links="false" rotate="false"></uib-pagination>
-
+            <nav class="pagination" ng-show="showPagination">
```
rjmackay: Why move the pagination `<nav>` and `<ul>` out of the template? We're still going to need them everywhere there's pagination right?

rowasc: Right so it's that way because we changes from Angular bootstrap 1.3 to angular bootstrap 2.5, and in the 2.5 version the directive has the be used as an attribute, which messes up everything with the old way of using the template

rowasc: I'm going to take another look and see if I can move it without breaking everything,  I'm just not sure I can

rowasc: I based this on the way angular ui bootstrap handle the pagination templates (https://github.com/angular-ui/bootstrap/blob/baf0a567f9f4641848a9a4699695f9cde5f28f70/template/pagination/pagination.html). It used to be that you could simply call it from the <uib-pagination and replace everything you wanted and the template would look like this : https://github.com/angular-ui/bootstrap/blob/dc85f67b2d25c7ceee16ea3325b2b6c9beee7363/template/pagination/pagination.html because it was an element directive. Now it's using the attribute inside the UL, so you can't do it that other way. The template now represents the different LIs that you can have, not much else. 
I might just be missing something. We can look at this together when you get back if you have a minute.

rjmackay: Thats depressing. Could you just attach it to a generic div? or to the nav element so the `<ul>` at least goes in the template?
If not, we can just merge this as is

jshorland:
FIXED. awesome. passes QA

PR #875, opened 2017-10-24
2228 set the form as dirty when media.changed is true

rowasc:
- This fixes the issue where if you change the image the form still think you didn't change anything .. 

Testing checklist 1:
- [x] Data mode: Edit post. 
- [x] Don't change anything. Hit cancel. 
- [x] It takes you to the detail view of the post in the data mode (no warning)

Testing checklist 2:
- [x] Data mode: Edit post. 
- [x] Change an image. Hit cancel. 
- [x] At this point It should warn you that you made changes.
- [x] Hit save from the warning modal. It should save the post and image. 

Testing checklist 3:
- [x] Data mode: Edit post. 
- [x] Change an image. 
- [x] Hit save. It should save the post and image. 

TODO: when changing images, the left side (card ) should update, not just the right side (detail) of the data view. Right now you need to reload or click on detail to see the new image. 

Fixes ushahidi/platform#2228.

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/modify/post-media.directive.js
+++ b/app/main/posts/modify/post-media.directive.js
@@ -25,6 +25,11 @@ function (
         },
         template: require('./media.html'),
         link: function ($scope, element, attr, formCtrl) {
+            $scope.$watch('media.changed', function (newValue, oldValue) {
+                if (newValue === true) {
+                    formCtrl.$setDirty();
+                }
+            });
             if ($scope.mediaId) {
                 MediaEndpoint.get({id: $scope.mediaId}).$promise.then(function (media) {
                     $scope.media = media;
--- a/app/main/posts/modify/post-media.directive.js
+++ b/app/main/posts/modify/post-media.directive.js
@@ -16,7 +16,7 @@ function (
     return {
         restrict: 'E',
         replace: true,
-        require: '^^form',
+        require: ['^^form', 'ngModel'],
         scope: {
             mediaId: '=',
             media: '=',
@@ -24,12 +24,20 @@ function (
             mediaHasCaption: '='
         },
         template: require('./media.html'),
-        link: function ($scope, element, attr, formCtrl) {
-            $scope.$watch('media.changed', function (newValue, oldValue) {
-                if (newValue === true) {
-                    formCtrl.$setDirty();
+        link: function ($scope, element, attr, requiredAttributes) {
+            var form = requiredAttributes[0];
+            var ngModelForm = requiredAttributes[1];
+            console.log(ngModelForm);
+
+            ngModelForm.$render = renderViewValue;
+
+            function renderViewValue() {
+                if (ngModelForm.$viewValue === 'changed') {
+                    form.$setDirty();
                 }
-            });
+                return ngModelForm.$viewValue;
+            }
+
             if ($scope.mediaId) {
                 MediaEndpoint.get({id: $scope.mediaId}).$promise.then(function (media) {
                     $scope.media = media;
--- a/app/main/posts/modify/post-media.directive.js
+++ b/app/main/posts/modify/post-media.directive.js
@@ -16,51 +16,79 @@ function (
     return {
         restrict: 'E',
         replace: true,
-        require: ['^^form', 'ngModel'],
+        require: 'ngModel',
         scope: {
-            mediaId: '=',
             media: '=',
             name: '@',
-            mediaHasCaption: '='
+            mediaHasCaption: '<'
         },
         template: require('./media.html'),
-        link: function ($scope, element, attr, requiredAttributes) {
-            var form = requiredAttributes[0];
-            var ngModelForm = requiredAttributes[1];
+        link: function ($scope, element, attr, ngModel) {
+            // Initialize media object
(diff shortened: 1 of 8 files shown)


rowasc:
@willdoran IDK, I don't like it and I don't think we should need it. . I'm pretty sure there's something more going on because I found out the image is not being updated on the left side card. I'll leave this here while I continue debugging and figuring out the issues but if we need a fix in the meantime (right now), we can get this merged.

willdoran:
@rowasc Could we instead mimic what is done for the location field and pass ngModel then manage it view the view renderer

https://github.com/ushahidi/platform-client/blob/develop/app/main/posts/modify/location.directive.js#L45

rowasc:
@willdoran ok yea, that's smart! 
Trying this now, I think I can fix the other issue where the left side does not update too if I get the ngmodel to update correctly

rowasc:
@willdoran I just sent an update here.

app/main/posts/modify/post-media.directive.js
```diff
@@ -16,15 +16,27 @@ function (
...
         template: require('./media.html'),
-        link: function ($scope, element, attr, formCtrl) {
+        link: function ($scope, element, attr, requiredAttributes) {
+            var form = requiredAttributes[0];
+            var ngModelForm = requiredAttributes[1];
+
+            ngModelForm.$render = renderViewValue;
+
+            function renderViewValue() {
+                if (ngModelForm.$viewValue === 'changed') {
```
rjmackay: You shouldn't need to do this at all. But you need to use `ngModelForm.$setViewValue();` to save the value

app/main/posts/modify/post-media.directive.js
```diff
@@ -16,15 +16,27 @@ function (
     return {
         restrict: 'E',
         replace: true,
-        require: '^^form',
+        require: ['^^form', 'ngModel'],
         scope: {
             mediaId: '=',
             media: '=',
```
rjmackay: If you're going to use ngModel it'd be better to remove the `media` input, and maybe `mediaId` too. And handle input/output via the ngModel controller

app/main/posts/modify/post-media.directive.js
```diff
@@ -16,15 +16,27 @@ function (
     return {
         restrict: 'E',
         replace: true,
-        require: '^^form',
+        require: ['^^form', 'ngModel'],
```
rjmackay: Take a look at how we handle this in the location directive. You really shouldn't need to require the form controller, just ngModel

app/main/posts/modify/post-media.directive.js
```diff
@@ -16,15 +16,27 @@ function (
...
+            var form = requiredAttributes[0];
+            var ngModelForm = requiredAttributes[1];
+
+            ngModelForm.$render = renderViewValue;
+
+            function renderViewValue() {
+                if (ngModelForm.$viewValue === 'changed') {
+                    form.$setDirty();
+                }
+                return ngModelForm.$viewValue;
```
rjmackay: This needs to actually render the output, ie. update the model used within this view somewhere

app/main/posts/modify/post-media.directive.js
```diff
@@ -16,39 +16,79 @@ function (
     return {
         restrict: 'E',
         replace: true,
-        require: '^^form',
+        require: 'ngModel',
         scope: {
-            mediaId: '=',
             media: '=',
```
rjmackay: I'm actually not super happy we still have shared state with `media` but getting rid of it would require a big refactor

app/main/posts/modify/post-media.directive.js
```diff
@@ -16,39 +16,79 @@ function (
...
             name: '@',
-            mediaHasCaption: '='
+            mediaHasCaption: '<'
```
rjmackay: One way binding, because 2 way bindings are horrible

app/main/posts/modify/post-media.directive.js
```diff
@@ -16,39 +16,79 @@ function (
...
+            $scope.showReplace = showReplace;
+            $scope.showDelete = showDelete;
+            $scope.deleteMedia = deleteMedia;
+
+            activate();
+
+            function activate() {
+                renderViewValue();
+
+                // Watch for media changes
```
rjmackay: Not the ideal way to detect changes, but it works. At some point we should refactor `file-upload` to accept `onChange` callbacks or similar.

rowasc: if you bind onchange on the  `element', it knows that it changed, It's just that the value is 'changed' and not an actual useful value. Obviously, still not ideal, just throwing it here in case it's something you wanted to do instead of watching over the media.changed. Not saying you `should` do  that because the benefit is not really that much .

jshorland:
@rjmackay it passes all of the tests. There's a small issue that the image doesn't update in the left pane list of post cards unless you refresh. But I'm fine if we push as is and create a new issue for that with a low low priority. See gif: http://d.pr/i/fI823Z

rjmackay:
I'd noticed that, but I think it was like that before.

PR #877, opened 2017-10-24
2206 post edit mode

willdoran:
- Fixes the issue where you could not select a Post other than the currently selected Post to edit from the post card action menu

Testing checklist:
- [x] On Data View, select a Post
  - [x] Click edit
  - [x] Click the ... menu on a different Post
    - [x] Click edit
      - [x] You should be shown the edit view for the new post
- [x] On Data View, select a Post
  - [x] Click the ... menu on a different Post
  - [x] Click edit
    - [x] You should be shown the Edit view for that Post
- [x] On Data View, select a Post
  - [x] Click Edit
   - [x] Change the value of one of the fields
     - [x] Click the ... menu on a different Post
     - [x] Click edit
       - [x] You should see a warning that you have unsaved changes
         - [x] Click the X in the top right hand corner
           - [x] You should remain on the current Post
         - [x] Click continue without saving 
           - [x] You should move to the new Post
         - [x] Click save
           - [x] You should move to the new Post
- [x] On Map view
  - [x] Click the ... menu on a Post
  - [x] Click edit
    - [x] You should be taken to the edit page for the Post

Fixes ushahidi/platform#2206

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/modify/post-data-editor.directive.js
+++ b/app/main/posts/modify/post-data-editor.directive.js
@@ -72,9 +72,8 @@ function PostDataEditorController(
   ) {
 
     // Setup initial stages container
-    $scope.post = angular.copy($scope.postContainer.post);
     $scope.everyone = $filter('translate')('post.modify.everyone');
-    $scope.isEdit = !!$scope.post.id;
+
     $scope.validationErrors = [];
     $scope.visibleStage = 1;
     $scope.enableTitle = true;
@@ -95,6 +94,12 @@ function PostDataEditorController(
     $scope.hasPermission = $rootScope.hasPermission('Manage Posts');
     $scope.leavePost = leavePost;
     $scope.selectForm = selectForm;
+
+    // Need state management
+    $scope.$on('event:edit:post:reactivate', function () {
+        activate();
+    });
+
     $scope.$on('event:edit:post:data:mode:save', function () {
         $scope.savePost();
     });
@@ -113,7 +118,8 @@ function PostDataEditorController(
     activate();
 
     function activate() {
-        $scope.editMode.editing = true;
+        $scope.post = angular.copy($scope.postContainer.post);
+        //$scope.editMode.editing = true;
         if ($scope.post.form) {
             $scope.selectForm();
         } else {
@@ -141,6 +147,10 @@ function PostDataEditorController(
      * routePrams or only on the location, I think, so we are going to use this for the moment
      */
     function doChangePage(url) {
+
+        leaveEditMode();
+        signalLeaveComplete();
+
         if (!url) {
             return;
         }
@@ -192,7 +202,7 @@ function PostDataEditorController(
                 // Failed to get a lock
                 // Bounce user back to the detail page where admin/manage post perm
                 // have the option to break the lock
-                $scope.editMode.editing = false;
+                leaveEditMode();
                 return;
             }
 
@@ -320,11 +330,9 @@ function PostDataEditorController(
                 id: $scope.post.lock.id,
                 post_id: $scope.post.id
             }).$promise.then(function (result) {
-                $scope.editMode.editing = false;
                 doChangePage(url);
             });
         } else {
-            $scope.editMode.editing = false;
             doChangePage(url);
         }
     }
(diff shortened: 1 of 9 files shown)


app/main/posts/modify/post-data-editor.directive.js
```diff
@@ -415,19 +429,24 @@ function PostDataEditorController(
             }
             request.$promise.then(function (response) {
                 var success_message = (response.status && response.status === 'published') ? 'notify.post.save_success' : 'notify.post.save_success_review';
+
                 // Save the updated post back to outside context
-                $scope.postContainer.post = response;
+                $scope.postContainer.post = $scope.post;
```
crcommons: I think you need to leave this set to response in order for the postcard to update correctly

rjmackay: This definitely needs to remain as `response` otherwise we get empty fields showing up in the detail view and other weirdness.

app/main/posts/modify/post-data-editor.directive.js
```diff
@@ -415,19 +429,24 @@ function PostDataEditorController(
...
+
+                // DEVNOTE: Not sure how this would ever happen in the case of data view
```
rjmackay: This handling exists because anonymous post submissions receive a 204 with no content, so no post id. Might be worth noting that in a comment

app/main/posts/modify/post-data-editor.directive.js
rjmackay: ^ this

app/main/posts/views/post-view-data.html
```diff
@@ -62,10 +62,9 @@
             <post-card
                 ng-repeat="post in posts"
                 can-select="bulkActionsSelected"
-                post="post.id === selectedPostId ? selectedPost.post : post"
```
Angamanga: Is there a reason why you removed this? It breaks the two-way-data-binding that is needed to update the post-card on the left correctly after editing. Don't need selectedPostId though, works with                 `post="selectedPost.post && post.id === selectedPost.post.id ? selectedPost.post : post"` :)

app/main/posts/modify/post-data-editor.directive.js
willdoran: This is not relevant to the data-editor though since it isn't used to create.

app/main/posts/views/post-view-data.html
willdoran: I'll double check but I think it was because the post was being always overwritten. I'll readd a connection to update the local post!

willdoran: Ah, it's because it was considering it to still be the selected post even when another Post becomes selected, in terms of the applying classes and when a new post moves to edit mode, as in overwrites the current selected post the old post is reverted.

This line applies the SelectedPost to the Post card if the selectedPost === to the Post's id, when that is no longer true it reverts to the original Post state.

app/main/posts/modify/post-data-editor.directive.js
willdoran: Reverted!

kinstelli:
I walked through the checklist and manual tests are passing.

app/main/posts/views/post-view-data.directive.js
```diff
@@ -154,6 +164,11 @@ function PostViewDataController(
         checkForNewPosts(30000);
     }
 
+    function persistUpdatedPost(updatedPost) {
+        _.extend(_.findWhere($scope.posts, { id: updatedPost.id }), updatedPost);
```
rjmackay: The use of extend could get a bit weird here if you remove a value from the post, it might not get removed.
It might be better to use `splice()` instead to swap the entire posts object out of `$scope.posts`

app/main/posts/views/post-view-data.directive.js
```diff
@@ -154,6 +164,11 @@ function PostViewDataController(
...
+        console.log();
```
rjmackay: Stray log?

app/main/posts/views/post-view-data.directive.js
willdoran: Sure

PR #874, opened 2017-10-24
saving values from post-data-editor

Angamanga:
This pull request makes the following changes:
- Changes the way post.values are copied and looped through when cleaning empty values in post-data-editor

Testing checklist:
- [x] go to post-data-view
- [x] select a post with some fields empty
- [x] choose to edit, add values in the empty fields
- [x] save
- [x] values should be saved and visible in the post-detail-view

Fixes ushahidi/platform#2215 .

Ping @ushahidi/platform

PR #878, opened 2017-10-25
2215 fix saving changes to post values

rjmackay:
- Normalize posts returned from the posts endpoint .query() method
- Revert #874

Testing checklist:
- [x] go to post-data-view
- [x] select a post with *all custom fields empty* (ie. it can have title + description but nothing else)
- [x] choose to edit, add values in the empty fields
- [x] save
- [x] values should be saved and visible in the post-detail-view

Fixes ushahidi/platform#2215

Ping @ushahidi/platform

### DIFF
--- a/app/common/services/endpoints/post-endpoint.js
+++ b/app/common/services/endpoints/post-endpoint.js
@@ -37,8 +37,17 @@ function (
                 return normalizePost(angular.fromJson(data));
             }
         },
+        save: {
+            method: 'POST',
+            transformResponse: function (data /*, header*/) {
+                return normalizePost(angular.fromJson(data));
+            }
+        },
         update: {
-            method: 'PUT'
+            method: 'PUT',
+            transformResponse: function (data /*, header*/) {
+                return normalizePost(angular.fromJson(data));
+            }
         },
         options: {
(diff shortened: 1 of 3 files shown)


rjmackay:
I think this might also resolve https://github.com/ushahidi/platform/issues/2226 but I don't know

kinstelli:
I can confirm that this is at least fixing the blank values attribute in the request on create and update.

screen shot 2017-10-24 at 8 40 08 pm

rjmackay:
> One small issue is that when you save, the post detail includes ALL attributes, even blank ones. But when you reload or select another post and then re-select the post, the blank attributes are not there. I think this is fine to leave for now though.

Ugh thats leaky and horrible though. I'll see if I can fix it.

app/common/services/endpoints/post-endpoint.js
```diff
@@ -22,24 +22,32 @@ function (
             params: {
                 order: 'desc',
                 orderby: 'post_date'
+            },
+            transformResponse: (data) => {
```
kinstelli: I didn't know jscs would let us use arrow syntax.

PR #879, opened 2017-10-25
2208 loadposts - show loading page while switching to the posts screen 

rowasc:
- show loading page while switching to the posts screen 

Testing checklist:
- [x] Open the data view mode in mobile viewport
- [x] Click on a post
- [x] You should see the full page ". . . " jumping around. 
- [x] The post should load and the '. . . 'dissapear

+ Test for this fixed scenario

Testing checklist:
- [x] Open the data view mode in mobile viewport
- [x] Click on a post that is unstructured (does not have a survey/form)
- [x] You should see the full page ". . . " jumping around. 
- [x] The post should load and the '. . . 'dissapear

Fixes ushahidi/platform#2208 .

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/detail/post-detail.controller.js
+++ b/app/main/posts/detail/post-detail.controller.js
@@ -71,6 +71,8 @@ function (
             $scope.$emit('setPageTitle', title);
         });
     }
+    angular.element(document.getElementById('bootstrap-app')).removeClass('hidden');
+    angular.element(document.getElementById('bootstrap-loading')).addClass('hidden');
 
     // Load the post form
     if ($scope.post.form && $scope.post.form.id) {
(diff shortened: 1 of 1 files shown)


app/main/posts/detail/post-detail.controller.js
```diff
@@ -71,6 +71,8 @@ function (
             $scope.$emit('setPageTitle', title);
         });
     }
+    angular.element(document.getElementById('bootstrap-app')).removeClass('hidden');
```
rowasc: besides the obvious error of only removing when we had a form id, I was waiting until fully loaded which isn't all that good , we need to be able to give the user feedback after we take them there. Also thinking this would be better handled as a general solution for when any route changes but I wasn't sure so I didn't want to go into that NOW .

PR #883, opened 2017-10-25
Update data view when posts no longer match search/filters

crcommons:
- Checks to see if, after a post has been edited, the post still matches existing filters;
- If it does not, it removes the post from the post list
- if it does still match, it does nothing

Testing checklist:
One example (can be replaced with any search)
- [ ] POST IS REMOVED FROM LIST WHEN NO LONGER MATCHING FILTERS
  - [ ] Filter posts to only show published posts
  - [ ] Select 'edit' on any post
  - [ ] Change status from Published to Under Review; select Save
  - [ ] Post should briefly change, then disappear from data view and detail view; next card in line should be select (unless you edited the last post in the list, then the selected post will be null)

- [ ] POST REMAINS IN LIST WHEN IT STILL MATCHES FILTERS 
  - [ ] Filter posts to only show published posts
  - [ ] Select 'edit' on any post
  - [ ] Make a change in the post, but leave the status set to Published; select Save
  - [ ] Postcard and post detail should update with change, but should remain in view

- [ ] Try this again using any and all filters!

Fixes ushahidi/platform#2227

Ping @ushahidi/platform

### DIFF
--- a/app/common/services/endpoints/post-endpoint.js
+++ b/app/common/services/endpoints/post-endpoint.js
@@ -22,24 +22,59 @@ function (
             params: {
                 order: 'desc',
                 orderby: 'post_date'
-            }
+            },
+            transformResponse: [
+                $http.defaults.transformResponse[0],
+                (data) => {
+                    if (!data) {
+                        return data;
+                    }
+                    if (data.results) {
+                        data.results = data.results.map(normalizePost);
+                    }
+
+                    return data;
+                }
+            ]
         },
         get: {
             method: 'GET',
-            transformResponse: function (data /*, header*/) {
-                data = angular.fromJson(data);
-                // Ensure values is always an object
-                if (_.isArray(data.values)) {
-                    data.values = _.object(data.values);
+            transformResponse: [
+                $http.defaults.transformResponse[0],
+                function (data /*, header*/) {
+                    if (!data) {
+                        return data;
+                    }
+
+                    return normalizePost(data);
                 }
-                if (!_.isArray(data.published_to)) {
-                    data.published_to = [];
+            ]
+        },
+        save: {
+            method: 'POST',
+            transformResponse: [
+                $http.defaults.transformResponse[0],
+                function (data /*, header*/) {
+                    if (!data) {
+                        return data;
+                    }
+
+                    return normalizePost(data);
                 }
-                return data;
-            }
+            ]
         },
         update: {
-            method: 'PUT'
+            method: 'PUT',
+            transformResponse: [
+                $http.defaults.transformResponse[0],
+                function (data /*, header*/) {
+                    if (!data) {
+                        return data;
+                    }
+
+                    return normalizePost(data);
+                }
+            ]
         },
         options: {
             method: 'OPTIONS'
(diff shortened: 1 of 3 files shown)


app/main/posts/views/post-view-data.directive.js
```diff
@@ -88,6 +88,41 @@ function PostViewDataController(
...
+            let query = PostFilters.getQueryParams($scope.filters);
+
+            let postQuery = _.extend({}, query, {
+                limit: $scope.itemsPerPage
+            });
+
+            postQuery.offset = ($scope.currentPage - 1) * $scope.itemsPerPage;
+
+            PostEndpoint.query(postQuery).$promise.then(function (postsResponse) {
+                for (let i = 0; i < postsResponse.results.length; i++) {
```
rjmackay: Why not just use foreach?

app/main/posts/views/post-view-data.directive.js
```diff
@@ -88,6 +88,41 @@ function PostViewDataController(
...
+                    if (postsResponse.results[i].id === $scope.selectedPostId) {
+                        return;
+                    }
+                }
+                removePostFromList();
```
rjmackay: I don't understand what this is doing? it looks like you're *always* removing it from the list?

app/main/posts/views/post-view-data.directive.js
```diff
@@ -88,6 +88,41 @@ function PostViewDataController(
...
+
+            let query = PostFilters.getQueryParams($scope.filters);
+
+            let postQuery = _.extend({}, query, {
+                limit: $scope.itemsPerPage
+            });
+
+            postQuery.offset = ($scope.currentPage - 1) * $scope.itemsPerPage;
+
+            PostEndpoint.query(postQuery).$promise.then(function (postsResponse) {
```
rjmackay: ... I'm missing something. Why are we reloading?

app/common/services/endpoints/post-endpoint.js
```diff
@@ -22,58 +22,24 @@ function (
...
-                    if (!data) {
-                        return data;
-                    }
-
-                    data.results = data.results.map(normalizePost);
-
-                    return data;
-                }
-            ]
+            }
```
rjmackay: Accidentally reverted changes??

app/main/posts/views/post-view-data.directive.js
```diff
@@ -88,6 +88,36 @@ function PostViewDataController(
      */
     var newPostsAfter = moment().utc();
     $scope.savingPost = {saving: false};
+
+    $rootScope.$on('event:edit:post:data:mode:saveSuccess', function () {
+        function removePostFromList() {
+            for (var i = 0; i < $scope.posts.length; i++) {
```
rjmackay: I'd generally suggest just using underscore for this ie. 
```
let index = _.findIndex($scope.posts, { id: selectedId });
$scope.posts.splice(i, 1);
```
Unless we see a real slowdown from doing this, I favour readability

app/common/services/endpoints/post-endpoint.js
crcommons: that was on purpose. When I wasn't getting the correct thing back from the DB, I made some unnecessary changes.

PR #886, opened 2017-10-25
2246 saved search filters

rowasc:
Fixes  :
- https://github.com/ushahidi/platform/issues/2239
- https://github.com/ushahidi/platform/issues/2246
- https://github.com/ushahidi/platform/issues/2211

### Checklist: 

- [x] In the sort/filter drop-down, select a saved search. 
- [x] You should see the filters that the saved search already had (from when it was created). 
- [x] You should NOT see the saved search bug icon as a filter. 
- [x] Change a filter, you should see the filters update in the drop-down bug icons
- [x] Apply filters, you should see the results update according to your filters. 
- [x] If your user has permissions to edit a saved search, then you can also see a button that says "Update Saved Search". 
- [x] Clicking on the update button will open a modal with the saved search information where you can change the saved search name/description and save it.
- [x] Save the saved search. This should add the filters you assigned to that saved search. 
- [x] Check that you are in /savedsearches/:id/:view after creating the saved search . You should see all your saved search filters in the dropdown, but not the saved search itself .
- [x] Hit "Clear filters." You should be taken to the view you were at (map, data)
- [x] After having cleared the filters/saved search, add a new filter. Check that we are not using the filters from the saved search. Apply filters and check that the result is OK.
- [x] Create a saved search with the filters you selected. You should be able to  create a saved search and be taken to the saved search mode view. 

### Testing that surveys and categories can be used as filters: 
- [x] I should be able to search/filter only by a sub-category
- [x] I should be able to filter by surveys. 
- [x] Switching the view from data to map or otherwise should not clear my filters

### Testing that everything works for the map
- [ ] Repeat the same steps on the OTHER mode (either map or data -- it should work the same way on both)

Fixes ushahidi/platform#2246 .
Fixes ushahidi/platform#2239 .
Fixes ushahidi/platform#2211 .

Ping @ushahidi/platform


crcommons:
- [ ] Update saved search should only show when the search has more filters than the original saved search filters
- [ ] When you select update, the model should say update or edit, rather than Create Saved Search
- [ ] There is no longer a blank option for saved search filter after you've used the saved search filter once, which means you can't select the saved search if it gets stuck https://d.pr/i/nq2Mp5
- [ ] trying to get out of a saved search is still a mess. The url remains even when you clear filters

rowasc:
Update saved search should only show when the search has more filters than the original saved search filters
.-----
Clarification : discussed this with Carolyn and should be when you have different filters than the original saved search filters.

rowasc:
@crcommons I think at this point we can't add the diff without some more refactoring, but we are cutting it close as it is in time. The problem is that basically I can compare against filters but the savedsearch with the ngmodel is tied to them, so it's not going to really be different. 
Maybe we can discuss this later today, I think it's worth putting that in another ticket with a P1 since I don't want to poke at this too much so close to freeze time (except for bugs that can block users of course)

rowasc:
My tests fail :(

crcommons:
I agree. We can hold off on the Update issue, because worst case scenario is that they update their saved search and it's the exact same saved search as before.

app/main/posts/views/filters/filter-saved-search.directive.js
```diff
@@ -23,22 +23,17 @@ function FilterSavedSearch(SavedSearchEndpoint, _,  $rootScope, ModalService) {
...
-                selectedSavedSearch.id.toString() : ngModel.$viewValue);
-        }
-
-        function renderModelValue() {
-            scope.selectedSavedSearch = ngModel.$viewValue;
-            // Update savedSearch w/o breaking references used by selectedSavedSearch model
+            if (selectedSavedSearch && selectedSavedSearch.hasOwnProperty('filter')) {
+                PostFilters.setFilters(selectedSavedSearch.filter);
+            }
+            ngModel.$setViewValue(selectedSavedSearch);
```
rowasc: only doing this for the transformers which render the values somewhere else. But I  started thinking, I don't think I actually need it. Don't want to mess too much with things today TBH but I'm not really sure about this.

test/unit/main/post/savedsearches/savedsearch-controller-spec.js
```diff
@@ -32,7 +32,9 @@ describe('set savedsearches controller', function () {
         $controller('savedsearchesController', {
             $scope: $scope,
             savedSearch: {
-                allowed_privileges: 'update'
+                allowed_privileges: 'update',
```
rowasc: ok that was silly of me to not realize . Thanks Will

crcommons:
As a regular user, I can't filter by a saved search. The saved search dropdown isn't showing, even though I set the permissions to everyone.

rowasc:
@crcommons ok that actually makes sense if you don't have your own saved searches. You have to make it featured for it to appear to all users.  If you set permissions to everyone all you say to it is "everyone can access it from the url" .

crcommons:
when a regular user creates a saved search that is visible to everyone, everyone cannot see it. only that user.

crcommons:
if you click update search and then select cancel, the "update search" button disappears as an option in the dropdown.

rowasc:
@crcommons nice catch. I will see what I can do :)

crcommons:
The SS dropdown in the filters dropdown doesn't always revert back to 'select one' when you clear filters. (specifically when the page doesn't re-route to a different view... i.e. the SS has a default view of Data and when I clear filters, the SS remains selected in the dropdown)

app/main/posts/views/filters/filters-dropdown.directive.js
```diff
@@ -39,8 +48,13 @@ function FiltersDropdown(PostFilters, ModalService, $rootScope) {
             // @TODO Prevent the user from creating one if they somehow manage to get to this point without being logged in
             $scope.savedSearch.user_id = $rootScope.currentUser ? $rootScope.currentUser.userId : null;
             ModalService.openTemplate('<saved-search-editor saved-search="savedSearch"></saved-search-editor>', 'set.create_savedsearch', 'star', $scope, false, false);
-
-            //ModalService.openTemplate('<savedsearch-create></savedsearch-create>', 'nav.saved_searches', '/img/iconic-sprite.svg#star', $scope.$parent, true, true);
+        };
+        $scope.editSavedSearchModal = function () {
+            $scope.savedSearch = $scope.filtersVar.saved_search;
+            $scope.savedSearch.filter = PostFilters.getActiveFilters($scope.filtersVar);
+            // @TODO Prevent the user from creating one if they somehow manage to get to this point without being logged in
```
crcommons: At this point, $scope.filtersVar.saved_search returns as an empty object in the $watch function, yet it still is populated correctly if you check the value within the editSavedSearchModal. If there is any asynchronous code, that could be causing the $watch method to mess up.

app/main/posts/views/filters/filters-dropdown.directive.js
```diff
@@ -39,8 +49,15 @@ function FiltersDropdown(PostFilters, ModalService, $rootScope) {
             // @TODO Prevent the user from creating one if they somehow manage to get to this point without being logged in
             $scope.savedSearch.user_id = $rootScope.currentUser ? $rootScope.currentUser.userId : null;
             ModalService.openTemplate('<saved-search-editor saved-search="savedSearch"></saved-search-editor>', 'set.create_savedsearch', 'star', $scope, false, false);
-
-            //ModalService.openTemplate('<savedsearch-create></savedsearch-create>', 'nav.saved_searches', '/img/iconic-sprite.svg#star', $scope.$parent, true, true);
+        };
+        $scope.editSavedSearchModal = function () {
+            SavedSearchEndpoint.get({id: PostFilters.getModeId()}, function (savedSearch) {
```
rowasc: Doing this so we don't directly depend on the saved search object being set in filters and use the modeId  to get the full object from the backend

app/main/posts/views/filters/filters-dropdown.directive.js
```diff
@@ -39,8 +49,15 @@ function FiltersDropdown(PostFilters, ModalService, $rootScope) {
...
+                $scope.savedSearch = savedSearch;
+                $scope.savedSearch.filter = PostFilters.getActiveFilters($scope.filtersVar);
```
rowasc: this sets the current active filters in the saved search so we can edit them

app/main/posts/savedsearches/savedsearches-controller.js
```diff
@@ -43,7 +43,10 @@ module.exports = [
 
         // Set initial filter state
         PostFilters.setMode('savedsearch', savedSearch.id);
-        PostFilters.setFilters(savedSearch.filter);
+
+        var filters = savedSearch.filter;
+        filters.saved_search = savedSearch.id;
```
rowasc: I think this can be removed. It's not doing anything there because we don't use the saved search in the filter query, and might be breaking something since it should be an object, unless we can just go back to how it was (with IDs) which would be best i think..

app/main/posts/views/filters/filter-saved-search.html
```diff
@@ -8,6 +8,7 @@
                         search in searches
                         track by search.id"
         >
+            <option value="" translate="app.select_one">Select one</option>
```
rowasc: I just checked and this does not revert back the search after being set since it's not bound to anything.

app/main/posts/views/filters/filters-dropdown.directive.js
```diff
@@ -16,20 +16,30 @@ function FiltersDropdown(PostFilters, ModalService, $rootScope) {
...
             role : []
         };
+
+        // Check if we can edit
+        $scope.canEdit = function (savedSearch) {
+            return _.contains(savedSearch.allowed_privileges, 'update');
+        };
         $scope.applyFiltersLocked = function () {
-            PostFilters.reactiveFilters = 'enabled';
+            PostFilters.reactiveFilters = true;
```
rowasc: this is just some sane refactoring along the way because for some reason I had written enabled instead of actually using a boolean 👎

app/main/posts/views/filters/filters-dropdown.directive.js
```diff
@@ -16,20 +16,30 @@ function FiltersDropdown(PostFilters, ModalService, $rootScope) {
...
             $scope.dropdownStatus.isopen = !$scope.dropdownStatus.isopen;
         };
         $scope.clearFilters = function () {
+            if (PostFilters.getMode() === 'savedsearch' && PostFilters.getModeId()) {
```
rowasc: when we clear filters we want to just go to the normal mode and redirect them as described in PR

app/main/posts/views/post-filters.service.js
```diff
@@ -103,6 +103,15 @@ function PostFiltersService(_, FormEndpoint, TagEndpoint, $q) {
         var query = _.omit(
             filters,
             function (value, key, object) {
+                if (key === 'saved_search') {
```
rowasc: just omitting the saved_search, reactiveFilters and qEnabled since we don't search for those in the Back

Angamanga:
Things I found while testing: 
1. Update-button does not appear until you click 'apply filters' and then open the dropdown again.
2. In data-view, when I have selected a saved search and then click 'clear filters' I get redirected to 'views/undefined' = the map without the mode-context-bar

app/main/posts/views/filters/filters-dropdown.directive.js
```diff
@@ -16,20 +16,30 @@ function FiltersDropdown(PostFilters, ModalService, $rootScope) {
...
+                PostFilters.setMode('all');
+                $location.url('/views/' + $scope.filtersVar.saved_search.view);
```
Angamanga: This is where you get redirected to views/undefined since $scope.filtersVar.saved_search.view is undefined.

Angamanga:
@rowasc Your comments above is soo helpful when reviewing! 👌

app/main/posts/views/filters/filters-dropdown.directive.js
rowasc: UGH. Thanks, will fix

rowasc:
I think this is OK. @jshorland when you have some time can you take a look? It should be fine now

crcommons:
@rowasc 

When I update a saved search, the bugs in the top of the dropdown don't match the saved search (droplr, it's long, but watch the whole thing: https://d.pr/i/FbG6eX)

I also think that sometimes clearing the filters does not de-select the saved search from the select-dropdown. But I need to figure out how to reproduce this reliably

rowasc:
uh super weird about the filters. I'll check it out, thanks @crcommons

rowasc:
the filters thing is something that happens when you switch page too. :/ weirdly enough only with the categories . Probably not related to this feature but I' digging :/

rowasc:
Confirmed this seems to be an issue with categories not showing up :/

rowasc:
Also this is not directly related to saved searches. Category filters seems to be just wrong , in general. Fixing now

crcommons:
@rowasc The Update Button no longer appears, so I can't test updating a saved search.

rowasc:
@crcommons I might have broken it when we were testing =( it was being shown before. I'm chasing a weird category thing but I'll look into this now too. Thanks

app/main/posts/views/filters/filter-saved-search.directive.js
```diff
@@ -17,28 +17,34 @@ function FilterSavedSearch(SavedSearchEndpoint, _,  $rootScope, ModalService) {
         scope.searchesLength = 0;
         scope.loading = false;
         scope.$on('savedSearch:update', loadSavedSearches);
-
+        scope.$watch(PostFilters.getModeId, function (newValue, oldValue) {
```
rowasc: We need to be aware of the modeid changes because that is used across the app to select saved searches and reset them .

app/main/posts/views/filters/filter-saved-search.directive.js
```diff
@@ -17,28 +17,34 @@ function FilterSavedSearch(SavedSearchEndpoint, _,  $rootScope, ModalService) {
...
             ModalService.openTemplate('<saved-search-list-editor-modal searches="searches"></saved-search-list-editor-modal>', 'set.delete_saved_searches', 'star', scope, false, false);
         };
 
         function activate() {
-            ngModel.$render = renderModelValue;
-            if (ngModel.$viewValue.length > 0) {
+            if (ngModel.$viewValue) {
                 scope.selectedSavedSearch = scope.searches[ngModel.$viewValue];
+            } else if (PostFilters.getModeId()) {
+                scope.selectedSavedSearch = scope.searches[PostFilters.getModeId()];
```
rowasc: this here attempts to solve the first load issue on saved searches. basically you should be able to get the modeid AFTER loading the .searches array which is what happens here in activate

app/main/posts/views/filters/filter-saved-search.directive.js
```diff
@@ -17,28 +17,34 @@ function FilterSavedSearch(SavedSearchEndpoint, _,  $rootScope, ModalService) {
...
-        }
-
-        function renderModelValue() {
-            scope.selectedSavedSearch = ngModel.$viewValue;
-            // Update savedSearch w/o breaking references used by selectedSavedSearch model
+            if (selectedSavedSearch && selectedSavedSearch.hasOwnProperty('filter')) {
+                PostFilters.setFilters(selectedSavedSearch.filter);
+                PostFilters.setMode('savedsearch', selectedSavedSearch.id);
+            } else if (selectedSavedSearch === null && PostFilters.getModeId() !== null) {
+                PostFilters.setMode('all');
```
rowasc: this clears the saved search filter by brute force.

app/main/posts/views/filters/filters-dropdown.html
```diff
@@ -17,7 +17,11 @@
         <post-active-search-filters ng-model="filtersVar" filters-var="filtersVar"></post-active-search-filters>
     </div>
     <div class="form-field">
-        <button type="button" class="button button-alpha button-right" ng-click="saveSavedSearchModal()">Save search</button>
+        <button type="button" class="button button-alpha button-right" ng-click="saveSavedSearchModal()" translate="set.save_savedsearch">Save search</button>
+    </div>
+
+    <div class="form-field">
+        <button type="button" class="button button-alpha button-right" ng-click="editSavedSearchModal()" ng-show="canUpdateSavedSearch" translate="set.update_savedsearch">Edit Saved search</button>
```
rowasc: this is shown only when you can update the saved search, meaning you have selected one and your permissions are OK

app/main/posts/views/main.html
```diff
@@ -16,8 +16,6 @@ <h1 class="mode-context-title"><a href="/" ng-bind="nav.site.name"></a></h1>
             <div class="mode-context-body">
                 <p markdown-to-html="nav.site.description"></p>
```
rowasc: this was legacy stuff that survived the initial merges but should have gone away much earlier. Basically replaced by the active search filters directive

app/main/posts/views/post-filters.service.js
```diff
@@ -100,14 +102,32 @@ function PostFiltersService(_, FormEndpoint, TagEndpoint, $q) {
...
+                if (key === 'qEnabled') {
+                    return true;
+                }
                 // Is value empty?
                 // Is it a date?
                 if (_.isDate(value)) {
                     return false;
                 }
+
+                // Is an array with all the same elements? (order doesn't matter)
```
rowasc: we don't want to send category or surveys when we have the same as the default values because that might cause backend issues specially around things that dont have categories

test/unit/main/post/views/filters/active-filters.directive.spec.js
```diff
@@ -1,81 +0,0 @@
-describe('post active filters directive', function () {
```
rowasc: this test is part of the things that are gone with the refactor because we don't have the directive anymore

app/main/posts/views/filters/active-filters.directive.js
```diff
@@ -1,63 +0,0 @@
-module.exports = ActiveFilters;
```
rowasc: old directive that was essentially dupe of the new one .

PR #889, opened 2017-10-26
filtering out posts when changing status from post-actions #2227

Angamanga:
- removes the post from filters if status is changed in the '...' menu

Testing checklist:
... button checklist
- [x] Go to data-view
- [x] set a status-filter (like 'under review')
- [x] update to a status that not matches the filters for post through the '...' -menu
- [x] post should go away and the next one selected
- [x] Try this for several posts and the '...' menu in the detailed view
- [x] confirm the opposite: change the status of a post to match current filters (e.g. set filters to 'under review and archived', change a post to from under review to archived and confirm it remains)

Bulk Actions checklist
- [x] run same checklist but edit status through bulk actions

Detail Edit View Checklist
- [x] run same checklist but through data view detail editor

No filters
- [ ] if there are no status filters set, change a post status to archived using each method above and confirm that it is removed. 
- [ ] then change a status from 'under review' to 'published' and confirm it remains

Fixes ushahidi/platform#2227 .

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/common/post-actions.directive.js
+++ b/app/main/posts/common/post-actions.directive.js
@@ -89,7 +89,7 @@ function PostActionsDirective(
 
             PostEndpoint.update($scope.post).$promise.then(function () {
                 // adding post to broadcast to make sure it gets filtered out from post-list if it does not match the filters.
-                $rootScope.$broadcast('event:edit:post:data:mode:saveSuccess', {post: $scope.post});
+                $rootScope.$broadcast('event:edit:post:status:data:mode:saveSuccess', {post: $scope.post});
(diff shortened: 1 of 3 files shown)


jshorland:
If I select all 20 posts without more being in view, and change their filters so they don't match my search anymore, it returns "0 of X posts, your search didn't match any posts" instead of loading the next 20. 

see gif: 
screen capture on 2017-10-26 at 11-55-07

jshorland:
fixed. AMAZING. passes QA.

Angamanga:
Works fine in Sandbox!

PR #890, opened 2017-10-26
Switching localStorage in favor of sessionstorage  for the caching issues

rowasc:
-Switching localStorage in favor of sessionstorage  for the caching issues

Testing checklist:
- [ ] clear your localstorage and session storage. 
- [ ] access the app
- [ ] verify that you have the caches now living in sessionStorage instead of localStorage 

Check that the app still works, that you can access forms, edit posts, access your settings ,etc
Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/common/configs/cache-config.js
+++ b/app/common/configs/cache-config.js
@@ -2,7 +2,7 @@ module.exports = ['CacheFactoryProvider', function (CacheFactoryProvider) {
     angular.extend(CacheFactoryProvider.defaults, {
         maxAge: 15 * 60 * 1000, // 15 mins
         cacheFlushInterval: 60 * 60 * 1000, // This cache will clear itself every hour.
-        storageMode: 'localStorage',
+        storageMode: 'sessionStorage',
         storagePrefix: 'ush-caches.',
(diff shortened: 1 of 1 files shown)


app/common/configs/cache-config.js
```diff
@@ -2,7 +2,7 @@ module.exports = ['CacheFactoryProvider', function (CacheFactoryProvider) {
     angular.extend(CacheFactoryProvider.defaults, {
         maxAge: 15 * 60 * 1000, // 15 mins
         cacheFlushInterval: 60 * 60 * 1000, // This cache will clear itself every hour.
-        storageMode: 'localStorage',
+        storageMode: 'sessionStorage',
         storagePrefix: 'ush-caches.',
```
tuxpiper: btw, I'm just realising this prefix and storage approach probably dooms the installation of two deployments under the same domain? i.e. mydomain.com/deployment1 , mydomain.com/deployment2

not a critique or something we have to support, just a realisation moment

rjmackay:
We might actually want to go as far as setting `storageMode: memory` if sessionStorage is still weird.

PR #891, opened 2017-10-26
Change to getFresh in post edit and datasource configuration

rowasc:
- Change to getFresh in post edit and datasource configuration

Testing checklist:
- [ ] in the network panel, check that datasources is going to get the config data to the backend (settings->datasources)
- [ ] in the network panel, check that post edit is going to get the attributes to the backend (select a post, edit) 
Fixes ushahidi/platform#.

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/modify/post-edit.controller.js
+++ b/app/main/posts/modify/post-edit.controller.js
@@ -36,7 +36,7 @@ function (
         });
 
         if (post.form) {
-            $scope.form = FormEndpoint.get({ id: post.form.id }, function (form) {
+            $scope.form = FormEndpoint.getFresh({ id: post.form.id }, function (form) {
                 // Set page title to post title, if there is one available.
                 if (post.title && post.title.length) {
                     $translate('post.modify.edit_type', { type: form.name, title: post.title }).then(function (title) {
--- a/app/settings/datasources/datasources.controller.js
+++ b/app/settings/datasources/datasources.controller.js
@@ -88,7 +88,7 @@ function (
         }
     };
 
-    $q.all([DataProviderEndpoint.queryFresh().$promise, ConfigEndpoint.get({ id: 'data-provider' }).$promise]).then(function (response) {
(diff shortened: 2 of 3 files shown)


app/main/posts/modify/post-edit.controller.js
```diff
@@ -36,7 +36,7 @@ function (
         });
 
         if (post.form) {
-            $scope.form = FormEndpoint.get({ id: post.form.id }, function (form) {
+            $scope.form = FormEndpoint.getFresh({ id: post.form.id }, function (form) {
```
willdoran: This won't use the cache at all, which might be fine but it will probably dramatically increase requests.

app/settings/datasources/datasources.controller.js
```diff
@@ -88,7 +88,7 @@ function (
         }
     };
 
-    $q.all([DataProviderEndpoint.queryFresh().$promise, ConfigEndpoint.get({ id: 'data-provider' }).$promise]).then(function (response) {
+    $q.all([DataProviderEndpoint.queryFresh().$promise, ConfigEndpoint.getFresh({ id: 'data-provider' }).$promise]).then(function (response) {
```
willdoran: Same as above, but less dramatically since we should really not cache data providers.

app/common/configs/cache-config.js
```diff
@@ -2,7 +2,7 @@ module.exports = ['CacheFactoryProvider', function (CacheFactoryProvider) {
     angular.extend(CacheFactoryProvider.defaults, {
         maxAge: 15 * 60 * 1000, // 15 mins
         cacheFlushInterval: 60 * 60 * 1000, // This cache will clear itself every hour.
-        storageMode: 'localStorage',
+        storageMode: 'sessionStorage',
```
willdoran: The cache lib is pretty old now and not supported and it had this issue from a few days ago, which ain't great
https://github.com/jmdobry/angular-cache/issues/268

tuxpiper:
merging for testing in uchaguzi sandbox

PR #892, opened 2017-10-26
2257 Adding back map by latitude and longitude

crcommons:
- Adds the old code that allows a user to search by latitude and longitude in input fields, rather than in the map search field. This ensures the map marker placement is _exact_.

Testing checklist:
- [x] Add a new post; add location coordinates using the inputs and select Update Map; Map should place marker at location from coordinates
- [x] Save post; confirm location persists in map
- [x] do the same when editing a post

Fixes ushahidi/platform#2257

Ping @ushahidi/platform

### DIFF
--- a/app/common/locales/en.json
+++ b/app/common/locales/en.json
@@ -1208,8 +1208,7 @@
         "click_map": "<em>You can search or click the area of map where you want to place the marker.</em>",
         "placeholder":"Search for addresses, place names, or coordinates...",
         "search_results": "Search results",
-        "error":"Sorry we could not find a location matching your search",
-        "coordinates_tooltip": "Example Coordinates"
+        "error":"Sorry we could not find a location matching your search"
     },
     "notification": {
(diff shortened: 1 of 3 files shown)


crcommons:
We didn't, but that's not a bad idea. I'm going to add an issue because there is some other refactoring that needs to happen. I'll include that.

Angamanga:
Works fine in sandbox! Passes qa

kkoxenberners:
Sorry - not working at our user level still :(

lat long box not working - machakos

The prompted location (C99, Machakos) is actually a location on road south of Machakos - 5 km south.
OK, could be worse - but when we have info about something happening in a settlement, we want to be able to be within 100s of metres at worst...not km.
If this were in Nairobi or Kisumu, it would mean we would be geolocating in totally the wrong place.

The issue requested that lat, long entry dictate where map marker is placed - that still hasn't happened...at least at our end.

PR #894, opened 2017-10-26
Fix initialization of cache, and clear expired items on cache load

rjmackay:
- We don't need to `new` the cache. Use createCache() instead
- Remove expired items from cache during initialization
- Don't skip the cache to load forms, tags or users

Testing checklist:

- [x] Browser 1: Load map view http://deployment.com/views/map
- [x] Browser 1: Close the window as soon as the map has finished loading (make sure no other windows are open for that deployment)
- [x] Browser 2: Edit a survey and change the title. Save the survey
... wait 5mins (so cache should expire)
- [x] Browser 1: Load the post again  ie http://deployment.com/views/map
... Hopefully you see up to date survey names

Fixes ushahidi/platform#2088

Ping @ushahidi/platform

### DIFF
--- a/app/common/configs/cache-config.js
+++ b/app/common/configs/cache-config.js
@@ -1,7 +1,7 @@
 module.exports = ['CacheFactoryProvider', function (CacheFactoryProvider) {
     angular.extend(CacheFactoryProvider.defaults, {
         maxAge: 15 * 60 * 1000, // 15 mins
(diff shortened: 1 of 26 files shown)


rjmackay:
Test url: http://qa.2-88-stale-cache-on-load-fixes.ushahidilab.com/

jshorland:
I had to wait 15 min, but it changed appropriately when i reopened it in browser 1

PR #900, opened 2017-10-30
unselecting post-card in data-view #2238

Angamanga:
- makes it possible to unselect a post in the data-view

Testing checklist:
- [x] Select a post in data-view
- [x] Click on it again
- [x] It should move back into original position and no post should be shown to the right

1.  Select a post again
2. Click edit
3. Make some changes
4. Click on the same post again
5. The unsaved-changes modal appears

- [x] Click continue without saving
  - [x] Postcard should be unselected, move back into original position and no post should be shown to the right
- [x] Do 1-5 above
  - [x] Click 'save'
  - [x] Changes should be saved and
  - [x] Postcard should be unselected, move back into original position and no post should be shown to the right
- [x] Do 1-5 above
  - [x] Click on the x on the modal
  - [x] You should get back to the edit-mode.

Fixes ushahidi/platform#2238 .

Ping @ushahidi/platform


rowasc:
Also tested mobile view opening, which it is. (Y)

PR #902, opened 2017-10-31
Adds ux to the update-saved-search

Angamanga:
- Disables the 'update search' button in the saved-search-mode-context-bar while saving
- Notifies the user when the update is complete

Testing checklist:
- [x] Select a saved search
- [x] Go to map-view
- [x] Add a filter
- [x] Click on the 'update search' button in the mode-context-bar
- [x] It should be greyed out and text changed to 'updating' while saving
- [x] When updating is complete, a notification should appear saying 'saved-search-name was updated'

Fixes ushahidi/platform#2263 .

Ping @ushahidi/platform


rowasc:
looks nice. testing now

rowasc:
I think we should do this also in the other saved search button.

PR #903, opened 2017-11-02
ResetDefaults function added as a workaround for filters not being cleared in login and logout

rowasc:
- ResetDefaults function added as a workaround for filters not being cleared in login and logout
- Fixes the bug that caused that `some` categories (public ones) were being shown as selected after login but not all of them (when you actually didn't have any selected filters)

Testing checklist:
### Categories and filters should be cleared in login and logout: 

- [x] Make sure you have at least 1 category that is set for "everyone" to see.  
- [x] Login
- [x] Look at the drop-down filters panel, make sure ALL categories are shown as selected in the categories checkboxes (they should because you have no filters) 
- [x] Logout.
- [x] Check your filters. You should only see the categories that you have set for "Everyone" to see in your categories checkboxes, and you should have no filters, in general. 
- [x] Login AGAIN
- [x] Look at the drop-down filters panel, make sure ALL categories are shown as selected in the categories checkboxes (they should because you have no filters) 
- [x] Uncheck one or more categories, select other filters. Make sure you see the bug icons appear and that they represent your current category filters. 
- [x] Make sure you Apply filters  (if you are in the data mode). Results should be consistent with the filters selected. 
- [x] Clear filters, and check that they are in fact clear. 

Fixes ushahidi/platform#2239.

Ping @ushahidi/platform

### DIFF
--- a/app/common/auth/authentication-events.run.js
+++ b/app/common/auth/authentication-events.run.js
@@ -1,7 +1,7 @@
 module.exports = AuthenticationEvents;
 
-AuthenticationEvents.$inject = ['$rootScope', '$location', 'Authentication', 'Session', '_', '$route', 'TermsOfService', 'Notify'];
-function AuthenticationEvents($rootScope, $location, Authentication, Session, _, $route, TermsOfService, Notify) {
+AuthenticationEvents.$inject = ['$rootScope', '$location', 'Authentication', 'Session', '_', '$route', 'TermsOfService', 'Notify', 'PostFilters'];
+function AuthenticationEvents($rootScope, $location, Authentication, Session, _, $route, TermsOfService, Notify, PostFilters) {
     let loginPath = null;
     $rootScope.currentUser = null;
     $rootScope.loggedin = false;
@@ -32,6 +32,7 @@ function AuthenticationEvents($rootScope, $location, Authentication, Session, _,
                  * references https://github.com/ushahidi/platform/issues/1714
                  */
                 adminUserSetup();
+                PostFilters.resetDefaults();
                 if (redirect) {
                     $location.url(redirect);
                 }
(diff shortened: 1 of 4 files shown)


app/common/auth/authentication-events.run.js
```diff
@@ -42,10 +43,14 @@ function AuthenticationEvents($rootScope, $location, Authentication, Session, _,
     function doLogout(redirect) {
         $rootScope.currentUser = null;
         $rootScope.loggedin = false;
-        if (redirect) {
-            $location.url(redirect);
-        }
-        $route.reload();
+        // we don' wat to reload until after filters are correctly set with the backend default that the user
+        // would get when logged out
+        PostFilters.resetDefaults().then(function () {
```
rowasc: the main reason behind this is to make sure that we don't reload before we reset, which causes some UI weirdness.

app/main/posts/views/post-filters.service.js
```diff
@@ -30,6 +31,14 @@ function PostFiltersService(_, FormEndpoint, TagEndpoint, $q) {
         qEnabled: false
     };
 
+    /**
+     * function to deal with the fact that logout and login don't really reset the defaults.
+     */
+    function resetDefaults() {
+        return activate().then(function () {
```
rowasc: just run activate to reset all the tags and such, and then run clearfilters which uses the defaults previously set

rowasc:
@crcommons I think this solves the main problem we saw today with the default filters. It does not necessarily solve any category related issues that we already had, but I want to address that in a separate PR because I really don't think it's even part of something 2239 broke, it's mainly that since sub-categories were not using children, so many bugs were masked :(.  and we can probably create an issue for the other bug if we can actually make it happen. I still cannot reproduce it for some reason.

app/main/posts/views/post-filters.service.js
rowasc: @crcommons some context on why this is async and why this is not a broadcast

test/unit/common/auth/authentication-events.run.spec.js
```diff
@@ -11,9 +12,16 @@ describe('global event handlers', function () {
         };
 
     beforeEach(function () {
-
         var testApp = makeTestApp();
-
+        mockedPostFiltersService = {
```
rowasc: @crcommons  just a silly test mock because otherwise the auth events failed due to not having the resetDefaults & PostFilters available =)

PR #906, opened 2017-11-02
TOS Session Hotfix

crcommons:
- hotfix that adds a grantType to the session so we know if we have a client or user access token; should make it so that do login/tos doesn't run by clearing local storage if grantType isn't password (user).

Testing checklist:
- [ ] 

Fixes ushahidi/platform#2220

Ping @ushahidi/platform

### DIFF
--- a/app/common/auth/authentication-interceptor.config.js
+++ b/app/common/auth/authentication-interceptor.config.js
@@ -33,7 +33,7 @@ function AuthInterceptor($rootScope, $injector, $q, CONST, Session, _) {
             // Save token expires time
             Session.setSessionDataEntry('accessTokenExpires', authResponse.data.expires);
 
-            Session.setSessionDataEntry('grantType', 'client');
+            Session.setSessionDataEntry('grantType', 'client_credentials');
             // Add Authorization header
             config.headers.Authorization = 'Bearer ' + accessToken;
             deferred.resolve(config);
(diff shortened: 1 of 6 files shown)


app/common/auth/authentication-interceptor.config.js
```diff
@@ -32,6 +32,8 @@ function AuthInterceptor($rootScope, $injector, $q, CONST, Session, _) {
             Session.setSessionDataEntry('accessToken', accessToken);
             // Save token expires time
             Session.setSessionDataEntry('accessTokenExpires', authResponse.data.expires);
+
+            Session.setSessionDataEntry('grantType', 'client');
```
rjmackay: ideally this might be `client_credentials` to match the actual grant type above

app/common/auth/authentication.service.js
```diff
@@ -22,11 +22,21 @@ function (
     ModalService
 ) {
 
-    // check whether we have initially an valid access_token and userId
-    // and assume that, if yes, we are still loggedin
-    var loginStatus = !!Session.getSessionDataEntry('accessToken') &&
-        Session.getSessionDataEntry('accessTokenExpires') > Math.floor(Date.now() / 1000) &&
-        !!Session.getSessionDataEntry('userId');
+    // check whether we have initially an valid access_token and assume that, if yes, we are still loggedin
```
rjmackay: @crcommons I restructured this a bit to make Auth service responsible for clearing auth state if the token has expired.

test/unit/common/auth/authentication.service.spec.js
```diff
@@ -44,23 +45,50 @@ describe('Authentication', function () {
...
-        Authentication = _Authentication_;
+        $injector = _$injector_;
     }));
 
-    describe('beeing still logged out', function () {
-        describe('getLoginStatus', function () {
-            it('should return false', function () {
-                expect(Authentication.getLoginStatus()).toBe(false);
-            });
+    describe('initial login status', function () {
```
rjmackay: And added some tests to check it works

PR #911, opened 2017-11-08
removing only selected value from filter-arrays

Angamanga:
- When you remove a filter by clicking on the bug-icon, it just removes that filter

Testing checklist:
- [ ] Go to the filter dropdown
- [ ] Select a series of filter, especially surveys, datasources and categories with many choices
- [ ] try remove one filter at the time from the bug-icons, only the removed filter should be removed, the other ones should stay

Fixes ushahidi/platform#2183 .

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/views/post-filters.service.js
+++ b/app/main/posts/views/post-filters.service.js
@@ -75,7 +75,13 @@ function PostFiltersService(_, FormEndpoint, TagEndpoint, $q) {
     }
 
     function clearFilter(filterKey, value) {
-        filterState[filterKey] = getDefaults()[filterKey];
+        // if filter is in an array, we only want to remove that specific value
+        if (Array.isArray(filterState[filterKey])) {
+            filterState[filterKey].splice(filterState[filterKey].indexOf(value), 1);
+        } else {
+            filterState[filterKey] = getDefaults()[filterKey];
+        }
+
         /**
          * Since q is a special type of filter that gets applied on the click
          * of a button separate from the input control, and since it should be automatically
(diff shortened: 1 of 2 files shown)


app/main/posts/views/post-filters.service.js
```diff
@@ -75,7 +75,13 @@ function PostFiltersService(_, FormEndpoint, TagEndpoint, $q) {
     }
 
     function clearFilter(filterKey, value) {
-        filterState[filterKey] = getDefaults()[filterKey];
+        // if filter is in an array, we only want to remove that specific value
```
rjmackay: So this just reverts to the previous behaviour?
Are there tests to avoid regressions in future?

Angamanga: Yes, I am not sure why I removed this before since I was not using that function, must have been temporary(?) confusion. Tests added now...

test/unit/main/post/views/filters/post-filters.service.spec.js
```diff
@@ -89,6 +89,41 @@ describe('Post Filters Service', function () {
                 order_unlocked_on_top: 'true'
             });
         });
+        it ('if filter is represented of an array, only selected filter-value should be removed', function () {
```
rowasc: @Angamanga might not be related, but we should also make sure that when all filters are removed from the array, we should set the default instead. LMK what you think. The behavior seems to be working fine so it would be missing just that test.

rowasc: other than that, it looks nice !

rowasc: new default test Looks nice. GJ!

PR #907, opened 2017-11-03
Migrate to use ui-router

rowasc:
### Known issues and TODOs 
- Test all routes again (after/while unit tests were being adjusted I changed a lot of things that I need to retest to make sure they work)
- [x] Clicking in a post in the data view results in a redirect to the post detail view (because the handler code was removed)
- [x] saved search navigation is broken (takes you out of it when switching pages), there are a few routes from posts that I need to finish migrating today like posts/:id/edit since I missed them .
- [ ] needs some tidying up ! needs to drop stateParams entirely. 

This pull request makes the following changes:
- Replaces ngrouter with uirouter
- This PR has only the "first stage" of uirouter migration, which is to have the uirouter replace ngrouter, but not the whole refactoring effort that comes afterwards to take full advantage.

Testing checklist:
- [ ] everything! 

Fixes ushahidi/platform# .

Ping @ushahidi/platform

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/907)

PR #912, opened 2017-11-09
UiRouter Collection/PostRoutes Fixes

rowasc:
- Adding collection view=data in list.data collections instead of :view param in url
- more to come
Testing checklist Collections:
- [x] Create your own collection/saved search so you know what's supposed to be in there. 
- [x] Open the collection/saved search in data mode. Check that only its items are visible. 
- [x] Open the collection/saved search in map mode.  Check that only its items are visible. 
- [x] Open the collection/saved search with no view param in the url to make sure that it loads in the correct default view 

Fixes ushahidi/platform# .

Ping @ushahidi/platform


rowasc:
Reference base branch: https://github.com/ushahidi/platform-client/pull/907/commits

rowasc:
@Angamanga can you take a look on monday? thanks! !

rowasc:
I'm not too sure but maybe this can be improved with sticky states like DSR?  :/ need to read the docs a bit more

rowasc:
thinking about this a bit more, I think I figured out a better way. Untagging Anna for now

rowasc:
@Angamanga this needs to be reviewed carefully, but also, I want to go through some of this stuff with you because I don't really understand the post stats bit all that well and I might have broken everything elsewhere :/ Anyway, when you can please take a look on Monday and when I get online if you have time we can go trough it?

app/main/posts/savedsearches/savedsearches-controller.js
```diff
@@ -42,6 +44,21 @@ module.exports = [
         var filters = savedSearch.filter;
         PostFilters.setFilters(filters);
         $scope.filters = PostFilters.getFilters();
+        /**
+         * @uirouter-refactor improve on this as we continue migrations?
```
rowasc: @Angamanga here

app/main/posts/views/mode-context-form-filter.directive.js
```diff
@@ -116,18 +116,26 @@ function ModeContextFormFilter($scope, FormEndpoint, PostEndpoint, TagEndpoint,
             include_unmapped: true
         });
 
+        // @uirouter-refactor - Ask Anna??
```
rowasc: @Angamanga here

app/main/posts/collections/collections-routes.js
```diff
@@ -0,0 +1,98 @@
+module.exports = [
+    '$stateProvider',
+    function (
+        $stateProvider
+    ) {
+
+    $stateProvider
+    .state(
+        {
+            url: '^/collections/:id',
```
rowasc: this is just a state that takes you to the right view if you don't tell it where you need to go.

app/main/posts/collections/collections-routes.js
```diff
@@ -0,0 +1,98 @@
...
+            name: 'collectionRedirector',
+            params: {
+                id: null
+            },
+            onEnter: function ($state, collection) {
+                var viewParam = collection.view;
+                if (viewParam === 'list' || viewParam === 'data') {
```
rowasc: list is just a fallback thing for users who might have saved the wrong url when we still had a list view

app/main/posts/collections/collections.html
```diff
@@ -1,8 +1,5 @@
 <div>
     <layout-class layout="{{layout}}"></layout-class>
     <collection-mode-context  ng-if="layout === 'a'"></collection-mode-context>
-
-    <main role="main">
-        <div ui-view></div>
-    </main>
+    <div ui-view="collectionData"></div>
```
rowasc: just named view for the collection to render here when in datamode

app/main/posts/collections/listing.directive.js
```diff
@@ -132,7 +133,12 @@ function CollectionListingController(
     // they will be directed to the collection page
     function goToCollection(collection) {
         $scope.$parent.closeModal();
-        $location.path(collectionsPath + collection.id + '/' + collection.view);
+        var viewParam = collection.view;
```
rowasc: I'm thinking I might not need this at all? Since we have the redirector anyway.. future-self will check because it's EOD

app/main/posts/savedsearches/savedsearches-routes.js
```diff
@@ -0,0 +1,98 @@
...
+    $stateProvider
+    .state(
+        {
+            url: '^/savedsearches/:id',
+            name: 'savedsearchesRedirector',
+            params: {
+                id: null
+            },
+            onEnter: function ($state, savedSearch) {
+                var viewParam = savedSearch.view;
```
rowasc: same as for collections, take the user where they want to go with their default values

app/main/posts/views/mode-context-form-filter.directive.js
rowasc: to reproduce the erorr I was having, go to sandbox and qa in different tabs/browsers.

Select a saved search. 
See that the sidebar/modecontextbar updates correctly in Sandbox but in QA it doesn't take the corrent count of posts by each survey . I noticed this was happening with categories at least.

app/main/posts/views/post-view-map.controller.js
```diff
@@ -0,0 +1,196 @@
+module.exports = ['PostEndpoint', 'Maps', '_', 'PostFilters', 'Leaflet', '$q', '$rootScope', '$compile', '$location', '$timeout', '$scope',
```
rowasc: mostly just moved the map code here because it gives us as bit of flexibility later on and just a smaller directive class which we could eventually separate further especially to have less logic and endpoint calls in . ? Not sure, maybe it was silly. LMK

rowasc:
@rjmackay I would appreciate a review from you here if you have some time

app/main/posts/views/post-view-data.controller.js
```diff
@@ -55,11 +56,12 @@ function PostViewDataController(
     $scope.selectBulkActions = selectBulkActions;
     $scope.bulkActionsSelected = '';
     $scope.closeBulkActions = closeBulkActions;
-    $scope.selectedPost = {post: null, next: {}};
+    $scope.selectedPost = {post: $scope.$resolve.selectedPost, next: {}};
```
crcommons: Will this error if there is $scope.$resolve.selectedPost is undefined? Is a ternary necessary here?

rowasc: no, because $resolve won't be undefined. An error is raised if you try to access undefinedVar.somethingReal, but not if you try to access somethingReal.undefinedVar
---
But, we probably want a default null on selectedPost tho in the routes, just to  make sure we are not assigning undefined where we want to assign null

app/main/posts/posts-routes.js
```diff
@@ -1,155 +1,158 @@
...
-                $state.go('collection', {view: viewParam, id: $transition$.params().id});
             },
             resolve: {
-                collection: ['$transition$', 'CollectionEndpoint', function ($transition$, CollectionEndpoint) {
-                    return CollectionEndpoint.get({collectionId: $transition$.params().id}).$promise;
+                //change to selectedPost and refactor the selectedposts in general
+                selectedPost: ['$transition$', 'PostEndpoint', function ($transition$, PostEndpoint) {
+                    if ($transition$.params().postId) {
+                        return PostEndpoint.get({ id: $transition$.params().postId }).$promise;
+                    }
```
rowasc: return null if no postId is available (for consistency)

crcommons:
@rowasc I'm getting this error when I try to navigate to data or map view: https://docs.angularjs.org/error/$injector/unpr?p0=eProvider%20%3C-%20e

I can't figure out where the dependency is erroring though. Sorry!

app/app.js
```diff
@@ -111,6 +111,8 @@ angular.module('app',
         // otherwise will take care of routing the user to the specified url
         //$urlRouterProvider.otherwise('/views/map');
 
+    }).config(function ($urlMatcherFactoryProvider) {
+        $urlMatcherFactoryProvider.strictMode(false);
```
rjmackay: Whats strictMode do?

app/main/posts/posts-routes.js
```diff
@@ -1,155 +1,158 @@
...
+    $stateProvider,
+    $urlMatcherFactoryProvider
 ) {
+    $urlMatcherFactoryProvider.strictMode(false);
 
     $stateProvider
     .state(
         {
-            name: 'list',
+            name: 'list', /** I agree, this is a terrible name for this state. Ideas are welcome.**/
```
rjmackay: I'd suggest just using `posts` so you get `posts.map` and `posts.data`

app/main/posts/posts-routes.js
```diff
@@ -1,155 +1,158 @@
...
+                filterState: {value: null, squash: true},
+                savedSearchState: {value: null, squash: true}
+            },
             resolve: {
                 isLoading: function () {
                     return {state: true};
+                },
+                collection: ['$transition$', 'CollectionEndpoint', function ($transition$, CollectionEndpoint) {
+                    if ($transition$.params().collectionId) {
+                        return CollectionEndpoint.get({collectionId: $transition$.params().collectionId}).$promise;
```
rjmackay: What happens if these resolve params fail?

app/main/posts/posts-routes.js
```diff
@@ -1,155 +1,158 @@
 module.exports = [
     '$stateProvider',
+    '$urlMatcherFactoryProvider',
 function (
-    $stateProvider
+    $stateProvider,
+    $urlMatcherFactoryProvider
 ) {
+    $urlMatcherFactoryProvider.strictMode(false);
```
rjmackay: Didn't we already set this?

app/main/posts/posts-routes.js
```diff
@@ -1,155 +1,158 @@
...
-                        $state.go('collection', {id: PostFilters.getModeId(), view: $transition$.params().view}, {reload: true});
-                        break;
-                    default:
-                        var view = $transition$.params().view ? $transition$.params().view : 'map';
-                        if (view === 'list') {
-                            $state.go('list', {view: 'data'}, {reload: true});
-                        }
+            url: '^/collections/:collectionId',
+            name: 'list.collectionRedirector',
+            onEnter: function ($state, $transition$, PostFilters, collection) {
```
rjmackay: 👍 I like this

app/main/posts/posts-routes.js
rjmackay: What happens if there's no postId? In core ngRouter a failing resolve condition would cancel the routing. Does the same happen for uirouter?

app/main/posts/savedsearches/mode-context.html
```diff
@@ -110,7 +110,7 @@ <h1 class="mode-context-title">
         </div>
         <!--// END IF //-->
         <div class="form-field">
-            <a ui-sref="list({view: collectionView})" type="button" class="button-link">
+            <a ui-sref="list.map" ui-sref-opts="{reload: true}" type="button" class="button-link">
```
rjmackay: why do we have to specify reload here? (and in the other places we're using it?)

app/main/posts/views/post-views.controller.js
```diff
@@ -4,14 +4,24 @@ PostViewsController.$inject = ['$scope', '$translate', '$transition$', 'PostFilt
 function PostViewsController($scope, $translate, $transition$, PostFilters) {
     // Set view and layout based out route
     $scope.currentView = $transition$.params().view;
+    $scope.transitionTo = $transition$.to().name;
```
rjmackay: Is this reliable? ie. if I reload the view does it get set to the same thing again?

app/main/posts/views/post-view-map.controller.js
rjmackay: Ah. We want smaller controllers, so that the controllers eventually become components. I would vastly prefer going the other direction -> move code into directive.
If you think the directive should be split we could do that in a follow up. But please make the controller as small as possible

app/app.js
rowasc: >Defines whether URLs should match trailing slashes, or not (the default behavior).

Since we want to allow not matching the trailing slash AFAIK, we need it off .

app/main/posts/posts-routes.js
rowasc: ok sweet, IDK why I didn't think of that.

app/main/posts/posts-routes.js
rowasc: yea but you need to do it for each routing file or it doesn't care for it. That's why we have it in post-routes and common routes , etc.

app/main/posts/posts-routes.js
rowasc: ok so in this scenario we  don't want the resolve to fail if there is no post id, because you are dealing with a possible param in a child state that can be null without the transition needing to fail. Does that make sense? 
If there's no postId, there's nothing wrong with it, it just means yout  don't have a selectedPost.

app/main/posts/savedsearches/mode-context.html
rowasc: Because you want to force a reload of the data, else you don't get the updated lists

app/main/posts/savedsearches/savedsearches.html
```diff
@@ -1,8 +1,5 @@
 <div>
     <layout-class layout="{{layout}}"></layout-class>
```
rowasc: note to self : this file can go away

app/main/posts/views/post-view-map.controller.js
rowasc: yea.. so, I can just revert back to the previous state then. 
The benefit of having a controller, big as it is (which, obviously, is just almost as big as the directive was) is that you can reference the controller functions at any point from a parent which is something we needed at one point because of a weird thing I had done. I guess I can just revert it back now. Will check with you in the afternoon when you are on

app/main/posts/views/post-views.controller.js
rowasc: yea. If you reload, you are still going to the same view and triggering the same transition. .from() can be weird, but to() should not be weird because you are still targeting the same "to "

app/main/posts/posts-routes.js
rowasc: so if the params fail, meaning you don't have a collectionId, nothing happens except you won't have a collection. Now, if the id is there in transition$.params().collectionId but it fails, I think I would need an error handler to take you to a 404.. right? it's something we didn't ever have (right now, if you go to a post that doesn't exist in any version of the web app, you get an empty view.. :( ) . 

I can handle that today

rowasc: TLDR: If a resolve returns a rejected promise, the transition WILL FAIL miserably . Need to handle that

app/main/posts/views/post-view-map.controller.js
rowasc: updated.

PR #914, opened 2017-11-10
auto linkify textual content in posts

tuxpiper:
This is intended as a hotfix to production for a client hosted in ushahidi.io

- enables auto-linkify and other goodies in ng-showdown
- uses linkify filter on text fields

I looked at tackling ushahidi/platform#2244 in the same go (angular-linkify has a twitter linking mode), but I'm not sure how well that would play with the fact that we consider post description (i.e. content) to be markdown by default 

Testing checklist:
- [x] Create a post where the description mentions some URLs (no markdown linking)
- [x] Add some markdown linking to previous post description
- [x] Create additional plain text field, edit previous post to mention some URLs in the new field
- [x] Check out what the post looks like in post detail and list views
- **TODO**: processing / displaying markdown is not safe together with the limitTo filter (applies to the map marker balloon). A markdown / html aware limitTo filter is needed.
- [x] Mention url in the post title
  - The URL should be clickable from the post detail view
  - ... but not from the post list view or the map marker balloon view

Fixes ushahidi/platform#2286 .

Ping @ushahidi/platform


tuxpiper:
I think most people will not be aware that they are editing a markdown field, or a what a markdown field is.

Plus, Github does it :smile: i.e:

www.google.com

tuxpiper:
@rjmackay @Angamanga  I've fixed conflicts with develop . Can we merge?

PR #918, opened 2017-11-13
Routing home and 404; plus reloading post list

crcommons:
- Disabled the post list from reloading when your edit a post; 
- set / and '' to load to map view; 
- set all other unknown routes to reroute to 404 and added 404 html and controller.

Testing checklist:
- [ ] Go to data mode
- [ ] Select a post
- [ ] Select edit
- [ ] Post list should not reload
-------
- [ ] Go to localhost:XXXX 
- [ ] App should reroute to localhost:XXXX/view/map
-------
- [ ] Go to localhost:XXXX/
- [ ] App should reroute to localhost:XXXX/view/map
-------
- [ ] Go to localhost:XXXX/anything
- [ ] App should reroute to 404 with 404 message

Fixes ushahidi/platform# .

Ping @ushahidi/platform


app/common/common-routes.js
```diff
@@ -49,6 +49,14 @@ module.exports = ['$stateProvider', function ($stateProvider) {
                 template: ''
             }
         )
+        .state(
+            {
+                name: '404',
+                url: '/404',
+                controller: require('./auth/404.controller.js'),
```
rjmackay: Could we just make this a component instead? especially since it has an empty controller.

rowasc: @crcommons will you check into this in this PR or look into it in a separate one? Not sure if I should test now or after this is addressed.

rowasc:
$state.defaultErrorHandler(function(error) {
            $state.go('404'); 
});
adding this to app.js in a run block would also make sure that if the user tries to access a resource that fails to resolve (such as /post/99999999) we go to a 404 (vs an empty page as it is now)
Not sure what you think @rjmackay , I couldn't find a neat(er) way to do it Tbh.  
Probably can send an error message too as a param to make it more specific but I'm not sure where we could get the proper message from (maybe using the viewname in $state.from and have take it from en.js )? )

rjmackay:
@rowasc that sounds good to me

PR #919, opened 2017-11-14
[WIP ] Uirouter migration dataeditmode cancel

rowasc:
-
- TODO : change the base branch to merge to uirouter-migration-init  **after** we get https://github.com/ushahidi/platform-client/pull/912 merged in. 

Testing checklist:
- [ ] Literally everything in the data mode. 

Fixes ushahidi/platform# .

Ping @ushahidi/platform


app/main/posts/common/post-actions.directive.js
```diff
@@ -55,7 +55,7 @@ function PostActionsDirective(
                     // Redirect to list
                     $location.path('/views/data');
                 } else {
-                    $state.go($state.$current, null, { reload: true }); // in favor of $route.reload();
+                    $state.go('list.data', null, { reload: true }); // in favor of $route.reload();
```
rjmackay: `$state.reload()` ?

app/main/posts/modify/post-data-editor.controller.js
```diff
@@ -84,20 +86,51 @@ function PostDataEditorController(
         $scope.savePost();
     });
 
-    $scope.$on('event:edit:leave:form', function () {
-        if ($scope.parentForm.form && $scope.parentForm.form.$dirty) {
-            $scope.leavePost();
-        } else {
-            $scope.cancel();
+    $transitions.onStart({}, function (transition) {
```
rjmackay: I don't really understand what this is doing?

app/main/posts/modify/post-data-editor.controller.js
```diff
@@ -84,20 +86,51 @@ function PostDataEditorController(
...
-     */
+
+    function leavePost(transition) {
+        /**
+         * just a wrapper function because we were doing the same thing
+         * over and over again in the promise
+         * @param resolve
+         * @param reject
+         */
+        var cancelResolveFn = function (resolve, reject) {
```
rjmackay: Generally prefer named functions over anonymous functions assigned to a var

app/main/posts/modify/post-data-editor.html
```diff
@@ -106,7 +106,7 @@
         </post-tabs>
       <div class="toolbar toolbar-secondary">
         <div class="button-group">
-          <button class="button-flat" ng-click="leavePost()" translate>app.cancel</button>
+          <button class="button-flat" ui-sref="list.data.detail({postId: post.id})" translate>app.cancel</button>
```
rjmackay: I see you've changed this handling to watch transitions, instead of directly calling `leavePost` to check for changes. Is the main reason because now you can watch for *all* state changes, not just the button press?

app/main/posts/modify/post-data-editor.controller.js
```diff
@@ -301,22 +308,24 @@ function PostDataEditorController(
...
-            PostLockEndpoint.unlock({
-                id: $scope.post.lock.id,
-                post_id: $scope.post.id
-            }).$promise.then(function (result) {
-                doChangePage(url);
-            });
-        } else {
-            doChangePage(url);
-        }
+    function cancel() {
```
rjmackay: These function names are really unhelpful. We've not cancelling, we're mid way through the cancelling process. 
This should probably be something more like `unlockPost`

app/main/posts/modify/post-data-editor.controller.js
```diff
@@ -84,20 +86,51 @@ function PostDataEditorController(
...
-    /**
-     * @uirouter-refactor check this and add back if neded
-     * var $locationChangeStartHandler = $scope.$on('$locationChangeStart', function (e, next) {
-        $scope.leavePost(next, e);
+        return true;
+
     });
-     */
+
+    function leavePost(transition) {
```
rjmackay: I think renaming this might make things clearer. ie. `checkUnsaved` or `saveAndLeave` ... I don't know but its hard to follow whats going on here.

app/main/posts/modify/post-data-editor.controller.js
```diff
@@ -84,20 +86,51 @@ function PostDataEditorController(
...
+            $scope.savingPost.saving = false;
+            return $scope.cancel().then(function () {
+                resolve(true);
+            }).catch(function () {
+                reject(false);
+            });
+        };
+
+        return new Promise (function (resolve, reject) {
+            if (!$scope.parentForm.form || ($scope.parentForm.form && !$scope.parentForm.form.$dirty)) {
```
rjmackay: Add some comments? ie. `// Do we have unsaved changes?`

app/main/posts/modify/post-data-editor.controller.js
```diff
@@ -84,20 +86,51 @@ function PostDataEditorController(
...
+                cancelResolveFn(resolve, reject);
+            } else {
+                // @uirouter-refactor if we end up having onbeforeunload features,we need to add this back
+                // if (ev) {
+                //     ev.preventDefault();
+                // }
+                Notify.confirmLeave('notify.post.leave_without_save').then(function () {
```
rjmackay: How does this work if both confirm or cancel lead to the same function??

app/main/posts/common/post-actions.directive.js
rowasc: no. Because you want to explicitly reload list.data , not whatever state you are in (you could be in list.data.detail, and you don't care about reloading that one.

app/main/posts/modify/post-data-editor.controller.js
rowasc: $transitions.onStart is a handler for when a transition starts. When a transition starts here you want to either let it go or cancel it here because you need the leavePost functionality with the warning modal and all that. So you have ignoreCancelEvent for a special case where you literally are transitioning after the user accepts (and so you don't want to show it again) and you have the transition.from().name making sure you were in a list.data.edit state (editing mode) , and if you are there, you either resolve or reject a promise depending on form state, user actions, etc. Resolve will let the transition continue. Reject will stop the transition meaning you stay in the edit mode
If you are not editing, you just return true to let the transition continue.

app/main/posts/modify/post-data-editor.html
rowasc: I changed this because it represents what you are trying to do in a better way than leave post did and it's more consistent to what you have to do everywhere in the cancel button. For instance, you now can use the same ui-sref in post-toolbar button and it will behave EXACTLY the same as this cancel button, versus being "kind of the same but it could also be completely different who knows" which was what we had =)

app/main/posts/modify/post-data-editor.controller.js
rowasc: for 25% of the cases, it doesn't =) 
thanks, I had missed the fact that because there's no edit mode bool I needed to do a reject .

app/main/posts/common/post-actions.directive.js
rjmackay: Got it

app/main/posts/modify/post-data-editor.controller.js
rjmackay: I'm confused. Why do both save and continue with out saving follow the same path? Where do we actually trigger saving?

rowasc: in the confirmLeave function there's a save function that resolves or reject depending on whether or not your save is successful. So resolve means it saved and you can continue. Reject means it didn't save and you CANNOT continue. 
 It also has a confirm function that always resolves to let the user continue without saving.

PR #920, opened 2017-11-15
Adding loading-functionality

Angamanga:
- Adding a loadingProgress-service
- Using the service instead of sending isLoading-variables between directives
- 

Testing checklist:
- [x] Go to data-view
- [x] Loading-bar and dots should be visible while loading posts
- [x] Select a post, the loading bar should be visible until the full post is loaded
- [x] Choose to edit, the loading-bar should be visible until the full edit mode is loaded
- [x] Make a change and save, the loading bar should be visible while saving and the saving-button should say saving and be disabled
- [x] Select filters and click apply, loading bar should be visible until all posts are loaded
- [x] bulk-delete posts, loading bar should be visible until all posts are loaded
- [x] delete one post, loading bar should be visible while deleting
- [x] bulk-change statuses on a number of posts, loading-bar should be visible while updating
- [x] change status on one post, loading-bar should be visible while updating

- [x] go to map-view
- [x] loading bar should be visible while loading posts
- [x] click on one post-card, loading-bar should be visible before/while redirecting to detail-view
- [x] delete one post, loading-bar should be visible while deleting
- [x] select filters, loading-bar should be visible while applying filters

TODO:
- [x] Check loading-bar for saved-searches and collections.
- [x] Adjust loading-bar for editing/saving a saved search
- [ ] Add tests for new service

Fixes ushahidi/platform# .

Ping @ushahidi/platform


app/main/posts/common/post-actions.directive.js
```diff
@@ -47,7 +49,10 @@ function PostActionsDirective(
         }
 
         function deletePost() {
+            LoadingProgress.setLoadingState({isLoading: true});
```
Angamanga: Sometimes the loading-state needs to be set manually since no transitions are made, thats why I am setting it sometimes and sometimes not.

app/main/posts/detail/post-messages.directive.js
```diff
@@ -40,7 +40,6 @@ function (
             $scope.totalItems = $scope.itemsPerPage;
             $scope.pageChanged = getMessagesForPagination;
             $scope.showPagination = false;
-            $scope.isLoadingMessages = false;
```
Angamanga: Removing the loading bar for messages since there already is a loading-bar in the toolbar = 2 loading bars in the same view = confusing.

Angamanga:
@rowasc @rjmackay I am not totally sure of this solution, for example, don't like the watcher in the loading.progress-service but have not figured out another way of doing it. Is there a better way that I haven't thought of?

app/main/posts/common/post-actions.directive.js
rowasc: why are we using the full state object instead of sending a bool, or even having a setLoadingStateOn/setLoadingStateOff ? I can't remember if there was a reason or just preference

rjmackay:
Is the loading functionality *broken* right now? If not, can we abandon this and reconsider next sprint?
We have lots more bugs to crank on right now.

app/main/posts/common/post-actions.directive.js
Angamanga: Its because we need to separate loading and saving, we don't want to make the save-button say 'saving' as soon as something is loading, only when it is actually loading and I tried to be clever and put them at the same place... not sure if thats the best way though...

Angamanga:
@rjmackay Yes it was broken because of isLoading-variables that we sent between the directives. Have added http-interceptors to avoid setting the loading-state manually now so this is now working as described in the testing-checklist

app/app.js
```diff
@@ -140,7 +140,36 @@ angular.module('app',
         return window.ushahidi.bootstrapConfig ?
             _.indexBy(window.ushahidi.bootstrapConfig, 'id') :
             { map: {}, site: {}, features: {} };
-    }])// inject the router instance into a `run` block by name
+    }])
+    .factory('loading', function (LoadingProgress, $q, $injector) {
```
Angamanga: Does this belong here, or should it live somewhere else?

app/app.js
```diff
@@ -139,12 +139,44 @@ angular.module('app',
             _.indexBy(window.ushahidi.bootstrapConfig, 'id') :
             { map: {}, site: {}, features: {} };
     }])
-    // .run(function ($uiRouter, $trace, $location) {
-    //     let Visualizer = require('@uirouter/visualizer').Visualizer;
+    .factory('loading', function (LoadingProgress, $q, $injector) {
```
rjmackay: Can we move this to a file in common?

app/app.js
```diff
@@ -139,12 +139,44 @@ angular.module('app',
...
+                }
+                // we want this to trigger isSaving everytime except when a lock-updates
+                if (config.method === 'PUT' && config.url.indexOf('lock') === -1) {
+                    LoadingProgress.setSavingState(true);
+                }
+                return config;
+            },
+            response: function (response) {
+                var httpService = $injector.get('$http');
+                if (httpService.pendingRequests.length === 0) {
```
rjmackay: Ooh. I wondered how to handle multiple requests at once. This is neat.

app/common/services/loadingProgress.service.js
```diff
@@ -0,0 +1,52 @@
...
+function LoadingProgress($rootScope, $transitions) {
+    function watchTransitions() {
+        $transitions.onStart({}, function (transition) {
+            $rootScope.isLoading = true;
+            if (transition.from().name === 'list.data.edit') {
+                $rootScope.isSaving = true;
+            }
+        });
+
+        $transitions.onFinish({}, function (transition) {
```
rjmackay: What happens if a transition finishes before the http requests? ie. should this check for http requests in progress first?

app/main/posts/modify/post-data-editor.controller.js
```diff
@@ -72,6 +74,12 @@ function PostDataEditorController(
     $scope.submitting = $translate.instant('app.submitting');
     $scope.hasPermission = $rootScope.hasPermission('Manage Posts');
     $scope.selectForm = selectForm;
+    $scope.isSaving = LoadingProgress.getSavingState();
+
+    LoadingProgress.subscribeOnSavingState(function (isSaving) {
```
rjmackay: This is tricky... it adds a watch, which will end up hanging around after moving to another view. We either need to unbind the watch on $scope destroy - kinda messy again. Or the easier method would be to call `getSavingState` direct ie.
`$scope.isSaving = LoadingProgress.getSavingState`
Then use `isSaving()` instead of `isSaving` in the templates. I think angular still does a watch behind the scenes but it happens automatically

app/main/posts/views/post-view-map.controller.js
```diff
@@ -0,0 +1,191 @@
+module.exports = ['PostEndpoint', 'Maps', '_', 'PostFilters', 'Leaflet', '$q', '$rootScope', '$compile', '$location', '$timeout', '$scope',
```
rjmackay: I think this is accidentally reverting changes from the main branch?

app/common/services/loadingProgress.service.js
Angamanga: Good point, yes, it should check that first...

app/app.js
Angamanga: Fixed

app/main/posts/modify/post-data-editor.controller.js
Angamanga: Good idea! And it is working :) :) Will change it.

app/main/posts/views/post-view-map.controller.js
Angamanga: yep, true

PR #924, opened 2017-11-16
Uirouter migration bug fix - Remove unused false path in logout and add the logout redirect

rowasc:
- Add the logout redirect to map
- Remove path that never happens 

Testing checklist:
- [ ] Logout
- [ ] Check that you are redirected to the map after logging out

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/common/auth/authentication-events.run.js
+++ b/app/common/auth/authentication-events.run.js
@@ -52,7 +52,7 @@ function AuthenticationEvents($rootScope, $location, Authentication, Session, _,
                 $location.url(redirect);
             }
 
-            $state.go($state.$current.name ? $state.$current : 'map', null, { reload: true });
+            $state.go('posts.map');
         });
     }
 
--- a/app/common/auth/authentication-events.run.js
+++ b/app/common/auth/authentication-events.run.js
@@ -11,7 +11,7 @@ function AuthenticationEvents($rootScope, $location, Authentication, Session, _,
         if (Authentication.getLoginStatus()) {
             doLogin(false, true);
         } else {
-            doLogout(false);
+            doLogout();
         }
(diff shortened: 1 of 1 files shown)


app/common/auth/authentication-events.run.js
```diff
@@ -42,17 +42,13 @@ function AuthenticationEvents($rootScope, $location, Authentication, Session, _,
...
         $rootScope.loggedin = false;
         // we don' wat to reload until after filters are correctly set with the backend default that the user
         // would get when logged out
         PostFilters.resetDefaults().then(function () {
-            if (redirect) {
-                $location.url(redirect);
-            }
-
-            $state.go($state.$current.name ? $state.$current : 'map', null, { reload: true });
+            $state.go('posts.map');
```
rjmackay: I don't think this should be the case. If I'm on a post detail view and I log out... I should stay on the same view.

rowasc: ok. 
This was reported as a requirement/bug so I thought it was the expected behavior. Will change it..

rjmackay: Hrm. I might be wrong? Where was it logged/who was it logged by?

rowasc: https://ushahidi.slack.com/archives/C03UU507R/p1510670345000387 Ah, ok I think Carolyn reported it to me in chat, and asked here. 
But we might be wrong?

rjmackay:
Closing. This isn't the correct behaviour.
Currently we keep you on the same page and try to reload it. I think thats right, with the exception that if you were on say settings or a draft post you should get redirected to map.

PR #926, opened 2017-11-16
Uirouter - Delete and status change with next post selected fixed

rowasc:
Properly removes posts from detail/list when they do not match filters; selects next post in detail view when posts do not match filters

Testing checklist:
Data View
- [x] Filter posts to show only published posts; Change the status of a post using the ...  button; be careful not to select a post; 
- The post should be removed from the list
- No post should be selected for detail view
- [x] Filter posts to show only published posts; Change the status of posts using the bulk actions dropdown; be careful not to select a post;
- The post should be removed from the list
- No post should be selected for detail view
- [x] Delete a post in data view.
 - Post should be removed from list 
 - The data view should not reload completely
-------
Data Detail View
- [x] Filter posts to show only published posts; Select a post, but not the first post; Change the status of a post using the ... button; be careful not to select a post; 
- The post should be removed from the list
- The next post in line should be displayed in detail view
- The post list should not show the selected post at the top of the list. The list should be the same minus the removed post
- [x] Search for posts using term (I suggest 'test'); Select a post; Remove all references to 'test' so that it will no longer match the filter query; Save post;
- The post should be removed from the list
- The next post in line should be displayed in detail view
- [x] select a post with no filters set; then set the filters so that the selected post does not match these filters
- The post should be removed from list
- The post should no longer be displayed in detail view
- The next post in line should be in detail view

Fixes ushahidi/platform# .

Ping @ushahidi/platform


app/main/posts/common/post-actions.directive.js
```diff
@@ -53,7 +53,7 @@ function PostActionsDirective(
                     // Redirect to list
                     $location.path('/views/data');
                 } else {
-                    $state.go('posts.data', null, { reload: true }); // in favor of $route.reload();
+                    $rootScope.$broadcast('event:edit:post:status:data:mode:saveSuccess', {post: $scope.post, deleted: true});
```
rjmackay: Just double check... does this work for *everywhere* that post-actions is used?

app/main/posts/common/post-actions.directive.js
```diff
@@ -51,9 +51,12 @@ function PostActionsDirective(
...
                     $location.path().indexOf('savedsearches') === -1) {
                     // Redirect to list
-                    $location.path('/views/data');
+                    $state.go('posts.data');
+                } else if ($state.current.name === 'posts.data' || $state.current.name === 'posts.data.detail') {
+                    // we only want to do this in the data detail view because it's where you can really update a list
+                    $rootScope.$broadcast('event:edit:post:status:data:mode:saveSuccess', {post: $scope.post, deleted: true});
                 } else {
-                    $state.go('posts.data', null, { reload: true }); // in favor of $route.reload();
+                    $state.reload();
```
rowasc: changed to $state.reload to emulate https://github.com/ushahidi/platform-client/blob/225ecb43baef0011cb995b53b26e5dc12670c210/app/main/posts/common/post-actions.directive.js#L58

+ kept the broadcast because that's what was being used for archive to work in the old data view/etc but wasn't there for delete .

`Works in all views. `

app/main/posts/common/post-actions.directive.js
```diff
@@ -51,9 +51,12 @@ function PostActionsDirective(
                     $location.path().indexOf('collections') === -1 &&
                     $location.path().indexOf('savedsearches') === -1) {
                     // Redirect to list
-                    $location.path('/views/data');
+                    $state.go('posts.data');
+                } else if ($state.current.name === 'posts.data' || $state.current.name === 'posts.data.detail') {
+                    // we only want to do this in the data detail view because it's where you can really update a list
+                    $rootScope.$broadcast('event:edit:post:status:data:mode:saveSuccess', {post: $scope.post, deleted: true});
```
rjmackay: I suspect we could *always* broadcast the event... and unless we're on the data view there shouldn't be any listeners

app/main/posts/common/post-actions.directive.js
```diff
@@ -51,9 +51,12 @@ function PostActionsDirective(
                     $location.path().indexOf('collections') === -1 &&
                     $location.path().indexOf('savedsearches') === -1) {
                     // Redirect to list
-                    $location.path('/views/data');
+                    $state.go('posts.data');
+                } else if ($state.current.name === 'posts.data' || $state.current.name === 'posts.data.detail') {
```
rjmackay: This should really use `$state.includes('posts.data')`

app/main/posts/common/post-actions.directive.js
```diff
@@ -51,9 +51,12 @@ function PostActionsDirective(
                     $location.path().indexOf('collections') === -1 &&
```
rjmackay: These could be `!$state.includes('posts')` I think

app/main/posts/views/post-view-data.controller.js
```diff
@@ -92,7 +92,7 @@ function PostViewDataController(
     // filters on the frontend. This makes it MUCH faster than using the API
     $rootScope.$on('event:edit:post:status:data:mode:saveSuccess', function (event, args) {
```
rjmackay: Forgot to note... I know this was here already, but its a little worrying listening on $rootScope because it won't get unbound when navigating away.
Does this still work if we use `$scope.$on`? Thats much safer, but I can't remember if it'll see the event on a directive scope.

app/main/posts/common/post-actions.directive.js
rowasc: ah, maybe. I'll check (but, that's depending on the not-using-rootscope thing that I need to check too)

app/main/posts/common/post-actions.directive.js
rowasc: ah neat I didn't think of that.

rowasc:
citing 
> What should it be? If we delete a post ...
... and we're on the post detail we should go to posts.data
... if we're on the map or data view we should probably stay where we are right?
Regardless of either of these, events should really broadcast all the time, not just in some states.

answer:
- if we are on map view we need to stay where we are but reload, because we need to card to go away.
- if we are on data view we stay there, but it needs to select the next post if there is one
- the detail edit mode will be deleted soon but it also reloads/should reload there.

app/main/posts/common/post-actions.directive.js
```diff
@@ -44,14 +44,14 @@ function PostActionsDirective(
 
         function deletePost() {
             PostActionsService.delete($scope.post).then(function () {
-                // If we're not already on some top level view
-                if ($location.path().indexOf('views') === -1 &&
-                    $location.path().indexOf('collections') === -1 &&
-                    $location.path().indexOf('savedsearches') === -1) {
+                $rootScope.$broadcast('event:edit:post:status:data:mode:saveSuccess', {post: $scope.post, deleted: true});
```
rowasc: broadcasting all the time , and listening on $scope in data .

app/main/posts/common/post-actions.directive.js
```diff
@@ -44,14 +44,14 @@ function PostActionsDirective(
...
+                // If we're not already on some of the posts views
+                if (!$state.includes('posts')) {
```
rowasc: it it's not a post.* view IDK how you even deleted a post, but we're taking you to the posts.data page anyway.

app/main/posts/common/post-actions.directive.js
```diff
@@ -44,14 +44,14 @@ function PostActionsDirective(
...
                     // Redirect to list
-                    $location.path('/views/data');
-                } else {
-                    $state.go('posts.data', null, { reload: true }); // in favor of $route.reload();
+                    $state.go('posts.data');
+                } else if ($state.includes('posts.map')) {
+                    // only map needs to reload
+                    $state.reload();
```
rowasc: @rjmackay  I am reloading which is what this used to do anyway , but only in the map

app/main/posts/views/post-view-data.directive.js
```diff
@@ -101,16 +101,18 @@ function PostViewDataController(
         }
     });
 
+
```
rowasc: moved to data directive :) was in the ctrl which was deleted ha

app/main/posts/common/post-actions.directive.js
rjmackay: Yup. This logic is a little backwards but much easier to parse not at least :)

app/main/posts/views/post-view-data.directive.js
```diff
@@ -101,16 +101,18 @@ function PostViewDataController(
...
     // And this is for when you change the post status using the ... button
     // Because it's just the post status only, we're just checking if it matches
     // filters on the frontend. This makes it MUCH faster than using the API
-    $rootScope.$on('event:edit:post:status:data:mode:saveSuccess', function (event, args) {
+    $scope.$on('event:edit:post:status:data:mode:saveSuccess', function (event, args) {
```
rjmackay: Yay. So these just work as $scope listeners? Much safer

rjmackay:
I made some updates to this, and merged the main branch back in but I haven't actually tested this yet. The code looks ok, but not sure if it all works.

crcommons:
@rowasc One final issue: 
-set filters to published only
-select a post
-change status of that post to under review
-post will disappear from that list but it will still show in post detail. 
-Post should be removed from data list and then next post should be selected

rjmackay:
Minor issue:
- select a post with no filters set; then set the filters so that the selected post does not match these filters
- The post should be removed from list
- The post should no longer be displayed in detail view
- The next post in line should be in detail view

however the post shown in detail view isn't shown as selected in the post list

rjmackay:
Fixed.

PR #927, opened 2017-11-16
Route to components where possible

rjmackay:
- Route post detail, edit, data view and map as components not template+controller
- Revert splitting directives - mostly to minimize the diff vs develop so I can audit changes
- Remove post-view level above map + data view
- Try to make sure we're populating all the bindings again
- Remove a bunch of defunct code \o/

Ping @ushahidi/platform


app/main/posts/posts-routes.js
```diff
@@ -115,30 +107,36 @@ function (
...
             params: {
                 view: {value: 'map', squash: true},
                 filterState: {value: null, squash: true}
+            }
+        }
+    )
+    .state(
+        {
+            url: '/views/map',
+            name: 'posts.map.all',
```
rjmackay: Added this as a child so that we can define the default modeContext

test/unit/inject.spec.js
```diff
@@ -1,51 +0,0 @@
-describe('injecting things', function () {
```
rjmackay: Not really sure what this was testing??

app/main/posts/common/post-actions.directive.js
```diff
@@ -69,18 +69,7 @@ function PostActionsDirective(
...
-                // At this point we are not certain we will switch to this Post so we back it up
-                // in anticipation of using it later if the current Post exists corectly
-                $scope.selectedPost.next = _.clone($scope.post);
-                $rootScope.$broadcast('event:edit:leave:form');
-            } else {
-                $scope.selectedPost.post = _.clone($scope.post);
-                $scope.editMode.editing = true;
-                $state.go('posts.data.edit', {postId: $scope.post.id});
-            }
+            $state.go('posts.data.edit', {postId: postId});
```
rjmackay: Not 100% sure about this, but its vastly simpler since we catch the transition in the editor now.

One issue I noticed was that the save action redirects us to post-detail though... so we don't actually get to the post-edit for the next post. Probably something to resolve in the next PR though

app/main/posts/collections/mode-context.directive.js
```diff
@@ -5,6 +5,10 @@ CollectionModeContext.$inject = [];
 function CollectionModeContext() {
     return {
         restrict: 'E',
+        scope: {
```
rjmackay: These get injected from the route `resolve` params now. A bit cleaner

rjmackay:
I've tried to fix the basics of post data detail and edit. There's more to go but I don't want to make this PR huge.

tuxpiper:
That's an epic changelog :smile: :+1:

app/main/posts/posts-routes.js
```diff
@@ -115,30 +107,36 @@ function (
...
+            views: {
+                'mode-context': 'modeContext'
             },
             onEnter: function ($state, $transition$, PostFilters, savedSearch) {
-                if (!savedSearch) {
-                    PostFilters.resetDefaults();
-                }
+                PostFilters.setMode('all');
```
rowasc: we cannot do resetDefaults and setMode here because the user needs to be able to move around and keep their filters.

rowasc: https://github.com/ushahidi/platform-client/pull/931

PR #929, opened 2017-11-17
Simplify editMode checks by relying on state instead

rjmackay:
- Rip out editMode wherever I could
- Use $state.$current to check where we are instead

Relies on https://github.com/ushahidi/platform-client/pull/927

### DIFF
--- a/app/main/posts/views/post-toolbar.directive.js
+++ b/app/main/posts/views/post-toolbar.directive.js
@@ -7,7 +7,6 @@ function PostToolbarDirective() {
         scope: {
             isLoading: '=',
             filters: '=',
-            editMode: '=',
             selectedPost: '=',
             savingPost: '='
         },
@@ -22,26 +21,32 @@ function PostToolbarController($scope, $rootScope, Notify, PostLockService, $sta
     $scope.savePost = savePost;
     $scope.hasPermission = $rootScope.hasPermission('Manage Posts');
     $scope.editEnabled = editEnabled;
-    $scope.saveButtonEnabled = saveButtonEnabled;
+    $scope.editMode = editMode;
+    $scope.cancel = cancel;
+
     function editEnabled() {
+        if (!$scope.selectedPost || !$scope.hasPermission) {
+            return false;
+        }
+
         return $scope.selectedPost ? !PostLockService.isPostLockedForCurrentUser($scope.selectedPost) : false;
     }
 
     function savePost() {
         $rootScope.$broadcast('event:edit:post:data:mode:save');
     }
-    function saveButtonEnabled() {
+
+    function editMode() {
         return $state.$current.name === 'posts.data.edit';
     }
 
     function setEditMode() {
         if (editEnabled()) {
-            if ($scope.editMode.editing) {
-                $state.go('posts.data.detail', {postId: $scope.selectedPost.id});
-            } else {
-                $state.go('posts.data.edit', {postId: $scope.selectedPost.id});
-                $scope.editMode.editing = true;
-            }
+            $state.go('posts.data.edit', {postId: $scope.selectedPost.id});
         }
+    }
+
+    function cancel() {
+        $state.go('posts.data.detail', {postId: $scope.selectedPost.id});
     }
 }
--- a/app/main/posts/views/post-toolbar.html
+++ b/app/main/posts/views/post-toolbar.html
@@ -6,36 +6,45 @@
 
     <filter-posts filters="filters" embed-only="false"></filter-posts>
 
-    <div class="button-group" embed-only="false" ng-switch="editMode.editing">
-        <post-share ng-switch-when="false" filters="filters" button="true"></post-share>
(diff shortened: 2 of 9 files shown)


app/main/posts/views/post-toolbar.directive.js
```diff
@@ -22,26 +21,32 @@ function PostToolbarController($scope, $rootScope, Notify, PostLockService, $sta
...
-    $scope.saveButtonEnabled = saveButtonEnabled;
+    $scope.editMode = editMode;
+    $scope.cancel = cancel;
+
     function editEnabled() {
+        if (!$scope.selectedPost || !$scope.hasPermission) {
+            return false;
+        }
+
         return $scope.selectedPost ? !PostLockService.isPostLockedForCurrentUser($scope.selectedPost) : false;
```
rjmackay: Oops I think I'm doubling up the selectedPost check. Oh well

app/main/posts/views/post-toolbar.directive.js
```diff
@@ -22,26 +21,32 @@ function PostToolbarController($scope, $rootScope, Notify, PostLockService, $sta
...
-                $state.go('posts.data.detail', {postId: $scope.selectedPost.id});
-            } else {
-                $state.go('posts.data.edit', {postId: $scope.selectedPost.id});
-                $scope.editMode.editing = true;
-            }
+            $state.go('posts.data.edit', {postId: $scope.selectedPost.id});
         }
     }
+
+    function cancel() {
```
rjmackay: Split this out of `setEditMode` ... easier to follow

app/main/posts/views/post-toolbar.html
```diff
@@ -6,36 +6,45 @@
...
-        <button ng-switch-when="true" class="button-alpha" disabled ng-if="savingPost.saving" translate>app.saving
-            <div class="loading">
-                <div class="line"></div>
-                <div class="line"></div>
-                <div class="line"></div>
-            </div>
-          </button>
-
+    <div class="button-group" embed-only="false">
+        <div ng-if="!editMode()">
```
rjmackay: Dropped the horror that is `ng-switch`

app/main/posts/views/post-toolbar.html
```diff
@@ -6,36 +6,45 @@
...
+            <post-share filters="filters" button="true"></post-share>
+            <button ng-if="editEnabled()" class="button button-alpha" ng-click="setEditMode()" translate>app.edit</button>
+        </div>
+
+        <div ng-if="editMode()">
+            <button ng-if="hasPermission" class="button" ng-click="cancel()" translate>app.cancel</button>
```
rjmackay: It kinda looks like I could bump the `hasPermission` check up a couple of levels since we don't show much of anything without it.

app/main/posts/views/post-view-data.directive.js
```diff
@@ -214,24 +212,26 @@ function PostViewDataController(
...
-                        $scope.selectedPost.post = tmpPost[0];
-                        $scope.selectedPostId = tmpPost[0].id;
-                        $scope.editMode.editing = false;
-                    }
-                }
-            }
-        });
+        // I don't know what this does but its terrifying
+        //
+        // $scope.$watch(function () {
```
rjmackay: @willdoran or @rowasc if you can explain what this did it would really help. Then we can hopefully reproduce it with something less scary looking

app/main/posts/views/post-view-data.directive.js
```diff
@@ -244,23 +244,20 @@ function PostViewDataController(
 
     function confirmEditingExit() {
         var deferred = $q.defer();
-        if (!$scope.editMode.editing) {
+        if ($state.$current.name !== 'posts.data.edit') {
```
rjmackay: Hey look `$state` already knows where we are!

app/main/posts/views/post-view-data.directive.js
rowasc: ehh. Robbie I dont know what branch you based this off, but that code was GONE , same as the editMode checks from below in my PR from a couple days ago were I completely replaced it.. we even discussed it.

rowasc: ok this is proof that I was both in need of coffee and that we have very similar looking weirdness in both data and data editor, because I replaced something similar a couple days ago which is what we actually discussed.  The stuff I found and fixed was the data editor version of this , which was just as wrong and scary. 
Let me take a look today,we shouldn't need it anymore . This was used to manage the selected post as the URL changed. I think we don't need to replace it at all, because it's done in the route state management  but I'll check and report back. 
I'm really sorry about the confusion :(

app/main/posts/views/post-view-data.directive.js
rowasc: yay. Hate hate hate the editmode checks

rowasc:
note:  edit modal checks are not really working. Need to figure this out  .

PR #930, opened 2017-11-17
2116 data view mobile

Angamanga:
- adds a button to the post-data-detail-view to be able to deselect a post-card in mobile
- adds new styles to mobile-screens in post-data-view. It should work as following:
 - if you go to /views/data in on a smaller screen, you should see the post-list.
 - if you click on a post-card, the detailed-view should be visible and the post-list hidden
- if you click on the post-card-close-button, the post-card should hide and the post-list should become visible again.
- allows you to navigate from map view or use direct link to post and view post in correct mobile or large view
- Lists selected post and 19 previous posts when navigating to detail view directly

Testing checklist:
In regular (large) view
- [x] Select a post from data view; post should display on right side of screen and postcard should be selected on left
- [x] Select various posts and confirm that selected post detail changes accordingly
- [x] Edit a post and confirm it still edits correctly and is selected when it is supposed to be
- [x] Select a post from map view; post should display on right side of screen and postcard should be selected at the TOP and 19 posts should be listed after it
- [x] visit a post-detail link directly (or reload page from previous step);  post should display on right side of screen and postcard should be selected at the TOP and 19 posts should be listed after it
----------
In small/mobile view
- [x] Select a post from data view; list should disappear and post should display; Edit button should be available
- [x] Close selected post; list should appear and NO post should be selected; Edit button should not be available
- [x] Select a post from map view; post should display without list
- [x] Close selected post; you should be rerouted to map view
- [x] visit a post-detail link directly (or reload page from previous step);  post should display without list
- [x] Close selected post; you should be routed back to data view

Fixes ushahidi/platform#2116

Ping @ushahidi/platform


app/main/posts/detail/post-detail-data.directive.js
```diff
@@ -83,6 +87,12 @@ function PostDetailDataController(
             });
         }
 
+        // Route and display accordingly if the we are in mobile or tablet view
+        let currentWidth = $window.innerWidth;
+        if (currentWidth < 1023) {
```
rjmackay: One issue with this is that if the user resizes the window after the initial load this won't have been run for them, yielding weird results. We should probably run this for everyone and always set the active classes.

app/main/posts/detail/post-detail-data.directive.js
```diff
@@ -226,5 +229,19 @@ function PostDetailDataController(
                 Notify.apiErrors(errorResponse);
             });
     };
+
+    $scope.close = function () {
```
rjmackay: So I changed approach a bit. Instead of trying to call the parent directive function, we just try to use whatever the previous state was.

app/main/posts/views/post-view-data.directive.js
```diff
@@ -99,49 +104,28 @@ function PostViewDataController(
...
-        filters.forEach((status) => {
-            if (postObj.status === status) {
-                matchingStatus = true;
-            }
-        });
-
-        return matchingStatus;
-    }
+    // Cleanup and remove all listeners
+    $scope.$on('$destroy', () => {
```
rjmackay: Did some clean up while here to avoid leaking a bunch of event listeners everytime we go to data view

app/main/posts/views/post-view-data.directive.js
```diff
@@ -99,49 +104,28 @@ function PostViewDataController(
...
-                    $scope.selectedPostId = nextInLine.id;
-                } else {
-                    $scope.selectedPost = {post: null, next: {}};
-                    $scope.selectedPostId = null;
-                    getPosts(false, false);
-                }
-            }
-        });
-    }
+    unbindFns.push($transitions.onSuccess({
```
rjmackay: Change to handling this via events. We check each state change to see which column should be active.

app/main/posts/posts-routes.js
```diff
@@ -70,7 +70,8 @@ function (
             name: 'posts.data',
             params: {
                 view: {value: 'data', squash: true},
-                filterState: {value: null, squash: true}
+                filterState: {value: null, squash: true},
+                activeCol: {value: 'timeline', squash: true}
```
rjmackay: This is what now toggles which column is visible for mobile views.

app/main/posts/posts-routes.js
```diff
@@ -156,16 +163,16 @@ function (
     .state(
         {
             name: 'posts.data.detail',
-            url: '/posts/:postId',
+            url: '^/posts/:postId',
```
rjmackay: Tweaked this to return to the original url of `/post/123` not `/views/data/posts/123`

rjmackay:
So I hit an issues that the active columns didn't work right if you resized from large to small screen size. My attempt to fix that hit some further issues with setting active classes, so I ended up change the method for those. All in all though this now looks ready to go.

rjmackay:
So I'm 99% sure this is fine. However the tests don't work because we've dropped the old controller. ideally we'd migrate those to test the directive.

crcommons:
I think I got the last bugs in this... it now clears the selected post when you click the 'x' button to close a post, that way the edit button doesn't show when no post is selected. It was also erroring when it tried to reroute you to previous state when the previous state was 'edit'. So I made it default to data view instead.

willdoran:
Passes QA for me.

Angamanga:
Passes QA

PR #931, opened 2017-11-17
Removing onEnter block to keep user filters as they navigate the app

rowasc:
We cannot do resetDefaults and setMode here because the user needs to be able to move around and keep their filters.

- Remove onEnter block with setMode/reset filters

Testing checklist:
- [ ] Add filters in data or map view
-  [ ] change to the "opposite " view
- [ ] verify your filters travel across

Test with saved search :
- [ ]  Select a saved search in data mode
- [ ] Change to map mode and verifiy that you see the mode context bar with update search button there. 

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/posts-routes.js
+++ b/app/main/posts/posts-routes.js
@@ -121,10 +121,6 @@ function (
             name: 'posts.map.all',
             views: {
                 'mode-context': 'modeContext'
-            },
-            onEnter: function ($state, $transition$, PostFilters, savedSearch) {
-                PostFilters.setMode('all');
-                PostFilters.resetDefaults();
             }
         }
     )


app/main/posts/posts-routes.js
```diff
@@ -121,10 +121,6 @@ function (
             name: 'posts.map.all',
             views: {
                 'mode-context': 'modeContext'
-            },
-            onEnter: function ($state, $transition$, PostFilters, savedSearch) {
-                PostFilters.setMode('all');
```
rjmackay: I suspect setMode still belongs here. In fact from memory that has some logic that is mode changes it resets the filters right?
But resetDefaults should go.

PR #933, opened 2017-11-20
3745 saved searches view hierarchy

rowasc:
Checklist for testing ->  it's on testpad right now
Two known bugs: 
- UI: Buttons need style fixes (alignment of save/update search button)
- In the saved search list, when you clear filters, the list is not updated to the default value (but the filters are cleared). 

[x] I ran my own checklist.  

Test report https://ushahidi.ontestpad.com/script/2/report
TLDR of Test report:  All but 1 test passed, and the one that didn't is not related to saved searches, it's related  to https://github.com/ushahidi/platform/issues/2298 (it works with other sources that are not SMS, for instance). 

---
This change is [Reviewable](https://reviewable.io/reviews/ushahidi/platform-client/933)


app/main/posts/posts-routes.js
```diff
@@ -86,12 +86,24 @@ function (
...
+     * @uirouter-refactor
+     * we need to  be able to set mode-context as collectionModeContext or as savedSearchModeContext
+     * without the route since the user would land in /views/map after the redirect
+     */
     .state(
         {
             url: '^/savedsearches/:savedSearchId/data',
             name: 'posts.data.savedsearch',
             onEnter: function ($state, $transition$, PostFilters, savedSearch) {
+                PostFilters.setMode('savedsearch', savedSearch);
```
rowasc: Users need to be able to go to their bookmarked saved searches and see them but we won't BE in the url anymore since it's not supported in general (so we shouldn't have a special "old" URL , but a redirect would be desirable, IMO). The problem is that we don't have a good way that I know of to set the mode context named view based on params. :/ not sure how to go about it.

app/main/posts/views/post-filters.service.js
```diff
@@ -250,7 +262,7 @@ function PostFiltersService(_, FormEndpoint, TagEndpoint, $q) {
         return !_.isEmpty(getActiveFilters(filters));
     }
 
-    function setMode(newMode, id) {
+    function setMode(newMode, obj) {
```
rowasc: setting the obj instead of the ID makes sense to avoid unnecessary calls to the backend (nothing wrong with calling it, but if we have the object, it's alright to set it here IMO and gain the access to filters directly.)

app/main/posts/views/post-filters.service.js
```diff
@@ -259,15 +271,19 @@ function PostFiltersService(_, FormEndpoint, TagEndpoint, $q) {
...
         return filterMode;
     }
 
     function getModeId() {
-        return entityId;
+        return entity ? entity.id : null;
+    }
+
+    function getModeEntity() {
+        return entity;
```
rowasc: literally works as getModeId worked, only that it sends the object instead

app/settings/roles/editor.directive.js
```diff
@@ -18,6 +18,7 @@ function (
             $scope.save = $translate.instant('app.save');
             $scope.saving = $translate.instant('app.saving');
             $scope.saving_role = false;
+            // @uirouter-refactor this is very wrong.
```
rowasc: self note because I noticed this old code from last week that is 100% wrong. We are not guaranteed a $resolve everywhere so we should get it from $state if possible

app/main/posts/posts-routes.js
```diff
@@ -44,9 +44,9 @@ function (
             name: 'posts.savedsearchRedirector',
             onEnter: function ($state, $transition$, PostFilters, savedSearch) {
                 if (savedSearch.view === 'data' || savedSearch.view === 'list') {
-                    $state.go('posts.data.savedsearch', {savedSearchId: savedSearch.id, view: 'data'});
+                    $state.go('posts.data.all', {savedSearchId: savedSearch.id, view: 'data'});
```
rowasc: just avoiding a second redirect.

app/main/posts/views/filters/active-search-filters.directive.js
```diff
@@ -31,12 +33,45 @@ function ActiveSearchFilters($translate, $filter, PostFilters, _, FilterTransfor
...
+             *  - If there is a saved search:
+             *  - - set it in the scope : this is what we will use to display the saved search and its filters in the UI
+             *  - - get clean version (no defaults) of the saved search filters
+             *  - - get a clean activeFilters array that does not include the saved search filters. $scope.activeFilters: this
+             *  is the array we will be using to show the "extra" filters the user can set AFTER they selected  a saved search
+             *  (keep in mind that when a saved search is selected all filters are erased in favor of the saved search)
+             *  - - - Diffing rules: value from activeFilters takes priority over value from search. this is because
+             *  the user will always have selected the value after they select a saved search,meaning they want to change it.
+             *  - - - Diffing rules: if the value is equal, just ignore it/send empty value.
+             *  - - - Diffing rules: if the value is a number or string, return Value because it is a priority (the one from filters not from saved search)
```
rowasc: everything is an array and can be compared with $difference in a sane way, I think??

app/main/posts/views/filters/active-search-filters.directive.js
```diff
@@ -31,12 +33,45 @@ function ActiveSearchFilters($translate, $filter, PostFilters, _, FilterTransfor
...
+             */
+            activeFilters = _.mapObject(activeFilters, makeArray);
+            if ($scope.savedSearch) {
+                // get clean version (no defaults) of the saved search filters
+                $scope.savedSearch.filter = _.mapObject(PostFilters.getCleanActiveFilters($scope.savedSearch.filter), makeArray);
+                $scope.activeFilters = _.mapObject(activeFilters, function (value, key) {
+                    return _.difference(value, $scope.savedSearch.filter[key]);
+                });
+            } else {
+                $scope.activeFilters = activeFilters;
```
rowasc: I set it here instead of pre-setting it just because we watch active filters, no need to change it in line 65 and then again in the saved search if . But might be one of those scenarios where we prefer to do it on line 65 because I admit it'd be more readable/easier to follow. Not sure

app/main/posts/views/filters/active-search-filters.directive.js
```diff
@@ -46,11 +81,36 @@ function ActiveSearchFilters($translate, $filter, PostFilters, _, FilterTransfor
...
+                savedSearch.filter = PostFilters.clearFilterFromArray(filterKey, value, savedSearch.filter);
+                PostFilters.clearFilter(filterKey, value);
+                PostFilters.setFilter('saved_search', savedSearch);
+                $scope.savedSearch = savedSearch;
+            } else {
+                PostFilters.clearFilter(filterKey, value);
+            }
         }
+        function removeSavedSearch(savedSearch, $event) {
+            $event.preventDefault();
```
rowasc: this whole function is a bit silly . I'll look into simplifying this removeFilter/removeSavedSearch feature if possible

app/main/posts/views/filters/filters-dropdown.directive.js
```diff
@@ -29,12 +29,9 @@ function FiltersDropdown(PostFilters, ModalService, $rootScope, _, $location, Sa
...
         function setSavedSearchUpdateStatus() {
-            var savedSearchId = PostFilters.getModeId();
-            if (savedSearchId) {
-                SavedSearchEndpoint.get({id: savedSearchId}, function (savedSearch) {
-                    $scope.canUpdateSavedSearch = _.contains(savedSearch.allowed_privileges, 'update');
-                });
-            }
+            var savedSearch = PostFilters.getModeEntity();
+
+            $scope.canUpdateSavedSearch = savedSearch && _.contains(savedSearch.allowed_privileges, 'update');
```
rowasc: IDK if this is worth it. Maybe we would prefer to go and get it , but since it's already available , not sure

app/main/posts/views/filters/filter-posts.html
```diff
@@ -26,5 +26,5 @@
         </a>
     </div>
 
-    <filters-dropdown uib-dropdown-menu class="dropdown-menu" dropdown-status="status" filters-var="filters" ng-model="filters" cancel="cancel" apply-filters="applyFilters"></filters-dropdown>
+    <filters-dropdown view="view" uib-dropdown-menu class="dropdown-menu" dropdown-status="status" filters-var="filters" ng-model="filters" cancel="cancel" apply-filters="applyFilters"></filters-dropdown>
```
rowasc: I can't figure out how to access view name (if it's even possible) in the filters-dropdown so I send it from the filter-posts so I don't have to mess with parents and assume things

app/main/posts/posts-routes.js
```diff
@@ -86,22 +86,34 @@ function (
...
+     * @uirouter-refactor
+     * we need to  be able to set mode-context as collectionModeContext or as savedSearchModeContext
+     * without the route since the user would land in /views/map after the redirect
+     */
     .state(
         {
             url: '^/savedsearches/:savedSearchId/data',
             name: 'posts.data.savedsearch',
-            onEnter: ['PostFilters', 'savedSearch', function (PostFilters, savedSearch) {
+            onEnter: function ($state, $transition$, PostFilters, savedSearch) {
```
rjmackay: You need to declare dependencies here ie `['PostFilters', 'savedSearch', function (PostFilters, savedSearch) {`
otherwise this breaks in production builds

app/main/posts/views/filters/active-search-filters.directive.js
```diff
@@ -46,11 +83,36 @@ function ActiveSearchFilters($translate, $filter, PostFilters, _, FilterTransfor
...
 
-        function removeFilter(filterKey, value, $event) {
+        function removeFilter(filterKey, value, savedSearch, $event) {
             $event.preventDefault();
             $event.stopPropagation();
-            PostFilters.clearFilter(filterKey, value);
+            if (savedSearch) {
+                savedSearch.filter = PostFilters.clearFilterFromArray(filterKey, value, savedSearch.filter);
+                PostFilters.clearFilter(filterKey, value);
+                PostFilters.setFilter('saved_search', savedSearch);
```
rjmackay: why do we need to set saved_search as a filter again here?

rowasc: We don't 👍

app/main/posts/views/filters/active-search-filters.directive.js
```diff
@@ -31,12 +35,55 @@ function ActiveSearchFilters($translate, $filter, PostFilters, _, FilterTransfor
             return value;
         }
 
-        function handleFiltersUpdate(filters) {
+        function handleFiltersUpdate(filters, oldValue) {
+            /**
+             * TODO: to handle removal correctly we need to make sure we take oldValue into account,
```
rowasc: @rjmackay added a comment here while I handle this. Removal is a mess. Working on it.

rowasc: still thinking if this makes sense but I think it does.

app/main/posts/views/filters/active-search-filters.directive.js
```diff
@@ -31,12 +35,65 @@ function ActiveSearchFilters($translate, $filter, PostFilters, _, FilterTransfor
...
         }
 
-        function handleFiltersUpdate(filters) {
+        /**
+         * Looks for keys that are NOT present in currentFilters but that are in the savedSearch filters (which means they are removed)
+         * and removes them from the saved search filters array.
+         * Does not handle removal where the key exists but the values are an array and some are missing. Need to fix that.
+         **/
+        function cleanDeprecatedValuesFromSavedSearch(currentFilters, savedSearch) {
+            //find filters in currentFilters that are NOT in savedSearch.filters
```
rowasc: strong case for a decent diffing library for sure. :/

rowasc: also I'm changing this name before it lands , thinking a bit more about the whole thing

rowasc:
@rjmackay can you take a new look at this? I'll ask for a new CR during the day from Will / Carolyn after I squash the bugs and CR feedback. 

# known bugs: not 100% sure
-  the buttons are shown in the wrong place and need to check conditions where they are supposed to be shown (need sync up with Will/Carolyn probably)
- I'm sure there are tons of quirks which I'll work on in the morning. Still need to check the mode context bar filters which we discussed on the sync call.

app/settings/webhooks/editor.directive.js
```diff
@@ -22,6 +22,7 @@ function (
             $scope.save = $translate.instant('app.save');
             $scope.saving = $translate.instant('app.saving');
             $scope.processing = false;
+            // @uirouter-refactor this is very wrong.
```
rowasc: need to switch scope.$resolve.$transition$.params().id

app/main/posts/posts-routes.js
rjmackay: Could you potentially use a component provider to switch mode context component based on some param?`https://ui-router.github.io/ng1/docs/latest/interfaces/ng1.ng1viewdeclaration.html#componentprovider`
Or course that won't update when the filters change, only on state change

ie:
```
             views: {
                  'mode-context': {
                        'componentProvider' : () {
                                 return 'someModeContext';
                         }
                  }
              }
```

app/main/posts/views/filters/active-search-filters.directive.js
```diff
@@ -31,12 +35,70 @@ function ActiveSearchFilters($translate, $filter, PostFilters, _, FilterTransfor
...
+            if ($scope.savedSearch) {
+                /**
+                 * to handle removal correctly we need to make sure we take currentFilters (which has up to date info) into account,
+                 * because that is where our savedsearch filters will stop being represented when we remove them.
+                 * If there's a key in our current filters that is in the saved search but is not in the active filters at this point,
+                 * it is because it was removed (since before saved search gets assigned, they are all assigned to the filters)
+                 * that means we have to remove it from the saved search.
+                 **/
+                $scope.savedSearch.filter = cleanDeprecatedValuesFromSavedSearch(activeFilters, _.mapObject(PostFilters.getCleanActiveFilters($scope.savedSearch.filter), makeArray));
+                $scope.activeFilters = _.mapObject(activeFilters, function (value, key) {
```
rjmackay: Just trying to figure out if I followed whats going on here. Would it be accurate to describe this as splitting the filters? ie. splitting into "saved search filters" and "other filters" ...
Also cleaning out 'removed but not yet saved' filters from the savedsearch... but I have no idea what to call that.

app/main/posts/views/filters/filters-dropdown.directive.js
```diff
@@ -44,7 +42,8 @@ function FiltersDropdown(PostFilters, ModalService, $rootScope, _, $location, Sa
         $scope.clearFilters = function () {
             if (PostFilters.getMode() === 'savedsearch' && PostFilters.getModeId()) {
                 PostFilters.setMode('all');
-                var viewParam = $scope.$resolve.$transition$.params().view ? $scope.$resolve.$transition$.params().view : 'data';
+                // @uirouter-refactor this is very wrong.
+                var viewParam = $scope.view ? $scope.view : 'data';
```
rjmackay: this is check map vs data view?
Can you inject $state and just check `$state.current.name`? or $state.params?

app/main/posts/views/filters/active-search-filters.directive.js
rowasc: right, yea I think that's accurate. It's clearing out filters that don't have to be shown in the "normal" active filters and then also clearing out filters that were removed from the saved search filters. Does that make sense?

app/main/posts/views/filters/filters-dropdown.directive.js
rowasc: :( I can't. It fails when you try to inject it in a directive that has a link and no controller

rowasc: it used to be possible to just call $stateParams, which is still a ridiculous thing to do because it's deprecated. So not sure

crcommons:
@rowasc Current bugs I found:
- When you create or select a SS and then try to create another one (even with the filters cleared), the previous SS name persists in the modal; 
- When you clear filters, the SS stays selected in the dropdown, which I think stems from the same problem as the first issue;
- When you create a new SS, and save it, it doesn't apply those filters from the SS when it reloads the page. Is this correct?

rowasc:
argh was not aware of those. Thanks ! Will look into them

crcommons:
@rowasc Also, when you manually 'x' out the entire saved search, the SS is still selected in the dropdown as well. Also, when you manually 'x' out the search criteria, those changes are not reflected in the checkboxes

rowasc:
>when you manually 'x' out the search criteria, those changes are not reflected in the checkboxes

ok that's really weird. Probably broke it today.

rowasc:
>Also, when you manually 'x' out the search criteria, those changes are not reflected in the checkboxes

:( Can't make this happen locally not sure what's going on

app/main/posts/views/post-filters.service.js
```diff
@@ -30,10 +30,28 @@ function PostFiltersService(_, FormEndpoint, TagEndpoint, $q) {
...
         qEnabled: false
     };
 
+
+    /**
+     * Looks for keys that are NOT present in currentFilters but that are in the savedSearch filters (which means they are removed)
+     * and removes them from the saved search filters array.
+     * Does not handle removal where the key exists but the values are an array and some are missing. Need to fix that.
+     **/
+    function cleanDeprecatedValuesFromSavedSearch(currentFilters, savedSearch) {
```
rjmackay: maybe just cleanRemovedValuesFromSavedSearch? Deprecated seems like the wrong term, but I'm not sure whats better

rowasc: yea that's a little better. Still weird.. but better. :) thanks

rjmackay:
## Test results:

Issue 1.
- Apply saved search which has all status filters selected
- Remove a published filter
- Reapply published filter
It should return to the original state, and not show the Update search button
Actually shows changed status filters outside the SS heirarchy and show 'update search'

Issue 2.
- Saved search drop down does not reset when clearing filters

Issue 3
Weird/unpredictable behaviour because selecting all surveys/categories etc.
Because we hide these from active when they match the default, it can be confusing.

    - Select a SS with only a few surveys selected
    - Add the missing surveys till all are selected
    - All survey filters disappear, not obvious which ones are part of the search or not.

Issues 4: 
- Select a saved search
- Remove a filter with the 'X' button
Filter is removed, but the change is applied immediately. 
Expected it to wait till I clicked 'Apply filters'

Issue 5 (maybe out of scope?)
- "Search all for 'test'" button appears at the top of the filters dropdown

Issue 6 (maybe out of scope?)
- Clicking inside text search box with filters open, closes the filers drop down
It'd probably make more sense to just have a single drop down.
ie. open the entire search dropdown when clicking in the text search box

rowasc:
Thanks Robbie ! Looking into this.

rowasc:
@rjmackay 
`Select a saved search
Remove a filter with the 'X' button
Filter is removed, but the change is applied immediately.
Expected it to wait till I clicked 'Apply filters'`

Is this in map mode? It's not happening to me in data mode, where that's the expected behavior (in map mode, it's supposed to change as you change filters, in data mode, it's not)

rowasc:
>@rjmackay 
`Select a saved search
Remove a filter with the 'X' button
Filter is removed, but the change is applied immediately.
Expected it to wait till I clicked 'Apply filters'`

>Is this in map mode? It's not happening to me in data mode, where that's the expected behavior (in map mode, it's supposed to change as you change filters, in data mode, it's not)

uh ok interesting, I just saw it happen but I have no idea how.Debugging

rjmackay:
>  Select a saved search Remove a filter with the 'X' button Filter is removed, but the change is applied immediately. E�xpected it to wait till I clicked 'Apply filters'
> 
> Is this in map mode?

Oops yea totally correct. It works fine. I actually thought as I was writing it "wait, was I just in map mode?" but didn't retest.

app/main/posts/views/filters/active-search-filters.directive.js
```diff
@@ -24,19 +35,80 @@ function ActiveSearchFilters($translate, $filter, PostFilters, _, FilterTransfor
...
+                $scope.savedSearch.filter = PostFilters.getUIActiveFilters(PostFilters.addIfCurrentObjectMatchesOriginal(PostFilters.getUIActiveFilters($scope.savedSearch.filter), PostFilters.getUIActiveFilters(originalSavedSearch.filter), PostFilters.getUIActiveFilters($scope.uiFilters)));
+                //!_.isEqual($scope.savedSearch.filter, PostFilters.getUIActiveFilters(originalSavedSearch.filter)
+                var savedSearchFiltersChanged = _.filter(originalSavedSearch.filter, function (obj, key) {
+                    if (!_.isArray(obj)) {
+                        return $scope.savedSearch.filter[key] !== obj;
+                    } else {
+                        return _.difference($scope.savedSearch.filter[key], obj).length !== 0 || _.difference(obj, $scope.savedSearch.filter[key]).length !== 0 ;
+                    }
+                }).length > 0 ;
+                var filtersDifferentToSavedSearch = !_.isEqual($scope.savedSearch.filter, filters);
```
rowasc: move this to a function

app/main/posts/views/filters/active-search-filters.directive.js
```diff
@@ -46,10 +118,41 @@ function ActiveSearchFilters($translate, $filter, PostFilters, _, FilterTransfor
             return value;
         }
 
-        function removeFilter(filterKey, value, $event) {
+        function removeFilter(filterKey, value, savedSearch, $event) {
+            $event.preventDefault();
+            $event.stopPropagation();
+            if (savedSearch) {
+                // commented - if we do this here we can' compare for removal and know that it's updated
+                // savedSearch.filter = PostFilters.clearFilterFromArray(filterKey, value, savedSearch.filter);
```
rowasc: remove commented out code

app/main/posts/views/filters/active-search-filters.directive.js
```diff
@@ -46,10 +118,41 @@ function ActiveSearchFilters($translate, $filter, PostFilters, _, FilterTransfor
...
+            } else {
+                PostFilters.clearFilter(filterKey, value);
+            }
+        }
+
+        function removeSavedSearch(savedSearch, $event) {
             $event.preventDefault();
             $event.stopPropagation();
-            PostFilters.clearFilter(filterKey, value);
+            originalSavedSearch = null;
```
rowasc: add comments

app/main/posts/views/filters/active-search-filters.html
```diff
@@ -1,8 +1,51 @@
-<div ng-repeat="(key, values) in activeFilters" class='inline-filters'>
-    <span class="bug" ng-repeat="value in values track by $index" >
-        <svg class="iconic" ng-click="removeFilter(key, value, $event)">
+<div class="form-field saved-search" ng-show="savedSearch">
+    <span class="bug">
+        <svg class="iconic" ng-click="removeSavedSearch(savedSearch, $event)">
             <use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="/img/iconic-sprite.svg#circle-x"></use>
         </svg>
-        <span><em><translate translate="global_filter.filter_tabs.{{key}}"></translate>:</em> {{ transformFilterValue(value, key) }}</span>
+        <span><em>Saved search: </em>{{savedSearch.name}}</span>
```
rowasc: needs translation

app/main/posts/views/post-filters.service.js
```diff
@@ -18,19 +18,73 @@ function PostFiltersService(_, FormEndpoint, TagEndpoint, $q) {
...
+     *  uiFiltersCurrent = {status: ['draft', 'review', 'archive']}
+     *  Original has 3 filters, then 1 was removed, so current has 2. uiFiltersCurrent has 3 again because the user re-added it.
+     *  We need to compare  original and uiFIltersCurrent.
+     *  If `uiFiltersCurrent` has values that are present in  `original`,
+     *  we need to add them to `current` (*some other function will remove them from uiFiltersCurrent too)
+     *
+     * @param current
+     * @param original
+     */
+    function addIfCurrentObjectMatchesOriginal(currentSS, originalSS, uiObj) {
```
rowasc: fix this signature for something less terrible

app/main/posts/views/post-filters.service.js
```diff
@@ -18,19 +18,73 @@ function PostFiltersService(_, FormEndpoint, TagEndpoint, $q) {
...
+        });
+        return ret;
+    }
+
+    /**
+     * Looks for keys that are NOT present in currentFilters but that are in the savedSearch filters (which means they are removed)
+     * and removes them from the saved search filters array.
+     * Does not handle removal where the key exists but the values are an array and some are missing. Need to fix that.
+     **/
+    function cleanRemovedValuesFromObject(currentFilters, oldFilters) {
```
rowasc: I think this name is slightly better than the previous name so it stays for the moment

rowasc:
` Issue 3

Weird/unpredictable behaviour because selecting all surveys/categories etc.
Because we hide these from active when they match the default, it can be confusing.

Select a SS with only a few surveys selected
Add the missing surveys till all are selected
All survey filters disappear, not obvious which ones are part of the search or not.`


@rjmackay Maybe this is something we should look into with @justinscherer ? I'm not sure what the alternative is since it kind of feels like that's the right behavior but as a user it confuses me a lot.

rowasc:
` Issue 5 +  Issue 6` I feel we could log into separate  tickets probably ?

rowasc:
# Issue found: 
Array filters (status, categories, surveys) are always added to the saved search hierarchy if we select a saved search. 

- Select a saved search with  some surveys
-  Add another survey category 
- Expected behavior: It should be added outside of the saved search as other new filters, but it's added to the saved search categories instead.

app/main/posts/views/post-filters.service.js
rowasc: ### Issue found here: 
Array filters (status, categories, surveys) are always added to the saved search hierarchy if we select a saved search. 

- Select a saved search with  some surveys
-  Add another survey category 
- Expected behavior: It should be added outside of the saved search as other new filters, but it's added to the saved search categories instead.

rowasc:
# Bug: 
When I deselect a category from the bug icons, it does not reflect it in the checkboxes (it happens with surveys too) . 
This might be fixed in https://github.com/ushahidi/platform-client/pull/935/files

rowasc:
Two bugs :
https://ushahidi.ontestpad.com/script/2#313// Related info: when I delete a filter in the saved search the active filters update handler runs TWICE . The first time it's correct, second time obviously not. 

https://ushahidi.ontestpad.com/script/2#321//

app/main/posts/views/filters/filters-dropdown.directive.js
```diff
@@ -1,77 +1,75 @@
...
-                PostFilters.setMode('all');
-                var viewParam = $scope.$resolve.$transition$.params().view ? $scope.$resolve.$transition$.params().view : 'data';
-                $location.url('/views/' + viewParam);
-            }
-            $scope.filtersVar = PostFilters.clearFilters();
-            $scope.dropdownStatus.isopen = !$scope.dropdownStatus.isopen;
-            PostFilters.reactiveFilters = true;
-            $scope.canUpdateSavedSearch = false;
+FiltersDropdownController.$inject = ['$scope', '$state', 'PostFilters', 'ModalService', '$rootScope', '_', '$location', 'SavedSearchEndpoint'];
+function FiltersDropdownController($scope, $state, PostFilters, ModalService, $rootScope, _, $location, SavedSearchEndpoint) {
```
rjmackay: @rowasc so I converted this to a controller simply so I could inject $state here. 
Unfortunately the indent change makes the diff huge :/

app/main/posts/views/filters/filters-dropdown.directive.js
```diff
@@ -1,77 +1,75 @@
...
+
+    $scope.applyFiltersLocked = function () {
+        PostFilters.reactiveFilters = true;
+        $scope.dropdownStatus.isopen = !$scope.dropdownStatus.isopen;
+    };
+    $scope.clearFilters = function () {
+        if (PostFilters.getMode() === 'savedsearch' && PostFilters.getModeId()) {
+            PostFilters.setMode('all');
+
+            if ($state.$current.includes['posts.map']) {
```
rjmackay: @rowasc so I think this is better than the previous `view` param and setting $location? What do you think?

rjmackay:
Reviewed 32 of 32 files at r1.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[app/main/posts/views/post-filters.service.js, line 48 at r1](https://reviewable.io:443/reviews/ushahidi/platform-client/933#-Kzfr5pbgSrEkyphLQFl:-Kzfr5pbgSrEkyphLQFm:b-rsv6ml) ([raw file](https://github.com/ushahidi/platform-client/blob/eb8f9d308991390d1262a65e42bf2bade9235573/app/main/posts/views/post-filters.service.js#L48)):*
> ```JavaScript
>      * @param original
>      */
>     function addIfCurrentObjectMatchesOriginal(currentSavedSearch, originalSavedSearch, currentFilters) {
> ```

This is... confusing. But it appears to work

---

*[app/main/posts/views/filters/active-search-filters.directive.js, line 24 at r1](https://reviewable.io:443/reviews/ushahidi/platform-client/933#-KzfrC4MuTM4Sn0ARPD3:-KzfrC4MuTM4Sn0ARPD4:b-rxucp7) ([raw file](https://github.com/ushahidi/platform-client/blob/eb8f9d308991390d1262a65e42bf2bade9235573/app/main/posts/views/filters/active-search-filters.directive.js#L24)):*
> ```JavaScript
> 
>         activate();
>         $scope.$on('savedSearch:update', function () {
> ```

This should really happen in activate

---

*[app/main/posts/views/filters/active-search-filters.directive.js, line 59 at r1](https://reviewable.io:443/reviews/ushahidi/platform-client/933#-Kzfrby_P1fXjiECE_6p:-Kzfrby_P1fXjiECE_6q:bfegrve) ([raw file](https://github.com/ushahidi/platform-client/blob/eb8f9d308991390d1262a65e42bf2bade9235573/app/main/posts/views/filters/active-search-filters.directive.js#L59)):*
> ```JavaScript
>             }
>             /**
>              * This handles the requirement to have saved search filters displayed in a different way
> ```

As much as I like the detailed descriptions. I think it would be easy to pass if we aim for inline comments closer to where things are actually happening.
Leave it as-is right now though

---

*[app/main/posts/views/filters/active-search-filters.directive.js, line 88 at r1](https://reviewable.io:443/reviews/ushahidi/platform-client/933#-KzfrY59z5rmABeuodKs:-KzfrY59z5rmABeuodKt:b-3wswuf) ([raw file](https://github.com/ushahidi/platform-client/blob/eb8f9d308991390d1262a65e42bf2bade9235573/app/main/posts/views/filters/active-search-filters.directive.js#L88)):*
> ```JavaScript
>                  * Add back in savedSearch.filter if an originally saved search filter is removed+added back
>                  */
>                 $scope.savedSearch.filter = PostFilters.getUIActiveFilters(PostFilters.addIfCurrentObjectMatchesOriginal(PostFilters.getUIActiveFilters($scope.savedSearch.filter), PostFilters.getUIActiveFilters(originalSavedSearch.filter), PostFilters.getUIActiveFilters($scope.uiFilters)));
> ```

Please split stuff like this over many lines. Its super hard to read right now.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/ushahidi/platform-client/933)*

rjmackay:
@rowasc I believe I've fixed one of those issues with ba096e3
> Switching from map to data mode with a saved search selected, should show it as selected in filters.

rjmackay:
> https://ushahidi.ontestpad.com/script/2#313// Related info: when I delete a filter in the saved search while having other filters outside of it, the active filters update handler runs TWICE . The first time it's correct, second time obviously not.

I'm super confused. I saw this happening... then went and had dinner... now I can't reproduce it. 
Maybe I fixed it accidentally? I don't know how.

rjmackay:
Bugs:
1. [ ] Save search button appears even when not logged in
2. [ ] Select a saved search with all statuses. Close the 'archived' bug. All status filter disappear (because we're now using the default status filter). But 'archived' is checkbox is still checked.
3. [ ] https://ushahidi.ontestpad.com/script/2#220/9/

rowasc:
> https://ushahidi.ontestpad.com/script/2#313// Related info: when I delete a filter in the saved search while having other filters outside of it, the active filters update handler runs TWICE . The first time it's correct, second time obviously not.

    > I'm super confused. I saw this happening... then went and had dinner... now I can't reproduce it.
       Maybe I fixed it accidentally? I don't know how.
       Edit: Pretty sure ba096e3 fixed that too. This stuff is all way to closely coupled its scary.

OK  yes that's fixed now!!! . Thanks!

app/main/posts/views/filters/active-search-filters.directive.js
```diff
@@ -24,19 +35,79 @@ function ActiveSearchFilters($translate, $filter, PostFilters, _, FilterTransfor
...
+                 * because that is where our savedsearch filters will stop being represented when we remove them.
+                 * If there's a key in our current filters that is in the saved search but is not in the active filters at this point,
+                 * it is because it was removed (since before saved search gets assigned, they are all assigned to the filters)
+                 * that means we have to remove it from the saved search.
+                 **/
+                $scope.savedSearch.filter = PostFilters.cleanRemovedValuesFromObject(filters, PostFilters.getUIActiveFilters($scope.savedSearch.filter));
+                /**
+                 * Add back in savedSearch.filter if an originally saved search filter is removed+added back
+                 */
+                $scope.savedSearch.filter = PostFilters.getUIActiveFilters(PostFilters.addIfCurrentObjectMatchesOriginal(PostFilters.getUIActiveFilters($scope.savedSearch.filter), PostFilters.getUIActiveFilters(originalSavedSearch.filter), PostFilters.getUIActiveFilters($scope.uiFilters)));
```
rowasc: FYI I moved this to a service and made it more sane too . Pushing today after testing everything again

app/main/posts/views/filters/active-search-filters.directive.js
```diff
@@ -24,19 +35,79 @@ function ActiveSearchFilters($translate, $filter, PostFilters, _, FilterTransfor
...
+                    if (!_.isArray(obj)) {
+                        return $scope.savedSearch.filter[key] !== obj;
+                    } else {
+                        return _.difference($scope.savedSearch.filter[key], obj).length !== 0 || _.difference(obj, $scope.savedSearch.filter[key]).length !== 0 ;
+                    }
+                }).length > 0 ;
+                var filtersDifferentToSavedSearch = !_.isEqual($scope.savedSearch.filter, filters);
+                $scope.userCanUpdateSavedSearch = _.contains($scope.savedSearch.allowed_privileges, 'update') && (savedSearchFiltersChanged || filtersDifferentToSavedSearch);
+                // remove values that are in the saved search frmo the uifilters.
+                _.each($scope.uiFilters, function (value, key) {
```
rowasc: FYI I moved this to a service and made it more sane . Pushing today after testing everything again

rowasc:
> Extra filters are being cleared when switching between data and map view https://ushahidi.ontestpad.com/script/2#221/10/

FYI, Solved this locally, pushing today after I clean up and test everything.

rowasc:
> Select a saved search with all statuses. Close the 'archived' bug. All status filter disappear (because we're now using the default status filter). But 'archived' is checkbox is still checked.

note:  this happens also with normal filters outside of a saved search ;/ checking

rowasc:
sent copy paste fix https://github.com/ushahidi/platform-client/pull/933/commits/b52e4201f40410b7e7a62f99b11f570745aeadc6 from Robbie's code for 
`Select a saved search with all statuses. Close the 'archived' bug. All status filter disappear (because we're now using the default status filter). But 'archived' is checkbox is still checked.`

rowasc:
Test report https://ushahidi.ontestpad.com/script/2/report

All but 1 test passed, and the one that didn't is not related to saved searches, just to https://github.com/ushahidi/platform/issues/2298 (it works with other sources that are not SMS, for instance).

rowasc:
There is a styling issue that I'm going to check after I re-run my checklist against some local changes. 
The saved search button is not correctly styled/there's some incorrect margin there and it looks squashed next to the filter bug icons

rowasc:
Re-run checklist against 07a95a6

rowasc:
@rjmackay I looked into styles for the UPDATE and SAVE button,  and apparently SAVE and UDPATE should still be in filters-dropdown =/ (or the html structure for active filters with buttons should all live inside active filters which is way way easier right now but it's more ugly ) .  

Not sure how to move the buttons without doing some data bindings of userCanUpdateSavedSearch / showSaveSavedSearchButton into active filters from filters-dropdown to know when it’s updateable/save-able .  
Thoughts?

I'm doing the ugly bindings later today . But I probably will want to do it in a different PR going into this one to avoid messing things up here.

rowasc:
Cherry picked @rjmackay  https://github.com/ushahidi/platform-client/pull/933/commits/52353f518cbd2e32934e9493f4b06777aa9621f7 into this so we have the correct styles. Thanks !

rowasc:
#Testing status
https://github.com/ushahidi/platform-client/pull/933/commits/0b8c9e795ea15914601d4e2cb3a12ca8b6cd0c8e

https://github.com/ushahidi/platform-client/pull/933/commits/52353f518cbd2e32934e9493f4b06777aa9621f7

and 
https://github.com/ushahidi/platform-client/pull/933/commits/47a4dd586367a19e34ac320e1656756d04538e66

- I have not gone through the full checklist for those 3 small commits . 
- Everything else was tested  in (see https://ushahidi.ontestpad.com/script/2#367/14/ + https://ushahidi.ontestpad.com/script/2#367/13/)
- I'll run again the test checklist again after Code review since those are small changes that should not break anything

app/main/posts/views/filters/filters-dropdown.directive.js
rowasc: yea, that's way better .

app/main/posts/views/post-filters.service.js
```diff
@@ -18,19 +18,96 @@ function PostFiltersService(_, FormEndpoint, TagEndpoint, $q) {
...
+    /**
+     *  Original has 3 filters, then 1 was removed, so current has 2. uiFiltersCurrent has 3 again because the user re-added it.
+     *  We need to compare  original and uiFIltersCurrent.
+     *  If `uiFiltersCurrent` has values that are present in  `original`,
+     *  we need to add them to `current` (*some other function will remove them from uiFiltersCurrent too)
+     *
+     * @param current
+     * @param original
+     */
+    function addIfCurrentObjectMatchesOriginal(currentSavedSearch, originalSavedSearch, currentFilters) {
```
rowasc: I can't figure out a name. Still thinking about this. It works and its tested but it feels like this should be way more generic

app/main/posts/views/filters/active-search-filters.directive.js
```diff
@@ -24,19 +35,84 @@ function ActiveSearchFilters($translate, $filter, PostFilters, _, FilterTransfor
...
+             *  the user will always have selected the value after they select a saved search,meaning they want to change it.
+             *  - - - Diffing rules: if the value is equal, just ignore it/send empty value.
+             *  - - - Diffing rules: return the _difference if the value is not equal, because we will want to show for instance:
+             *  saved search: tag id 1 + filters tag id 2 (so it's not just ignoring the arrays)
+             */
+            $scope.userCanUpdateSavedSearch = false;
+            $scope.uiFilters = currentFilters;
+
+            if ($scope.savedSearch) {
+                var originalSavedSearchFilters = PostFilters.getUIActiveFilters(originalSavedSearch.filter);
```
rowasc: we should be calling getUIActiveFilters way less and this whole thing should be more simple... But I need way more test coverage to be confident changing this right now TBH .  :/ It's kind of scary .

app/main/posts/views/filters/active-search-filters.directive.js
```diff
@@ -122,7 +127,7 @@ function ActiveSearchFilters($translate, $filter, PostFilters, _, FilterTransfor
             $event.stopPropagation();
             if (savedSearch) {
                 // commented - if we do this here we can' compare for removal and know that it's updated
```
rjmackay: Is this comment still relevant?

rowasc: yes , but misplaced. The commented out code should just be removed along with the comment. It was more of an internal note , no longer relevant .

rowasc:
@rjmackay updated to remove commented out code + comments since it will only cause confusion

rjmackay:
New test run: https://ushahidi.ontestpad.com/script/2#10/15/
- Pretty much all passed
- but I noticed the save and clear search button in the sidebar aren't consistent with the dropdown ones - save still shows up when it shouldn't. I suggest we just remove them TBH.
- Also hit a weird state with a old saved search filters showing after update. We must be keeping the searches in local state somewhere and not reloading :/

rjmackay:
> Also hit a weird state with a old saved search filters showing after update. We must be keeping the searches in local state somewhere and not reloading :/

I can't reproduce this right now. Ignore it until we can figure out a consisten way to reproduce

rjmackay:
The main outstanding issue on this is that it fails to handle collections correctly/at all. 
In places it assume that `getModeId` always returns a saved search, rather than sometimes a collection. And when a collection is active it doesn't show that in the filters at all.

@rowasc I'm wary that this could drag on though, and searching is kinda broken on develop right now anyway. I'd suggest you merge this and then branch to fix collections.

The most you should do on this branch is to fix instances where a collection is mistaken for a saved search. I started look at that on the `fixup-collections` branch

PR #936, opened 2017-11-23
Allow background of postcards to be tappable

kinstelli:
Per discussion with @justinscherer , everything but the dropdown ellipses should be clickable in order to select the post. The only complication is that clicking the dropdown toggle shouldn't select the post.

Moved `clickAction()` to the whole postcard body.
Stops propagation if the ellipses is clicked.

Testing checklist:
- [ ] Click anywhere in the `postcard-body` except the ellipses. That post should get selected.
- [ ] Click the ellipses on the selected card. This should still toggle the dropdown menu.
- [ ] Click the ellipses on an _unselected_ card. This should still toggle the dropdown menu, but not select the post.
 
Fixes ushahidi/platform#2273 .

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/common/post-actions.html
+++ b/app/main/posts/common/post-actions.html
@@ -1,5 +1,5 @@
 <div class="postcard-actions" dropdown >
-  <button class="button-gamma button-flat init" data-toggle="dropdown-menu" dropdown-toggle>
+  <button class="button-gamma button-flat init" data-toggle="dropdown-menu" dropdown-toggle ng-click="$event.stopPropagation()">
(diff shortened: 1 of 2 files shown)


rjmackay:
@kinstelli does the styling work for this too? ie. does the pointer change to be clickable for the entire card now?

kinstelli:
I believe the cursor was always switching to a pointer for the whole card, so that wasn't changed and ushahidi/platform#2236 should be a non-issue now.

rjmackay:
Works perfectly. Merging

PR #939, opened 2017-11-26
Remove double locks

rjmackay:
- Remove icons for status from post list and other post metadata uses
- Use 'document' icon for under review in post actions

Testing checklist:
- [x] When a post is locked it shows a lock icon in the list
- [x] Status icons are not shown in post detail
- [x] Status icons are not shown in post list
- [x] Status icons are not shown in post map popups
- [x] Status icons show in post actions dropdown but uses 'document' for under review.
- [x] Status filter uses 'document icon'
- [x] Editing post: Under review status uses 'document icon' 

Fixes ushahidi/platform#2197

Ping @ushahidi/platform


crcommons:
@rjmackay 
Bulk actions should also use the correct under review icon, but it still uses the lock.

PR #941, opened 2017-11-27
Fix collection handling after filter and search changes

rjmackay:
- [x] Switch mode context component based on current mode
- [x] Fix issues handling collections as if they were saved searches
- [x] Show a collection bug in filters when collection mode is active.

Testing checklist:
- [ ] TODO

- [x] I certify that I ran my checklist https://ushahidi.ontestpad.com/script/4#54//

Fixes ushahidi/platform#2299

Ping @ushahidi/platform


app/main/posts/views/filters/filters-dropdown.directive.js
```diff
@@ -40,9 +40,8 @@ function FiltersDropdownController($scope, $state, PostFilters, ModalService, $r
         $scope.dropdownStatus.isopen = !$scope.dropdownStatus.isopen;
     };
     $scope.clearFilters = function () {
-        if (PostFilters.getMode() === 'savedsearch' && PostFilters.getModeId()) {
+        if (_.contains(['savedsearch', 'collection'], PostFilters.getMode()) && PostFilters.getModeId()) {
```
rjmackay: Is the point of this check just to change the url?

rowasc: yes... and to do the mode reset to 'all' . It was checking against savedsearch only before that and not resetting the mode when you cleared the filters in a collection.

app/main/posts/collections/mode-context.html
```diff
@@ -112,7 +112,7 @@ <h1 class="mode-context-title">
         <!--// END IF //-->
         <div class="form-field">
             <!-- mode context bar is only available in the map so it should take you back there -->
-            <a ui-sref="posts.map.all" type="button" class="button-link">
+            <a ng-click="clearFilters()" type="button" class="button-link">
```
rowasc: FYI because I'm not 100% on what expected here . . but this will clear any filters that the user set while being on a saved search or collection, which is not what happens usually when they navigate away from data to map, map to data, etc. Is that OK?  -I expect it would be OK but, not sure-

app/main/posts/posts-routes.js
```diff
@@ -43,8 +47,11 @@ function (
     )
     .state(
         {
-            url: '^/savedsearches/:savedSearchId',
+            url: '^/savedsearches/{savedSearchId:int}',
```
rowasc: yay types

app/main/posts/posts-routes.js
```diff
@@ -235,13 +260,14 @@ function (
...
     )
-    .state(
-        {
-            name: 'postEdit',
-            url: '/posts/:id/edit',
-            controller: require('./modify/post-edit.controller.js'),
-            template: require('./modify/main.html')
-        }
-    );
+    // .state(
```
rowasc: can we just remove this ? or do we want it here for some particular reason?

app/main/posts/views/filters/active-search-filters.directive.js
```diff
@@ -39,7 +39,7 @@ function ActiveSearchFilters($translate, $filter, PostFilters, _, FilterTransfor
             var currentFilters = angular.copy(PostFilters.getUIActiveFilters(filters));
             FilterTransformers.rawFilters = angular.copy(filters);
             // Remove set filter as it is only relevant to collections and should be immutable in that view
-            delete currentFilters.set;
+            // delete currentFilters.set;
```
rowasc: can we remove it + the comment?   (I know you didn't change it, I can remove it in my PR,, just asking if it'd be ok  )

app/main/posts/collections/mode-context.html
rjmackay: Yes. Thats expected. The button says 'return to all posts' ... its not really 'all posts' unless we clear filters.

app/main/posts/posts-routes.js
rjmackay: Yea otherwise savedsearches//map did some weird things

app/main/posts/posts-routes.js
rjmackay: Didn't have time to check if it was being used yet

app/main/posts/views/filters/active-search-filters.directive.js
rjmackay: I changed it... I have no idea what impact that had. If its ok to drop that then yes remove the line

app/main/posts/collections/listing.directive.js
```diff
@@ -135,12 +135,9 @@ function CollectionListingController(
...
         var viewParam = collection.view;
         if (viewParam === 'list' || viewParam === 'data') {
-            /** need to reload: true for the collection to resolve on the parent state, views code to run, etc.
-             * for Child states to force a parent to run, reload is needed.
-            **/
-            $state.go('posts.data.collection', {collectionId: collection.id}, {reload: true});
+            $state.go('posts.data.collection', {collectionId: collection.id});
         } else {
-            $state.go('posts.map.collection', {collectionId: collection.id}, {reload: true});
+            $state.go('posts.map.collection', {collectionId: collection.id});
```
rowasc: since we are not relying on parent states we would not need the reload here.

app/main/posts/views/filters/active-search-filters.directive.js
```diff
@@ -13,21 +13,46 @@ function ActiveSearchFilters($translate, $filter, PostFilters, _, FilterTransfor
...
 
         $scope.isArray = angular.isArray;
 
         function activate() {
+            /**
+             * we need to listen for saved searches and collections update events
+             * so that we are notified when a user changes a collection or
+             * saved search
+             */
+            $scope.$on('savedSearch:update', function () {
```
rowasc: I moved this per a previous comment in the saved search PR

app/main/posts/views/filters/active-search-filters.directive.js
```diff
@@ -13,21 +13,46 @@ function ActiveSearchFilters($translate, $filter, PostFilters, _, FilterTransfor
...
+            });
+            $scope.$on('collection:update', function () {
+                handleFiltersUpdate(PostFilters.getActiveFilters(PostFilters.getFilters()), null, true);
+            });
+            /**
+             * we don't have filters set when we are on a collection
+             * so we need to watch for collection mode changes specifically instead of relying on the filters
+             * changing like we do for saved searches
+             */
+            $scope.$watch(function () {
```
rowasc: finding a nicer alternative   to this would be nice.. :/

app/main/posts/views/filters/active-search-filters.directive.js
```diff
@@ -13,21 +13,46 @@ function ActiveSearchFilters($translate, $filter, PostFilters, _, FilterTransfor
...
+                return PostFilters.getModeEntity('collection');
+            }, function (newValue, oldValue) {
+                if (newValue !== oldValue) {
+                    $scope.collection = newValue;
```
rowasc: just resetting to be safe

app/main/posts/views/filters/active-search-filters.directive.js
```diff
@@ -50,10 +75,10 @@ function ActiveSearchFilters($translate, $filter, PostFilters, _, FilterTransfor
              setup the originalSavedSearch (which NEVER changes) and the savedSearch, which changes
              and is used for showing the filters as the user adds/removes filters
              **/
-
-            if (resetSearch === true || !$scope.savedSearch || originalSavedSearch.id !== PostFilters.getModeId()) {
-                originalSavedSearch = angular.copy(PostFilters.getModeEntity());
-                $scope.savedSearch = PostFilters.getModeEntity();
+            else if (resetSearch === true || !$scope.savedSearch || originalSavedSearch.id !== PostFilters.getModeId()) {
+                $scope.collection = null;
```
rowasc: just resetting the col because the user changed modes.

app/main/posts/views/filters/active-search-filters.directive.js
```diff
@@ -151,8 +176,22 @@ function ActiveSearchFilters($translate, $filter, PostFilters, _, FilterTransfor
...
+             * other modes, we need to explicitly reset the filters or it will be shown
+             * as if we had manually set the collection status defaults [published, draft, archived]
+             */
+            PostFilters.setFilter('status', PostFilters.getDefaults().status);
+        }
+
+        // collections should not be saved within a saved search
         function showSaveSavedSearchButton() {
-            return !_.isEmpty($scope.uiFilters) && !$scope.savedSearch && $rootScope.loggedin;
+            return !$scope.collection && !_.isEmpty($scope.uiFilters) && !$scope.savedSearch && $rootScope.loggedin;
```
rowasc: as I understand it we shouldn't be able to save a saved search with a collection. Is that correct?

app/main/posts/views/filters/active-search-filters.html
```diff
@@ -1,3 +1,11 @@
+<div class="form-field saved-search" ng-show="collection">
+    <span class="bug">
```
rowasc: so, IDK about this. I didn't want to go into a refactor that made the saved search logic more confusing , so I just opted for having collections in the scope. Which seems the lesser evil. I think..

app/main/posts/views/filters/active-search-filters.html
```diff
@@ -1,3 +1,11 @@
...
+        <svg class="iconic" ng-click="removeCollection(collection, $event)">
```
rowasc: same as above. I chose a more verbose option rather than having a 'removeEntity' to have more control and keep things easy to read , but could be unified..

app/main/posts/views/filters/active-search-filters.directive.js
```diff
@@ -13,21 +13,46 @@ function ActiveSearchFilters($translate, $filter, PostFilters, _, FilterTransfor
...
+            $scope.$on('collection:update', function () {
+                handleFiltersUpdate(PostFilters.getActiveFilters(PostFilters.getFilters()), null, true);
+            });
+            /**
+             * we don't have filters set when we are on a collection
+             * so we need to watch for collection mode changes specifically instead of relying on the filters
+             * changing like we do for saved searches
+             */
+            $scope.$watch(function () {
+                return PostFilters.getModeEntity('collection');
```
rjmackay: Does this still work if we just watch getModeId? I assume comparing `int`s is lower overhead than comparing objects

rowasc: yea, that would work if I watch getModeId and then just go and run getModeEntity('collection') . I'll switch it

PR #947, opened 2017-11-30
Guidelines for internal testing

rowasc:
This is a simple doc to help us improve how we do testing in ushahidi. While we don't want to pollute the PR or ISSUE templates with more guidelines, it's helpful to know what constitutes a good PR checklist and how to make one. We can iterate on this before it goes into develop.

- Adds some basic guidelines for our testing efforts

Testing checklist:
- Does not apply (oh the irony =) ) 

- [ ] I certify that I ran my checklist => Does not apply, it's a doc

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/TEST_CHECKLIST_GUIDELINES.md
+++ b/TEST_CHECKLIST_GUIDELINES.md
@@ -1,0 +1,56 @@
+Testing Scripts should
+- Be easy to follow
+- Be grouped in a logical way (either page or overarching component, like "Login" or "Filters" or "Saved search") as much as possible
+- Reference other tests or groups of tests (ie: when we are done testing saved searches, we make sure that collections are also OK since we might have impacted them) 
+- Assume a basic comprehension of the system, to not be overly verbose. 
+
+    ie: we can assume that the user knows how to login if that's a feature we have had for a long time, OR at least link to the login test checklist in the "Login" instruction if we need to be more specific 
+
+- Explicitly test for non happy paths
+
+ie:  If you are desgining a login test checklist, it should include validation errors
+
+- Reflect problems we had in the past (while doing dev or testing, etc), to keep adding to our test suites as we move forward with the platform development.  
+    
+    Example: if you faced a bug like "the filters don't close when I click the x button" make sure you create a checklist for that. 
+
+- Have tests that consist of steps and preconditions
+
+    Examples: 
+    - [Test Script] Saved searches and filters
+    - [TEST GROUP] The "SAVE SEARCH" button is displayed correctly
+    - [TEST Title] The SAVE SEARCH button is displayed when there are filters.
+        - Precondition: There are no filters applied.
+        - Precondition: User is logged in
+        - View: Map mode
+            - Open the filters dropdown.
+            - Select any filter.
+            - [ ] Expected Result: the SAVE SEARCH button is shown.
+        - View: Data mode
+            - Open the filters dropdown.
+            - Select any filter.
+            - [ ] Expected Result: the SAVE SEARCH button
+            
+- For community members, PULL REQUESTS should include the checklist in the PR itself, as shown in the PULL REQUEST Template.
+
+- For internal usage, our PULL REQUESTS should link to the testing tool where the checklists live, and have notes to what exactly needs to be tested. 
+
+    Example: 
+    
+    - Testing checklist (https://ushahidi.ontestpad.com/script/2#60/18/)
+    - `Specific checklist for this fix:`
+        - [x] Go to map-view
+        - [x] Select a saved search
+        - [x] Add extra filters that don't belong to the saved search
+        - [x] Go to data-view
+        - [x] The saved search + the filters you added should be active in the data mode 
+        
+    - Relevant checklists in testpad that are fixed here: 
+        - [x] https://ushahidi.ontestpad.com/script/2#60//
+        - [x] https://ushahidi.ontestpad.com/script/2#60/18/
+
+    - General checklists that need to be run afterwards to verify nothing new is broken: 
+        - [x] Full Filtering and saved searches testpad script :  https://ushahidi.ontestpad.com/script/2
+        
+- Internal: The I certify that I ran my checklist should ideally link to the test checklist report with the result of the test run.
+    - [x] I certify that I ran my checklist . Results here https://ushahidi.ontestpad.com/script/2/report


TEST_CHECKLIST_GUIDELINES.md
```diff
@@ -0,0 +1,56 @@
+Testing Scripts should
+- Be easy to follow
+- Be grouped in a logical way (either page or overarching component, like "Login" or "Filters" or "Saved search") as much as possible
```
willdoran: +1

TEST_CHECKLIST_GUIDELINES.md
```diff
@@ -0,0 +1,56 @@
...
+- Reference other tests or groups of tests (ie: when we are done testing saved searches, we make sure that collections are also OK since we might have impacted them) 
+- Assume a basic comprehension of the system, to not be overly verbose. 
+
+    ie: we can assume that the user knows how to login if that's a feature we have had for a long time, OR at least link to the login test checklist in the "Login" instruction if we need to be more specific 
+
+- Explicitly test for non happy paths
+
+ie:  If you are desgining a login test checklist, it should include validation errors
```
willdoran: +1

TEST_CHECKLIST_GUIDELINES.md
```diff
@@ -0,0 +1,56 @@
...
+
+- Reflect problems we had in the past (while doing dev or testing, etc), to keep adding to our test suites as we move forward with the platform development.  
+    
+    Example: if you faced a bug like "the filters don't close when I click the x button" make sure you create a checklist for that.
```
willdoran: So this sounds like a mix of two elements, one is a side effect "weirdness", the second is the failure of a very small behaviour - in this case the x - I'm not sure I follow how either of those would become a checklist. We can require that a checklist clicks every single button but it seems more like the failure of a button like that is a systemic issue. All x buttons should be governed by the same code to try to reduce the possibility of this happening. If they are and something that changed, unrelated to the code that controls the x, caused the bug then the only way I imagine we can control for that is to require that ALL buttons within an interface that changed be clicked - that seems like the reason headless browser style testing existed.

May be we need to work on defining a way to move things like - "every button should pass clicking" - into basic page tests. We had those kind of tests originally but we were over using them and they became very heavy. We might have been better off just using them for extremely basic behaviours.

If the issue was caused by some kind of weird side effect, I would worry that by adding every odd case to the checklists we would create a lot of wasted time on elements that were once off bugs that should never repeat. Instead it might be better if we can stick to structure and classing the issues. I'd be happy to chat about this in order to fully explain my thinking. But in short I think, "This is the set of behaviours expected in this view as a checklist" is better than "Along with all the other tests on this checklist also make sure you try to reproduce this weird bug that happened once a long time ago that is unrelated to the interfaces behaviour". 

It seems like the latter would be better captured through super simple headless browser tests.

TEST_CHECKLIST_GUIDELINES.md
```diff
@@ -0,0 +1,56 @@
...
+    
+    Example: if you faced a bug like "the filters don't close when I click the x button" make sure you create a checklist for that. 
+
+- Have tests that consist of steps and preconditions
+
+    Examples: 
+    - [Test Script] Saved searches and filters
+    - [TEST GROUP] The "SAVE SEARCH" button is displayed correctly
+    - [TEST Title] The SAVE SEARCH button is displayed when there are filters.
+        - Precondition: There are no filters applied.
```
willdoran: This feels very manual, how do we enumerate all of them? Would it make sense to have an example of the most atomic level for one of these tests? Or may be to have some kind of "state" concept?

TEST_CHECKLIST_GUIDELINES.md
```diff
@@ -0,0 +1,56 @@
...
+            - Select any filter.
+            - [ ] Expected Result: the SAVE SEARCH button is shown.
+        - View: Data mode
+            - Open the filters dropdown.
+            - Select any filter.
+            - [ ] Expected Result: the SAVE SEARCH button
+            
+- For community members, PULL REQUESTS should include the checklist in the PR itself, as shown in the PULL REQUEST Template.
+
+- For internal usage, our PULL REQUESTS should link to the testing tool where the checklists live, and have notes to what exactly needs to be tested.
```
willdoran: +1

TEST_CHECKLIST_GUIDELINES.md
```diff
@@ -0,0 +1,56 @@
...
+        - [x] Select a saved search
+        - [x] Add extra filters that don't belong to the saved search
+        - [x] Go to data-view
+        - [x] The saved search + the filters you added should be active in the data mode 
+        
+    - Relevant checklists in testpad that are fixed here: 
+        - [x] https://ushahidi.ontestpad.com/script/2#60//
+        - [x] https://ushahidi.ontestpad.com/script/2#60/18/
+
+    - General checklists that need to be run afterwards to verify nothing new is broken:
```
willdoran: It seems that this would probably be "all" until the code is significantly more stable.

TEST_CHECKLIST_GUIDELINES.md
```diff
@@ -0,0 +1,56 @@
...
+        
+    - Relevant checklists in testpad that are fixed here: 
+        - [x] https://ushahidi.ontestpad.com/script/2#60//
+        - [x] https://ushahidi.ontestpad.com/script/2#60/18/
+
+    - General checklists that need to be run afterwards to verify nothing new is broken: 
+        - [x] Full Filtering and saved searches testpad script :  https://ushahidi.ontestpad.com/script/2
+        
+- Internal: The I certify that I ran my checklist should ideally link to the test checklist report with the result of the test run.
+    - [x] I certify that I ran my checklist . Results here https://ushahidi.ontestpad.com/script/2/report
```
willdoran: The one element this list doesn't discuss is if I am running your scripts, how do I
1. Report the failure with reference to the test that failed + screenshots + gif
2. How should the report of failure be structured? Piecemeal or all at once or...

I think this document should explicitly discuss how QA/other devs will interact with this and what expected feedback looks like.

TEST_CHECKLIST_GUIDELINES.md
```diff
@@ -0,0 +1,56 @@
...
+    
+    - Testing checklist (https://ushahidi.ontestpad.com/script/2#60/18/)
+    - `Specific checklist for this fix:`
+        - [x] Go to map-view
+        - [x] Select a saved search
+        - [x] Add extra filters that don't belong to the saved search
+        - [x] Go to data-view
+        - [x] The saved search + the filters you added should be active in the data mode 
+        
+    - Relevant checklists in testpad that are fixed here:
```
willdoran: We should also have a concept of dependencies, for example, to test SMS conversations with an Author you need to have an SMS datasource setup. So there are configuration dependencies.

This could be addressed by pre-configuring flavours of platform for automatic QA deployment.
@tuxpiper if we create a set of config files describing datasources to be enabled automatically we could then do that via openshift deployments, yes?

TEST_CHECKLIST_GUIDELINES.md
rowasc: Ah ok, good point. Right now we are reporting failures in the specific step in testpad and they should include the result, at the very least, but a screenshot would be neat if it's something that benefits from a screenshot too. They should be linked in the PR (this failed + link to failed step) 
I'll write a full section on testpad-specific or at least testing-tools specific usage for this too.

TEST_CHECKLIST_GUIDELINES.md
rowasc: yea, it is very manual. :/ that's kind of the nature of a checklist tho. Unless we roll our own or buy some super fancy testing tool (not sure which of them has this concept but I agree it would be nice) I don't see a way around it. 
Also, keep in mind that the whole thing is manual in this doc.
There has to be a separate section on e2e testing, which IMO is completely necessary at this point and should be a priority. Having this manual tests is a stepping point for then translating them to tests in our suite. But I still think you absolutely need the english only suite, because you need non- developers to check things too, you need it as documentation for the company,  and because it forces you to write it down and think through everything. I admit that to me, the process of writing the checklist is probably as valuable as the end result.

Something else about e2e is that right now we lack proper classes and ids in a bunch of parts of the app, and end of doing selectors like  "second span after tag <bla> with sibling X" instead of  "first span with this class inside this id" when doing e2e testing (at least for me that was the experience when I did it a couple weeks ago while testing testcafe) . That's another "to solve" thing for sure..

rowasc: > you need non- developers to check things too
clarifying: this is mostly because you want to have access to good documentation on the system behavior.You need a good reference that you can share with Jess for instance and say : "this is how we thought this would work when we wrote our tests, is it working correctly? can you edit it and improve it? "  then we can adjust our suite to match that.
But, after writing it, I also think "yea but that should have been in the ticket and in the manuals instead, and the test suite would have gotten the info from the manual and story ideally". So maybe we need to sit down and think more about this in general because it might point to a bigger issue in our process. :/

rowasc:
Closing. Backed up for discussion with Walter in a doc https://docs.google.com/document/d/1VfWY6JPM046nhgJiSuY166TCvQSwkNZNlRFKbtnjamE/edit?usp=sharing

PR #981, opened 2018-02-01
feat(survey-editor): Add toggle for hiding location and time info

rjmackay:
-  Add toggle for hiding location and time info in survey settings

Testing checklist:
- [ ] Edit a survey
- [ ] Open configure tab
- [ ] See options for hiding location and time
- [ ] Toggle options and check they save correctly

Refs ushahidi/platform-sivico#3
Relies on https://github.com/ushahidi/platform/pull/2398

Ping @ushahidi/platform


crcommons:
@willdoran Do you think we should put a limit on the precision level so you can't actually go beyond 10? I see that it disables the save button but it also is weird user experience where you can change the number and get: When hide location is enabled for a survey, locations will be precise to: NaNkm. Perhaps getting error text that says something like "You must choose a number between 1 and 10"

crcommons:
Issues I found:

PR #993, opened 2018-02-14
Add routes to ignore when authentication fails

rowasc:
-

Testing checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/common/auth/authentication-interceptor.config.js
+++ b/app/common/auth/authentication-interceptor.config.js
@@ -85,16 +85,12 @@ function AuthInterceptor($rootScope, $injector, $q, CONST, Session, _) {
             config.ignorable = true;
         }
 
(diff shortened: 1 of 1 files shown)


app/common/auth/authentication-interceptor.config.js
```diff
@@ -165,10 +149,29 @@ function AuthInterceptor($rootScope, $injector, $q, CONST, Session, _) {
...
             deferred.reject(rejection);
         }
         return deferred.promise;
     }
+
+    /**
+     * Returns true if url is ignorable, false if not
+     * @param config
+     */
+    function shouldIgnoreAuthError(config) {
```
willdoran: Ironically, this makes the code better :D

We need to get to a new framework. I don't see anything obvious wrong with this but it is probably fragile and should be replaced sooner rather than later.

PR #1001, opened 2018-02-19
2410 - fixing bugs

rowasc:
- fixes behavior in survey editor and add post. Two new test cases are added to the test script. 

Testing checklist:

#### Parent categories are disabled/cannot be selected if they have children  in the survey creation/survey edition screens when adding a category field
- [x] Open the "Add new post"  screen as an admin. 
- [x] Fill out required info
- [x] Try to select a parent category (that has children) 
- [x] You should NOT be able to select the parent category

#### When I only select 1 child out of many in the Survey Editor for a category field, I only see that child and its parent when I open the "Add new post" screen for that type.
Screenshots as example:
survey editor screen
screen shot 2018-02-19 at 12 40 02 pm

add new post screen
screen shot 2018-02-19 at 12 40 09 pm

- [x] Open the "Survey editor" for a survey, and add a category field with only 1 child selected for a parent category
- [x] Fill out required info, and save. 
- [x] Open the "add new post" screen
- [x] Verify that you do not see other child categories besides the one you made available for that category field

Fixes ushahidi/platform#2410 .

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/common/post-actions.service.js
+++ b/app/main/posts/common/post-actions.service.js
@@ -31,6 +31,26 @@ function (
         },
         getStatuses: function () {
             return ['published', 'draft', 'archived'];
+        },
+        filterPostEditorCategories: function (attributeOptions, categories) {
+            // adding category-objects attribute-options
+            return _.chain(attributeOptions)
+                .map((category) => {
+                    const ret = angular.copy(_.findWhere(categories, {id: category}));
+                    if (ret && ret.children.length > 0) {
+                        ret.children = _.chain(ret.children)
+                            .map((child) => {
+                                if (attributeOptions.find((o) => o === child.id)) {
+                                    return _.findWhere(categories, {id: child.id});
+                                }
+                            })
+                            .filter()
+                            .value();
+                    }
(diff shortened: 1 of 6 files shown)


rowasc:
@Angamanga this PR has two demos.Fix for select-all bug reported by Jess, and fix the add post vs add field scenario
https://drive.google.com/open?id=1ByLEYB-iDwb0kliJcPHQbhPJCreHePBu

https://drive.google.com/open?id=1drPLd2pQNxph6b8M9uQm1jBXNIpL9VRC
--

app/main/posts/common/post-actions.service.js
```diff
@@ -31,6 +31,26 @@ function (
         },
         getStatuses: function () {
             return ['published', 'draft', 'archived'];
+        },
+        filterPostEditorCategories: function (attributeOptions, categories) {
```
rowasc: moved here because the post-editor vs post-data-editor thing where we duplicate logic was driving me crazy

Angamanga: :tada:

PR #1008, opened 2018-02-23
Collections navigation and loading bar

crcommons:
- replaced onFinish with onExit so that loading status would change when state redirected; 
- added reload to ensure map reloads when going between collections.

Testing checklist:
- [x] Make sure you are logged in and have multiple collections saved;
- [ ] navigate to a collection; then try to navigate to another collection; you should be able to go between categories successfully and map should load properly
- [ ] from a category-map page, select the map button; (oddly enough) you should stay on the same page and the loading bar should NOT appear
- [ ] from a category-data page, select the data button; (oddly enough) you should stay on the same page and the loading bar should NOT appear

Fixes ushahidi/platform#2492

Ping @ushahidi/platform


app/common/services/loadingProgress.service.js
```diff
@@ -10,7 +10,7 @@ function LoadingProgress($rootScope, $transitions, $injector) {
             }
         });
 
-        $transitions.onFinish({}, function (transition) {
+        $transitions.onExit({}, function (transition) {
```
rowasc: Is this always the correct hook?does it work for other loading bar usages? e

crcommons: I honestly don't know every place we use the loading bar, but I checked as many places as I could.

rowasc: fair enough. Just wanted to make sure that it was tested in other scenarios.

rowasc:
> navigate to a collection; then try to navigate to another collection; you should be able to go between categories successfully and map should load properly

@crcommons was this meant to say "between collections"  ? Not sure how to go between categories/ what the steps would be if not

rowasc:
> from a category-map page, select the map button; (oddly enough) you should stay on the same page and the loading bar should NOT appear

@crcommons not sure what this means.

rowasc:
https://drive.google.com/open?id=1_FCSXLXtzuYY5SsEGOaB5iXm4rQTouEq
See this demo for the bug I reported here

PR #1013, opened 2018-03-06
Routing for new targeted survey

crcommons:
Added new route for targeted survey with placeholder html and controller.

NOTE: This does not have any feature flagging!!! 

- Adds a new route for targeted sms surveys
- Adds the link in the mode-bar of the survey creator page
- Adds new translation in en.json

Testing checklist:
- [x] Go to settings/surveys, then click the + button to create a new survey; On the page, in the modebar, you should see "Send a targeted survey via SMS >" 
- [x] click the "Send a targeted survey via SMS >" and you should be taken to settings/surveys/create-targeted-survey, which has placeholder text

Fixes ushahidi/platform#2524

Ping @ushahidi/platform


app/settings/settings.routes.js
```diff
@@ -71,6 +71,14 @@ function (
             template: require('./surveys/survey-edit.html')
         }
     )
+    .state(
+        {
+            name: 'settings.surveys.targeted',
+            url: '/create-targeted-survey',
```
rowasc: Can this be /targetedSurvey/createinstead? to follow the /:thing/:action way that we are using for other routes?

crcommons: For sure. I hadn't even noticed the pattern. Looks like I should open my eyes more :)

rowasc:
@crcommons cool. Can you delete the branch ? thanks :)

PR #1014, opened 2018-03-06
Update export job states

kinstelli:
-updates the status flags that the poller looks for in export_jobs to include a 'FAILED' state
-changes name of 'done' to 'SUCCESS' to differentiate between a done state and a done state that is successful

Testing checklist:
- [ ] Create an export job

Success Condition:
- [ ] Set the status of that job in the 'export_job' table to 'SUCCESS' and add a URL to a file in 'url'
- [ ] An alert should appear to let the user know that an export was successful
- [ ] A download should be initiated to whatever the url points to 

Failure Condition:
- [ ] Set the status of that job in the 'export_job' table to 'FAILED'
- [ ] An ApiError notification should appear

### DIFF
--- a/app/common/services/data-export.service.js
+++ b/app/common/services/data-export.service.js
@@ -20,7 +20,7 @@ function DataExport($rootScope, ExportJobEndpoint,  Notify, $window, $timeout, $
         var queries = [];
         ExportJobEndpoint.queryFresh({user: 'me'}).$promise.then(function (response) {
             _.each(response, function (job) {
-                if (job.status !== 'done') {
+                if (job.status !== 'SUCCESS' && job.status !== 'FAILED') {
                     queries.push(ExportJobEndpoint.getFresh({id: job.id}));
                 }
             });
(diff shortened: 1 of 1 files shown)


kinstelli:
Hi @Angamanga -- I took the liberty of updating this, because I thought it would be quicker to just add this, but I didn't realize that the apierror notification doesn't really inform the user of what has happened.

I was thinking it might make sense to either add a more details error notification that included some information, much like what is shown in exportNotifications, but I wanted to hear what you thought about this approach before I did it.

app/common/services/data-export.service.js
```diff
@@ -20,7 +20,7 @@ function DataExport($rootScope, ExportJobEndpoint,  Notify, $window, $timeout, $
         var queries = [];
         ExportJobEndpoint.queryFresh({user: 'me'}).$promise.then(function (response) {
             _.each(response, function (job) {
-                if (job.status !== 'done') {
+                if (job.status !== 'SUCCESS' && job.status !== 'FAILED') {
```
Angamanga: Yay! Yes, this is good to add!

PR #1019, opened 2018-03-08
Added form validation for each step in directive; added show/hide on …

crcommons:
- Adds form validation to targeted survey directive
- adds form html to targeted survey template
- adds test for form validation
- show/displays different "views" of text when proceeding past step 1

Testing checklist:
- [x] Survey Name field is available and user can enter data
- [x] Survey Description field is available and user can enter data
- [x] User can not proceed is one or both of the fields are blank
- [x] User can proceed if both of the fields are filled
- [x] When moving to next step, styles for fields should condense and look neat and nice (ignore light/smokey front in design. Justin told me it's wrong. Just defer to normal <p> font.

- [X] I certify that I ran my checklist

Fixes ushahidi/platform#2527

Ping @ushahidi/platform

### DIFF
--- a/app/common/locales/en.json
+++ b/app/common/locales/en.json
@@ -1206,7 +1206,16 @@
             "destroy_stage_success" : "Survey task {{name}} deleted",
             "delete_attribute_confirm" : "Are you sure you want to delete this field?",
             "delete_attribute_confirm_desc" : "This action cannot be undone. Deleting this field will remove its data from all existing posts.",
-            "destroy_attribute_success" : "Field {{name}} deleted"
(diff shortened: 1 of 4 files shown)


crcommons:
@Angamanga I need to figure out what the actual validations are for each field (min/max length, etc...) I also need to figure out how to add the asterik. If you know how to do those, can you add them? Otherwise, you can either merge and we'll figure it out together tomorrow or branch off of this branch and we'll still figure it out tomorrow.

Angamanga:
@crcommons I added the required asterix, error-label and also changed to ng-form, i think that will make most sense in the end, to keep it as one form instead of several, when its time to publish.

PR #1033, opened 2018-03-13
BUGFIX(data-export.service.js, data-export.controller.js): export history list is updated when user requests a new job, expiration date is empty when still pending

rowasc:
- Export history list is updated when user requests a new job or when a job status changes
- Expiration date is empty when job status is pending

Testing checklist:
- [x] https://github.com/ushahidi/platform/issues/2582 (edited the script and added a new one for updates)

- [x] I certify that I ran my checklist (this depends on https://github.com/ushahidi/platform/pull/2598)

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/common/services/data-export.service.js
+++ b/app/common/services/data-export.service.js
@@ -76,14 +76,17 @@ function DataExport($rootScope, ExportJobEndpoint,  Notify, $window, $timeout, $
 
     function updateExportJobsList(job) {
         var _exportJobs = getExportJobs();
-        const foundJob = _.findWhere(_exportJobs, { id: job.id });
+        const foundJobIndex = _.findIndex(_exportJobs, (_job) => {
+            return _job.id === job.id;
+        });
 
-        if (foundJob) {
-            setExportJobs(_.extend(foundJob, job));
+        if (foundJobIndex >= 0) {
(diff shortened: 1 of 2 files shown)


app/common/services/data-export.service.js
```diff
@@ -67,6 +70,19 @@ function DataExport($rootScope, ExportJobEndpoint,  Notify, $window, $timeout, $
         }, CONST.EXPORT_POLLING_INTERVAL);
     }
 
+    function updateExportJobsList(job) {
```
Angamanga: I think you need to process the list here as well. When I start a new job, the expiry-date shows up as 'invalid' if I don't reload the page

rowasc: good catch, thanks!

rowasc:
@Angamanga updated. 
FYI, I was not able to reproduce the user problem (where you see data that does not belong to your user) locally,  I'll test that again tomorrow morning when you are around.

app/common/services/data-export.service.js
rowasc: ok so it ended up being a different problem involving the date of pre processed jobs (which was being converted again and resulting in an invalid date) . I also found a bug with the sorting of the jobs which is fixed now.

PR #1039, opened 2018-03-19
Fixing share survey URL

crcommons:
- adjusts share url of survey so that it uses current url's survey id and post/create

Testing checklist:
- [ ] go to settings/surveys and select a survey; go to the share tab; url should be .../posts/create/ with the survey ID

Fixes ushahidi/platform#2431

Ping @ushahidi/platform


app/main/posts/views/share/share-menu.directive.js
```diff
@@ -39,10 +39,8 @@ function ShareMenuController(
         // an individual post that the URL is correct
         if ($scope.postId) {
             $scope.shareUrl = $window.location.origin + '/posts/' + $scope.postId;
-        }
-
-        if ($scope.surveyId) {
-            $scope.shareUrl = $window.location.origin + '/posts/create/' + $scope.surveyId;
+        } else {
+            $scope.shareUrl = $window.location.origin + '/posts/create/' + $scope.shareUrl.slice(-1);
```
Angamanga: This doesn't work on survey-ids > 10, we should really have surveyId but its not there yet, not sure why, maybe thats the underlying problem :/. Maybe you could use the transition-params similar to in edit-controller: https://github.com/ushahidi/platform-client/blob/develop/app/settings/surveys/edit.controller.js#L10.

PR #1043, opened 2018-03-23
2612 country code

Angamanga:
NOTE! This pr only works together with branch 2599-country-codes in Platform-api. Remember to do migrations.

- Adds the country-code-endpoint that communicates with the api-endpoint
- Uses the country-code-endpoint in targeted-survey-directive
- Adds translation-strings for each country
- Adds a loading-message while country-codes are fetched from the api

Testing checklist:
- [x] As logged in, go to settings-surveys-add targeted survey
- [x] Fill in name and description, click on next
- [x] A button displaying ('loading country-codes') should be viisble while loading
- [x] Once country-codes are loaded, a dropdown with a list of countries and country-codes should be visible
Select a country
- [x] Add a faulty number
- [x] An error-message should appear
- [x] Add a correct number
- [x] You should be able to move on to step 3

Fixes ushahidi/platform#2612 .

Ping @ushahidi/platform


app/settings/surveys/targeted-surveys/targeted-edit.controller.js
```diff
@@ -118,6 +120,7 @@ function (
         if (isStepComplete($scope.targetedSurvey.stepOne)) {
             $scope.activeStep = 2;
             $scope.stepOneWarning = false;
+            $scope.getCountryCodes();
```
crcommons: Do you think we should get the country codes with the page load, rather than when step one is complete? For slow connections, it may be better to have that running in the background while the user completes step 1. Thoughts?

Angamanga: I was thinking about that, not sure, at the same time, this divides the pageload into smaller pieces. But maybe the user is more accepting of longer loading-time at the inital-pageload? I'll make a test and compare...

crcommons: Or we could run that request once the rest of the page loads but before you move to the next step? This way we allow the page to be interactive and usable, but still get the country codes before the user moves on to step 2. Also this isn't really that big of a deal, so you can tell me to shut up if you want.

Angamanga: I moved it to page-init, works fine now so merging

PR #1044, opened 2018-03-23
Ordering of fields on Post edit respects Survey order

willdoran:
- Allows Title and Description fields to be reordered by adding toggles to those fields
- Ensure Post Edit views(Data and Classic) respect the ordering of the fields

Testing checklist:
- [x] Create a new Survey
- [x] The Title and Description fields should have re-order toggle buttons
- [x] When clicked the toggle buttons should behave as expected
- [x] Save the Survey
- [x] From the Survey List view, click on the newly created Survey
- [x] The order of the Title and Description is correct as previously saved
- [x] Add a new field to the Survey
- [x] Reorder the new field relative to Title and Description
- [x] Save the Survey
- [x] From either the Map or Data view create a new Post from the Survey that was created in the previous steps
- [x] Create a new Post
- [x] The order of the fields - including Title and Description - should be the same as shown in the Survey Editor
- [x] Save the Post
- [x] On the Post detail page click edit
- [x] The order of the fields - including Title and Description - should be the same as shown in the Survey Editor

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/modify/post-value-edit.html
+++ b/app/main/posts/modify/post-value-edit.html
@@ -1,229 +1,286 @@
 <div>
-<!-- Container for form-field structured elements -->
-  <div
-      ng-if="!isFieldSetStructure(attribute)"
-      class="form-field init"
-      adaptive-form
-      ng-class="{
-        'error': form['values_' + attribute.key].$invalid && form['values_' + attribute.key].$dirty,
-        'success': ! form['values_' + attribute.key].$invalid && form['values_' + attribute.key].$dirty,
-        'required': attribute.required,
-        'date' : isDate(attribute) || isDateTime(attribute),
-        'location' : attribute.input === 'location'
-      }">
+    <!-- Container for form-field structured elements -->
+    <div ng-if="attribute.type === 'description' || attribute.type === 'title'">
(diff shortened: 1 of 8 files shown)


app/main/posts/modify/post-value-edit.html
```diff
@@ -1,229 +1,286 @@
...
+                        attribute="attribute"
+                        post-value="post.values[attribute.key]"
+                    ></add-category>
+                </div>
+                <div class="alert error" ng-show="form['values_' + attribute.id].$dirty && taskIsMarkedCompleted()" ng-repeat="(error, value) in form['values_' + attribute.id].$error">
+                    <svg class="iconic">
+                    <use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="/img/iconic-sprite.svg#warning"></use>
+                    </svg>
+                    <span translate="{{'post.valid.values.' + error}}" translate-values="{ label : attribute.label }"></span>
+                </div>
```
rowasc: @willdoran not sure if from this PR, but there's a </div> missing here before </fieldset>

willdoran: Think that's fixed now

PR #1050, opened 2018-03-23
Importing updates

willdoran:
- Merging fixes from core

Testing checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


rjmackay:
If we're just merging fixes from master into a feature branch I don't think that justifies another PR. We should simply update the original PR (https://github.com/ushahidi/platform-client/pull/981) and retest it.

PR #1051, opened 2018-03-26
fix: adding routes to ga

kinstelli:
- Includes route transitions among data being sent to google analytics

Fixes ushahidi/platform#1524.

Ping @ushahidi/platform

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -151,19 +151,26 @@ angular.module('app',
             { map: {}, site: {}, features: {} };
     }])
     // inject the router instance into a `run` block by name
-    // .run(['$uiRouter', '$trace', '$location', function ($uiRouter, $trace, $location) {
+    //.run(['$uiRouter', '$trace', '$location', function ($uiRouter, $trace, $location) {
     //     // * uncomment this to enable the visualizer *
     //     let Visualizer = require('@uirouter/visualizer').Visualizer;
     //     let pluginInstance = $uiRouter.plugin(Visualizer);
     //     $trace.enable('TRANSITION');
     // }])
-    .run(['$rootScope', 'LoadingProgress', function ($rootScope, LoadingProgress) {
+    .run(['$rootScope', 'LoadingProgress', '$transitions', function ($rootScope, LoadingProgress, $transitions) {
         $rootScope.$on('$stateChangeError', console.log.bind(console));
         // this handles the loading-state app-wide
         LoadingProgress.watchTransitions();
+        if (window.ushahidi.gaEnabled)
+        {
+            $transitions.onSuccess({}, function (transition) {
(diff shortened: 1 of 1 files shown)


app/app.js
```diff
@@ -151,16 +151,21 @@ angular.module('app',
...
-    // .run(['$uiRouter', '$trace', '$location', function ($uiRouter, $trace, $location) {
+    //.run(['$uiRouter', '$trace', '$location', function ($uiRouter, $trace, $location) {
     //     // * uncomment this to enable the visualizer *
     //     let Visualizer = require('@uirouter/visualizer').Visualizer;
     //     let pluginInstance = $uiRouter.plugin(Visualizer);
     //     $trace.enable('TRANSITION');
     // }])
-    .run(['$rootScope', 'LoadingProgress', function ($rootScope, LoadingProgress) {
+    .run(['$rootScope', 'LoadingProgress', '$transitions', function ($rootScope, LoadingProgress, $transitions) {
         $rootScope.$on('$stateChangeError', console.log.bind(console));
```
rjmackay: Does this even work?

app/app.js
```diff
@@ -151,16 +151,21 @@ angular.module('app',
...
         // this handles the loading-state app-wide
         LoadingProgress.watchTransitions();
+        if (window.ushahidi.gaEnabled) {
+            $transitions.onSuccess({}, function (transition) {
+                window.ga('send', 'pageview', transition.to().name);
```
rjmackay: That makes sense. Does transition.to().name translate to a URL though? or just a state name?
We should try to make it a URL since thats expected by GA

kinstelli: Good point. I'll change it.

app/app.js
kinstelli: I don't think $stateChangeError or $stateChangeSuccess are recognized at all now. Should we remove this or replace with the $transitions equivalent?

app/app.js
kinstelli: I looks like we have two options, `transition.to().name` and `transition.to().url`.

In the GA content list, the `name` format produces the URLs separated by dots. The `url` format produces a URL with slashes. Using `name` shows more deployment-specific info, including record IDs). Both are shown in the example image here, since I switched them out in the middle of collecting data:

screen shot 2018-04-03 at 10 57 10 am

kinstelli:
I tried it locally with my own GA code. 
Let me know if this looks right to you:

screen shot 2018-04-03 at 10 57 10 am

crcommons:
I think that looks right! @rowasc Can you confirm that this look correct to you, since you're the GA expert here :)

rowasc:
ha. Not an expert but, as long as this doesn't result in analytics entries where some are dots and other  are slashes,it looks good.

PR #1056, opened 2018-03-28
fix: Populate manual lat / lon entry with current location

rjmackay:
- Populate manual lat / lon entry on load
- Populate manual lat / lon entry when map is clicked
- Populate manual lat / lon entry when search result chosen
- Populate manual lat / lon entry when current location used
- Show location lat/lon entry below the map on post detail view

Refs: https://github.com/ushahidi/platform-sivico/issues/33

Testing checklist:
- [x] Edit a post with a location
- [x] See lat / lon is shown when edit loads
- [x] Click on map. See lat /lon is updated
- [x] Search for a location and chose a result. See lat /lon is updated
- [x] Click in search and select 'use my current location'. See lat /lon is updated
- [x] Edit lat / lon. Click Update Map. See map updates
- [x] View a post. See `lat, lon` displayed below map

Ping @ushahidi/platform

### DIFF
--- a/app/common/locales/en.json
+++ b/app/common/locales/en.json
@@ -1226,7 +1226,9 @@
         "click_map": "<em>You can search or click the area of map where you want to place the marker.</em>",
         "placeholder":"Search for addresses, place names, or coordinates...",
         "search_results": "Search results",
-        "error":"Sorry we could not find a location matching your search"
+        "error":"Sorry we could not find a location matching your search",
+        "update_map":"Update map",
+        "no_matching_locations":"No matching locations"
     },
     "notification": {
         "title": "Notifications",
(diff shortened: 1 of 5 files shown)


rjmackay:
@willdoran 

> Click in search and select 'use my current location'. See lat /lon is updated

Were you testing over https? I believe browsers block pulling user location on insecure pages.

> View a post. See lat, lon displayed below map

I just retested and this is working for me. Where did you test? Which post?

willdoran:
@rjmackay Tested both locally

- Current location worked correctly to find the location but didn't update the fields
- No lat long were displayed under the map 

But it does appear to be working on the deployment so perhaps it's an edge case?

rjmackay:
> Current location worked correctly to find the location but didn't update the fields

Ok confirmed this breaks on my local too. Let me fix it :(

rjmackay:
> Current location worked correctly to find the location but didn't update the fields

This is now fixed. The update was running outside the angular digest cycle and so not being rendered.

Also fixed some missing translation strings for 'Update map' and 'No matching locations.'

rjmackay:
FYI the auto deployment for this seems a bit broken because its trying to hit the lumen/staging API

PR #1059, opened 2018-03-29
Sms wire contacts

Angamanga:
- Adds a form-contact endpoint to client
- Saves contacts of a targeted-survey to form-contacts
- Adds error-notifications if something goes wrong in each saving-form-step
- Adds form-id and form-stage-id to survey in case something goes wrong in any step. In that way we avoid double-saving a survey
- Small ui-fix, fixing the mode-bar dropdown on small screens + setting correct layout

Testing checklist:
- [x] Go to create targeted survey
- [x] Fill in each step
- [x] Click on "Publish"
- [x] Survey should be saved (stats are still incorrect) and user should be notified
- [x] The numbers that are added should be saved in the form-contacts table in the db
- [x] Decrease size of screen until the modebar is visible at the top of the screen
- [x] There should be a dropdown on the modebar that shows the breadcrumbs and the link to create a crowdsourced survey

Fixes ushahidi/platform# .

Ping @ushahidi/platform


crcommons:
@Angamanga Do we need to route the "sent" view once it sends? If yes, you can either add here or one of us can do it in another PR.

Angamanga:
@crcommons Good idea about the reroute, rerouting to targeted-surveys/published/{{form-id}}, same as when a user clicks that route in the survey-list...

PR #1076, opened 2018-04-13
changing order to priority #2734

Angamanga:
- Changes naming of "order" to "priority"
- Changes first priority to 1
- Updates tests 

Testing checklist:
- Create a new targeted survey with more than 1 question
- Save it
- Check the db, the form_attributes should be set to same priorities as the order they are listed in the targeted survey
- Create another targeted survey with more than 1 question
- change order of the questions by dragging one question to a new place in the list
- [x] Check the db, the priority for each attribute in the survey should reflect the same order thats in the survey after moving the questions around

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/settings/surveys/targeted-surveys/targeted-edit.controller.js
+++ b/app/settings/surveys/targeted-surveys/targeted-edit.controller.js
@@ -142,7 +142,7 @@ function (
         _.each(items, function (item, index) {
             _.each($scope.survey.attributes, function (question) {
                 if (question.label === item.getAttribute('data')) {
-                    question.order = index + 1;
+                    question.priority = index + 1;
                 }
             });
         });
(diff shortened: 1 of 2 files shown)


app/settings/surveys/targeted-surveys/targeted-edit.controller.js
```diff
@@ -142,7 +142,7 @@ function (
         _.each(items, function (item, index) {
             _.each($scope.survey.attributes, function (question) {
                 if (question.label === item.getAttribute('data')) {
-                    question.order = index + 1;
+                    question.priority = index + 1;
```
rowasc: Important to consider: priority is selected by users, when they sort attributes. We should never force a higher priority on items unless our UI let them set an equal priority in 2 items.

rowasc: (basically,we don't choose priority, users do)

rowasc: nvm, I just got it sorry

rowasc:
This looks good. I couldn't test but code looks ok

test/unit/settings/surveys/targeted-edit.controller.spec.js
```diff
@@ -132,26 +132,26 @@ describe('setting create targeted survey controller', function () {
                 };
                 $scope.editQuestion = {question: 'Test question?', input: 'textarea', newQuestion: true};
                 $scope.addNewQuestion();
-                let adjustedQuestion = {question: 'Test question?', input: 'textarea', order: 3, type: 'text', label: 'Test question?'};
```
rowasc: well done on the tests 🎉

PR #1105, opened 2018-04-24
Missing translations

Angamanga:
- Adds missing translation-strings to app where we have full control over the strings within the client
- Adds translation-strings and adds translation-attribute to the client to roles, permissions and default saved-searches, (the actual text is controlled in the api)

Testing checklist:
- [ ] Check that the translations look ok in the code
- [ ] Double-check that the translations/mapping with en.json works properly in the client where it was changed in the code

Fixes ushahidi/platform# .

Ping @ushahidi/platform


app/main/posts/savedsearches/mode-context.html
```diff
@@ -3,7 +3,7 @@
         <ol class="breadcrumbs" ng-controller="navigation as nav">
             <li><a href="/">{{nav.site.name}}</a></li>
         </ol>
-        <h1 class="mode-context-title">
+        <h1 class="mode-context-title" translate>
```
crcommons: What are we translating here?

Angamanga: Ah, the saved-search name. When we create a deployment, we also create 3 saved searches, one for the posts from the users, one for unstructured posts and one for all published posts. This is preparation for those strings, but they will come from the api.

app/common/directives/filter-system/filter-role.html
```diff
@@ -7,8 +7,7 @@
                 checklist-model="model"
                 checklist-value="role.name"
                 >
-            {{role.display_name}}
+            {{role.display_name | translate}}
```
rjmackay: Isn't display name user editable? Any reason users can't just edit the role names to match their language?

app/main/posts/savedsearches/mode-context.html
rjmackay: I think this will strip the icon when translated. So
1. you need to wrap just the search name
2. the user should really just edit the saved search name. I don't think we should be trying to translate user editable content.

app/common/directives/filter-system/filter-role.html
Angamanga: I added this so we can translate the pre-added role-names. If we support many different languages, I think we should translate those as well? Same as with the pre-added saved searches. However, for this specific client, we could do it manually. Leaving this in prepares for translating the role-names, it still doesnt mess up if a user edits it (if they don't edit it to one of our translation-ids, which I think is unlikely)

Angamanga: I finally removed these translations since they can be done in the api instead

rjmackay: Ok. The experience will be a bit weird when editing though. ie. if we ship a translation for 'Admin' to French, but the user goes to edit the Admin role. They'll see the original English version in the edit view, not the French.
Anyway - a real solution has to wait till we have time to make roles/categories and other user editable translatable in the API - in the interrim this makes it easier to support clients I guess

PR #1134, opened 2018-05-08
feat: Add routing using a wrapper controller

rjmackay:


app/settings/TestRouteContainer.jsx
```diff
@@ -0,0 +1,21 @@
...
+import PropTypes from 'prop-types';
+
+const TestRouteContainer = props => {
+    console.log(props);
+    return (
+        <div><main role="main">
+            <h1>Hello from react</h1>
+            { props.id }
+            <a onClick={ () =>  { props.state.go('posts.data') } }>Click me</a>
+            <a href="/views/map">Click me</a>
```
rjmackay: It seems a simple href works just as well as using `state.go` so I'd probably stick with hrefs for simpler refactoring in future.

app/settings/settings.routes.js
```diff
@@ -190,5 +190,16 @@ function (
...
             template: require('./plans/plan.html')
         }
+    )
+    .state(
+        {
+            name: 'testroute',
+            url: '/settings/testroute/:id?',
+            controller: ['$scope', '$state', '$transition$', ($scope, $state, $transition$) => {
+                $scope.$state = $state;
+                $scope.id = $transition$.params().id;
```
rjmackay: Handling params here keeps that from polluting our react code for now

PR #1136, opened 2018-05-09
Feature/1119 hdx button data export WIP

Angamanga:
- Adds a warning-text to the user if no api-key is saved
- Adds a button to hdx
- Adds a route, controller and test-file for hdx

*Waiting on*
User api-key-settings-UI (can be tested through adjusting the db)
2807-user-settings in the api is needed

Testing checklist:
### QA-test checklist
Assuming HDX-flag is on
- Go to settings-data-export without any api-key saved
- [ ] 3 buttons should be visible, the last one greyed out
- [ ] An alert with a link to api-key-config should be visible
- Configure api-key and go back to settings-data-export
- [ ] 3 buttons should be visible, all of them active
- [ ] Click on the button that says "Assign tags and attributes"
- [ ] You should be directed to the tags and attributes page (doesnt need to contain anything at this stage, but the final result should look like this: https://xd.adobe.com/view/2690f082-d88d-4788-5db7-b04c9474a404-50a1/screen/9a17d59a-cdfd-4ed4-aec5-573cdfeadb97/Data-export-HDX-single-survey)

Fixes ushahidi/platform# .

Ping @ushahidi/platform


app/settings/data-export/data-export.controller.js
```diff
@@ -54,6 +56,13 @@ function (
         }
     });
 
+    // Check if hxl-api-key is added
+    UserSettingsEndpoint.getFresh({id: $rootScope.currentUser.userId}).$promise.then((user) => {
+        if (user.count !== 0) {
+            $scope.hxlApiKey = true;
```
willdoran: Do we need to add a specific check for the presence of the hxl_api_key? I think this checks to see if there are user settings associated with the given user and if so then we don't need them to enter it again, otherwise we have to ask for it. Is that right or am I way off?

Angamanga: I am just checking if there is a hxl_api_key associated by the user here. Am not using the key for anything else than i displaying a warning and disable a button if its not there.

PR #1138, opened 2018-05-10
test(ush-logo test files): got tests working with store data

crcommons:
- Made it so that Jest isn't mad at angular and now tests are passing

Testing checklist:
- [ ] Run tests

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/react-transition/connectWithStore.jsx
+++ b/app/react-transition/connectWithStore.jsx
@@ -8,13 +8,9 @@ const connectWithStore = (WrappedComponent, ...args) => {
   const ConnectedComponentWithStore = props => {
     // Get the injector + store at render time
     // Otherwise it happens before the angular app is set up
-    if (!props.store) {
-      const injector = angular.element(document).injector();
-      /* eslint-disable no-param-reassign */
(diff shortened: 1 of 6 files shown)


app/react-transition/connectWithStore.jsx
```diff
@@ -1,15 +1,20 @@
...
 
   const ConnectedComponentWithStore = props => {
     // Get the injector + store at render time
     // Otherwise it happens before the angular app is set up
-    const injector = angular.element(document).injector();
-    const store = injector.get("$ngRedux");
-    return <ConnectedWrappedComponent store={store} {...props} />;
+    if (!props.store) {
+      const injector = angular.element(document).injector();
+      /* eslint-disable no-param-reassign */
```
crcommons: I couldn't figure out how to do this so that the linter wouldn't get mad. @rjmackay do you know what to do here?

rjmackay:
From the jest readme:
> The minimum supported Node version is v6.0.0 by default.

Travis only breaks on Node 5. We should just drop support for that and probably add newer version into our matrix too.

PR #1139, opened 2018-05-10
Migration/link api

crcommons:
- Creates a Link component that returns <a> tag that will navigate to angular or react components

Testing checklist:
- [ ] Go to /settings/testroute/1 and click the link
- [ ] you should be taken to data view
Tests fail because of problems with angular and jest

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/react-transition/Link.spec.jsx
+++ b/app/react-transition/Link.spec.jsx
@@ -1,7 +1,14 @@
 import React from "react";
 import renderer from "react-test-renderer";
 import { shallow } from "enzyme";
+import angular from "angular";
+import uirouter from "@uirouter/angularjs";
 import Link from "./Link";
+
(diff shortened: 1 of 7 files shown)


app/react-transition/Link.spec.jsx
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+import renderer from "react-test-renderer";
+import { shallow } from "enzyme";
+import Link from "./Link";
+
+test("Link accepts a route string, replace bool, id, className, title, and children", () => {
+  const component = renderer.create(
```
rjmackay: We probably need to set up a fake angular app before this. So that `angular.element(document)` returns something useful

PR #1141, opened 2018-05-13
Feature/2848 hdx fetch surveys

Angamanga:
WIP need https://github.com/ushahidi/platform-client/pull/1136 and https://github.com/ushahidi/platform/pull/2923 to be merged

- Adds the tags/attribute-screen to hdx-exports

Testing checklist:
Assuming HDX-feature-flag is ON
- Go to data-export-settings, click on "assign attributes and tags" button
- [x] All surveys (not targeted) and survey-fields should be visible
- [x] No targeted surveys should be visible
- [x] There should be a select-all button, when clicking on that, all fields should become selected
- [x] There should be checkboxes in front of each survey-field
- [ ] The numbers of fields selected should be the same as the count in the text in the end of the table
- [x] After each field there should be two dropdowns, one for tags and one for attributes
- [ ] Each dropdown should be populated with the tags and attributes fetched from the api
- select one field
- [x] the dropdowns should become active
- Select a tag
- [ ] the tag should appear as an icon in the end of the row according to designs
- [ ] the selected tag should be greyed out in the dropdown
- Select an attribute
- [ ] the attribute should be attached to the icon according to designs <--note, this is just a text-list right now, known bug
- [ ] selected attributes should be greyed out in the dropdown
- Click on the yellow x in the icon <--note, this is just a text-list right now, and there is no yellow x, known bug
- [ ] The whole tag should be removed and the dropdowns should contain no greyed out<--note, this is just a text-list right now so cannot be tested, known bug
 tags/attributes
- Select another field
- Select a tag and an attribute
- Click on the "Add another hdx attribute link"
- [ ] Another dropdown should be visible with all attributes available, the previously selected attributes shoud be disabled/greyed out.
- deselect the last attribute
- [ ] this attribute should disappear from the icon
- add another attribute
- [ ] this new attribute should appear in the icon
- select another attribute in the same dropdown
- [ ] the newly selected attribute should appear in the icon instead of the old one

Fixes ushahidi/platform# .

Ping @ushahidi/platform


app/common/locales/en.json
```diff
@@ -1080,8 +1080,26 @@
...
-        "export": "Export",
-        "export_history": "Export History"
+        "tags_attributes": "Assign tags and attributes",
+        "hxl_apikey_alert_1": "It looks like you haven't set up your Humanitarian Data Exchange (HDX) API. Without this set up, you can't tag data and upload to a Humanitarian Data Exchange (HDX) account.",
+        "hxl_configure": "Configure HDX API",
+        "hxl_apikey_alert_2": "We strongly advide you set this up now in",
+        "hxl_apikey_alert_3": "before continuing as you won't be able to upload your data to HDX",
+        "hxl_title": "Assign your HXL tags and attributes, select fields and export or upload",
+        "hxl_desc": "Select specific fields from a survey, assign a tag per field and then choose to assign multiple attributes. Preview your tag and attributes in tag preview.",
+        "hxl_instructions_1": "You can then choose to export your tagged fields to a CSV file or upload to a",
```
rowasc: @Angamanga the translations seem to be broken, can you check? (ie data_export.export instead of the text)

app/settings/data-export/data-export.controller.js
```diff
@@ -54,6 +56,16 @@ function (
         }
     });
 
+    // Check if hxl-api-key is added
+    UserSettingsEndpoint.getFresh({id: $rootScope.currentUser.userId}).$promise.then((settings) => {
+        $scope.hxlApiKey = false;
+        _.each(settings, (setting) => {
```
rowasc: you should iterate over `settings.results ` to get the items in settings (sorry if someone covered this already in another PR )

app/settings/data-export/hdx-export.controller.js
```diff
@@ -0,0 +1,149 @@
...
+        if (_.findWhere(selectedHxlAttributes, {id: hxlAttribute.id})) {
+            disabled = true;
+        }
+        return disabled;
+    }
+
+    function selectAll(form) {
+        if (form.selected) {
+            _.each(form.attributes, (attribute) => {
+                attribute.selected = [attribute.id];
```
rowasc: Hey, do we need to use a for each here or can we use a `map` instead?  
ie : 
```
form.attributes =  form_attributes.map(attribute => {
    attribute.selected = [attribute.id];
    return attribute;
}
```

feel free to ignore if too nitpicky/can't do it now, it's just a suggestion since I'd prefer it if we use `map` when we can and when changing objects in the array. It's also faster than foreach 🎉 and  allows for chaining if you return it which is handy.

app/settings/data-export/data-export.controller.js
Angamanga: Yep, true, fixed now

app/common/locales/en.json
Angamanga: fixed!

app/settings/data-export/hdx-export.controller.js
Angamanga: Fixed 🎉

app/settings/data-export/hdx-export.controller.js
```diff
@@ -0,0 +1,152 @@
...
+        let data = [];
+        _.each($scope.forms, (form) => {
+            _.each(form.attributes, (formAttribute) => {
+                if (formAttribute.selected && formAttribute.selected.length > 0) {
+                    let ids = {
+                            form_attribute_id : formAttribute.id,
+                            hxl_tag: null
+                        };
+
+                    if (formAttribute.selectedTag) {
```
rowasc: pending structure change

Angamanga: I will fix the structure-change in another pr

PR #1167, opened 2018-06-01
3039 messages missing from post edit

willdoran:
- Adding conversation with author section to post edit views

Testing checklist:
- [ ] Follow test checklist for [3038](https://github.com/ushahidi/platform-client/pull/1164)
- [ ] On the Post edit view for one of the Posts generated for the recipients of the targeted survey you should see the conversation with author section below the Post fields section

Fixes ushahidi/platform#3039

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/modify/post-data-editor.html
+++ b/app/main/posts/modify/post-data-editor.html
@@ -59,6 +59,7 @@
           ng-if="form"
         >
         </post-tabs>
+        <post-messages post="post" ng-if="post.contact.id"></post-messages>
       <div class="toolbar toolbar-secondary">
         <div class="button-group">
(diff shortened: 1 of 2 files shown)


kinstelli:
Looking at this more closely, I'm not sure 'Title' makes sense as the first outgoing item in conversation with author. Is that intended behavior?
screen shot 2018-06-01 at 6 17 55 pm

willdoran:
@kinstelli NO good catch, I'll take a look at it.

willdoran:
@kinstelli Could you retest this to ensure that it is only sending the first question, then the second question and NOT Title or Description

PR #1182, opened 2018-06-13
Feature/3054 people list

rowasc:
- adds a list of people
- applies old styles
- has broken language :| 

Testing checklist:
- [ ] Criteria on platform/3054

Fixes ushahidi/platform# .

Ping @ushahidi/platform


app/react/settings/people/PeopleNavigationContainer.jsx
```diff
@@ -0,0 +1,24 @@
...
+    <div className="mode-context init">
+        <header className="mode-context-header">
+            <ol className="breadcrumbs">
+                <li>
+                    <a href="/" ng-controller="navigation as nav">
+                        nav.site.name
+                    </a>
+                </li>
+                <li>
+                    <a href="/settings" translate>
```
crcommons: Are you trying to use translate the same way we use it in Angular here?

app/react/settings/people/PeopleToolbarContainer.jsx
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+import Link from "../../react-transition/Link";
+
+const PeopleToolbarContainer = () => (
+    <div className="toolbar">
+        <div className="fab">
+            <Link
```
crcommons: We should be able to user react router Link here, sine we're navigating to a react component.

app/react/settings/people/PeopleToolbarContainer.jsx
```diff
@@ -0,0 +1,25 @@
+import React from "react";
+import Link from "../../react-transition/Link";
+
+const PeopleToolbarContainer = () => (
```
crcommons: Should we make this a reusable component now, or refactor when we need it next?

app/react/settings/people/PersonCard.jsx
```diff
@@ -0,0 +1,30 @@
...
+import { Link, BrowserRouter } from "react-router-dom";
+
+const PersonCard = data => (
+    <BrowserRouter>
+        <Link to={`/settings/usersEdit/${data.person.id}`}>
+            <div className="listing-item">
+                <div className="listing-item-primary">
+                    <div className="listing-item-image">
+                        <img
+                            className="avatar"
```
crcommons: The designs indicate we don't need the avatar anymore for the person in list view

app/react/settings/people/PersonCard.jsx
```diff
@@ -0,0 +1,30 @@
+import React from "react";
+import { Link, BrowserRouter } from "react-router-dom";
+
+const PersonCard = data => (
+    <BrowserRouter>
```
crcommons: Not sure we need BrowserRouter at this level. Would need to experiment around a bit though.

app/react/settings/people/PersonCard.jsx
rowasc: Right, but we are using the old designs so it made sense to have it

app/react/settings/people/PeopleToolbarContainer.jsx
rowasc: this will probably go out with the new designs, I'd refactor into generic after we start getting them

app/react/settings/people/PersonCard.jsx
rowasc: You can't really do it without the browser router AFAIK, the link goes down in flames. Maybe I'm doing something wrong using link tho, need to recheck docs/egghead

app/react/settings/people/PeopleNavigationContainer.jsx
rowasc: nope I just did some copy-pasting html and this stayed. Removing

app/react/common/state/people/people.actions.js
```diff
@@ -1,12 +1,16 @@
 import PersonEndpoints from "react/common/endpoints/people";
 
 export const SAVE_NEW_PERSON = "SAVE_NEW_PERSON";
+export const FETCH_PEOPLE = "FETCH_PEOPLE";
 export const RECEIVE_PERSON = "RECEIVE_PERSON";
 export const HANDLE_REQUEST_FAILURE = "HANDLE_REQUEST_FAILURE";
-
+export const RECEIVE_PEOPLE = "RECEIVE_PEOPLE";
 export function receivePerson(person) {
```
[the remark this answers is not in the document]
rowasc: no, not really because people is the correct term here, IMO. But we could use receivePeople instead of receivePerson

app/react/settings/people/PersonCard.jsx
rowasc: ah we didn't have one at all. Moved to peoplelistcontainer, but we should probably have a <main> somewhere for that sort of thing

PR #1186, opened 2018-06-26
Adding component-library to client WIP

Angamanga:
- Adds component-library to migration-branch
- Adds story-book to migration-branch
- Adds necessary libraries for storybook
- Changes webpack config to import sass/css in css-modules

Testing checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


package.json
```diff
@@ -24,7 +24,9 @@
     "retry": "git-cz --retry",
     "test": "karma start ./test/karma.conf.js --single-run && jest",
     "tdd:karma": "karma start ./test/karma.conf.js --auto-watch",
-    "tdd:jest": "jest --watch"
+    "tdd:jest": "jest --watch",
+    "storybook": "start-storybook -p 6006 -c ./component-library/.storybook",
+    "build-storybook": "build-storybook"
```
rowasc: can you add the -c here as well? that way storybook doesn't complain about not finding the configg

package.json
```diff
@@ -53,6 +55,11 @@
     "tabWidth": 4
   },
   "devDependencies": {
+    "@storybook/addon-a11y": "^3.4.8",
```
rowasc: can we have @storybook/cli here as well? that way we don't need a global install for it

component-library/components/molecules/breadcrumbs/breadcrumbs.js
```diff
@@ -0,0 +1,36 @@
+import * as React from "react";
+import PropTypes from "prop-types";
+
+import classnames from "classnames";
+import breadcrumbs from "./breadcrumbs.scss";
+
+const propTypes = {
+    className: PropTypes.string,
+    navigation: PropTypes.shape.isRequired
```
rowasc: const propTypes = {
    className: PropTypes.string,
    navigation: PropTypes.shape({
        path: PropTypes.string.isRequired,
        name: PropTypes.string.isRequired
    }).isRequired
};
note: a shape type needs an object shape

rowasc: no wait I'm wrong because this is an array. Need an array with shapes

rowasc:
```
navigation: PropTypes.arrayOf(
        PropTypes.shape({
            path: PropTypes.string.isRequired,
            name: PropTypes.string.isRequired
        })
    ).isRequired
```

okk like this

PR #1195, opened 2018-07-05
3013 Auto-tick "include" checkbox 

willdoran:
- Changing behaviour of hxl attribute select

Testing checklist:
- [ ] A survey field row that has at least one hxl tag assigned should have the checkbox for the field automatically 'ticked' when the tag has been added

Fixes ushahidi/platform#3013

Ping @ushahidi/platform

### DIFF
--- a/app/settings/data-export/hdx-export.controller.js
+++ b/app/settings/data-export/hdx-export.controller.js
@@ -108,6 +108,7 @@ function (
     }
 
     function selectTag(attribute) {
+        attribute.selected = true;
         if (attribute.selectedTag && attribute.selectedTag.tag_name) {
             attribute.pretty = '#' + attribute.selectedTag.tag_name;
         } else {
--- a/app/settings/data-export/hdx-export.html
+++ b/app/settings/data-export/hdx-export.html
@@ -1,169 +1,170 @@
 <div>
-    <div class="mode-context init" dropdown>
-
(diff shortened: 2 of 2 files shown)


rowasc:
@willdoran much better than where we are right now. Just a small issue / comment: 
- I think you shouldn't be able to de-select a field while it has tags assigned, or at least it should automatically unselect the hxl tags/attributes, right?

willdoran:
@rowasc 
> I think you shouldn't be able to de-select a field while it has tags assigned, or at least it should automatically unselect the hxl tags/attributes, right?

Yeah, I'm not actually sure I thought of that originally but actually I think this way makes more sense. When the field is deselected it won't be exported but while you are on the interface the tagging is preserved.

If you deselected by mistake then you can just reselect if you did all the mapping then changed your mind you can deselect and later reselect and just change the tag. My inclination is to leave it this way because it preserves information that someone went to the trouble of entering without any impact on what is imported and it avoids loss of information if they accidentally unselect a row.

Does that sound reasonable or what do you think?

Erioldoesdesign:
@willdoran @rowasc Reading this Will's solution sounds good to me. So the user has mapped tags/attributes. Unselects (either on purpose or not) and the mapped tags/attributes stay there (maybe grey'd out?) and is not exported/uploaded until the check box to include is selected again.

PR #1196, opened 2018-07-05
Export tags - when a user selects #geo+lat or #geo+lon, they should automatically get the opposite attribute (lat/lon) added

rowasc:
-

Test with https://github.com/ushahidi/platform/pull/3136
Testing checklist:
## Expected behavior (UI)

- [ ] User checks 'Location field'
- [ ] Chooses #geo as the HXL tag
- [ ] Attribute drop-down becomes 'active' (not greyed out)
- [ ] User chooses +lon as the 1st attribute.
- [ ] A second attribute drop-down populated with +lat is automatically added. The tag preview is populated with the 'two' tag&attribute previews shown in the above screenshot. (#geo+lat and #geo+lon)
- [ ] User adds in a third attribute drop down (#geo can also be used with +elevation). The tag preview will populate with #geo+lat+elevation and #geo+lon+elevation (this is the way the csv will display it)

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/settings/data-export/hdx-export.controller.js
+++ b/app/settings/data-export/hdx-export.controller.js
@@ -151,7 +151,7 @@ function (
         let needsMatchLatLon = ignoreMatch;
         if (!!ignoreMatch == false && attribute.selectedTag.tag_name === 'geo' && attribute.selectedHxlAttributes.length < 2) {
             needsMatchLatLon = _.filter(attribute.selectedHxlAttributes, (selected) => {
-                return selected.hxl_tag_id == geoTag.id && (selected.attribute === 'lon' || selected.attribute === 'lat');
+                return selected.attribute === 'lon' || selected.attribute === 'lat';
             }).length;
             needsMatchLatLon = needsMatchLatLon === 1;
         }
@@ -176,11 +176,7 @@ function (
         }
         return attribute;
     }
-
-    function createPrettyAttribute() {
-
-    }
-
+    
     function formatIds() {
         let hxlData = [];
         _.each($scope.forms, (form) => {
--- a/app/settings/data-export/hdx-export.controller.js
+++ b/app/settings/data-export/hdx-export.controller.js
@@ -68,7 +68,7 @@ function (
     function hxlAttributeSelected(hxlAttribute, formAttribute, index) {
         let disabled = false;
         // hack to not disable the value in the current dropdown
-        let selectedHxlAttributes = angular.copy(formAttribute.selectedHxlAttributes);
+        let selectedHxlAttributes = _.toArray(angular.copy(formAttribute.selectedHxlAttributes));
         delete selectedHxlAttributes[index];
         if (_.findWhere(selectedHxlAttributes, {id: hxlAttribute.id})) {
             disabled = true;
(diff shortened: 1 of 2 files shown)


rowasc:
I'm reworking this to not handle #geo+elevation  since we don't need it and will never be an automatically added option for lat+lon as it doesn't make sense. Also found a bug when setting up lat+lon then going for "leave empty" on one of them (it doesn't reset the tags displayed)

app/settings/data-export/hdx-export.controller.js
```diff
@@ -108,23 +108,67 @@ function (
...
     }
 
     function selectHxlAttribute(attribute) {
-        attribute.pretty = '#' + attribute.selectedTag.tag_name;
+        const hxl_attribute = attribute.selectedHxlAttributes[attribute.selectedHxlAttributes.length - 1];
+        if (!needsMatchedAttribute(attribute)) {
+            attribute.hxl_label = createHxlLabel(attribute);
+            return attribute;
+        }
+        // check if we have lat or lon and assign the correct opposite attribute for it
```
willdoran: I have a suggestions that you are free to ignore, drop line 124

and move line 127 - 141 to it's own function called something like addGeoMatchedAttribute()

then at 125 add an else
and put attribute = addGeoMatchedAttribute():
then just return attribute.

If there are other matched attributes later we should consider expanding out to handle them. This is only a suggestion ignore if you hate it.

willdoran: I should say that the naming of the function is actually important so that it can be genrified but I don't know what to call it :/

app/settings/data-export/hdx-export.controller.js
```diff
@@ -108,23 +108,67 @@ function (
...
         }
-        attribute.nbAttributes = 1;
-        attribute.selectedHxlAttributes = {};
+
+        attribute.selectedHxlAttributes = [];
     }
 
     function selectHxlAttribute(attribute) {
-        attribute.pretty = '#' + attribute.selectedTag.tag_name;
+        const hxl_attribute = attribute.selectedHxlAttributes[attribute.selectedHxlAttributes.length - 1];
```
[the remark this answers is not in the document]
rowasc: the var is for lookup and should be considered inmutable yea. Not really a speedup thing, just that I didn't want to reference that gigantic lookup in more than one place :D

app/settings/data-export/hdx-export.controller.js
```diff
@@ -108,23 +108,70 @@ function (
...
+
+        attribute.selectedHxlAttributes = [];
     }
 
     function selectHxlAttribute(attribute) {
-        attribute.pretty = '#' + attribute.selectedTag.tag_name;
+        if (!needsMatchedAttribute(attribute)) {
+            attribute.hxl_label = createHxlLabel(attribute);
+        } else {
+            attribute = addGeoMatchedAttribute(attribute.selectedHxlAttributes[attribute.selectedHxlAttributes.length - 1], attribute);
```
rowasc: uh wait I think I should just grab this attribute.selectedHxlAttributes[attribute in the addGeoMatchedAttribute function

PR #1203, opened 2018-07-10
Changes the data sent to GA to include original url and the actual :id rather than ':id'

kinstelli:
- Rather than sending `/posts/:postId` to GA, this should send `posts/2`

To test:
- [ ] Setup a local install with a test google analytics key
- [ ] Do lots of navigating around the client site, specifically for paths with params that would be matched with Ui-Router's urlMatcher, such as `posts/2`
- [ ] Check that the data logged into GA includes the actual numeric parameters


rjmackay:
@kinstelli any chance you could retest that this works?

PR #1219, opened 2018-07-26
Using Post ID to query for posts created before/after

crcommons:
- when checking for new posts, it now uses the created_after_by_id to work with possible anonymized posts; it also does the same for selected posts but with created_before_by_id

Testing checklist:
- [ ] coming soon!

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/views/post-view-data.directive.js
+++ b/app/main/posts/views/post-view-data.directive.js
@@ -87,7 +87,7 @@ function PostViewDataController(
     /**
      * setting "now" time as utc for new posts filter
      */
-    var newPostsAfter = moment().utc();
+    let timeOfPageLoad = moment().utc();
 
     let unbindFns = [];
 
@@ -145,7 +145,7 @@ function PostViewDataController(
         // Otherwise, go get posts just normal
         if ($scope.selectedPost.post) {
             let query = PostFilters.getQueryParams($scope.filters);
-            query.created_before = $scope.selectedPost.post.created;
+            query.created_before_by_id = $scope.selectedPost.post.id;
             getPosts(query, false);
             $scope.shouldWeRunCheckForNewPosts = false;
         } else {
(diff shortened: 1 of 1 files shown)


app/main/posts/views/post-view-data.directive.js
```diff
@@ -145,7 +145,7 @@ function PostViewDataController(
         // Otherwise, go get posts just normal
         if ($scope.selectedPost.post) {
             let query = PostFilters.getQueryParams($scope.filters);
-            query.created_before = $scope.selectedPost.post.created;
+            query.created_before_by_id = $scope.selectedPost.post.id;
```
willdoran: Can we add a comment above this line, something like:
// As some posts may have their created date redacted by hide_time
// we instead use the Post Id to allow the API to retrieve the date itself

app/main/posts/views/post-view-data.directive.js
```diff
@@ -202,7 +202,7 @@ function PostViewDataController(
         });
 
         if ($scope.shouldWeRunCheckForNewPosts) {
-            checkForNewPosts(30000);
+            checkForNewPosts(5000);
```
willdoran: can we make this everyone minute

app/main/posts/views/post-view-data.directive.js
```diff
@@ -327,18 +327,28 @@ function PostViewDataController(
...
-            // Use the most recent post date as the date to search for new posts since
-            if ($scope.posts.length > 0 && $scope.posts[0].created) {
-                newPostsAfter = moment($scope.posts[0].created).utc().add(1, 's');
-            }
-
             // Merge grouped posts into existing groups
             groupPosts(postsResponse.results);
 
+            // if the selected post isn't in the post list
+
```
willdoran: drop the blank line

PR #1228, opened 2018-08-15
Targeted & Crowdsourced Survey stats in activity view

crcommons:
- Adds view for target survey and crowdsourced survey stats tables in activity view

Testing checklist:
- [ ] go to activity without targeted surveys enabled, you shouldn't see the tables
- [ ] go to activity not logged in, you shouldn't see the tables
- [ ] go to activity with targeted surveys enabled and logged in, tables should appear!
------
while logged in with targeted surveys enabled:
- [ ] go to activity without any surveys or posts, you should see a nice message telling you that you have no posts
-------
now make some surveys and posts, both targeted surveys and regular ones with various data-source types
- [ ] go to activity and see that each table shows the correct data 
- [ ] the totals should add up correctly at the bottom of each table
-------
add some filters
- [ ] see that the loading dots appear while we get new data
- [ ] see that each table now shows the correct data with your date filters applied

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/main/activity/crowdsourced-survey-table.directive.js
+++ b/app/main/activity/crowdsourced-survey-table.directive.js
@@ -14,11 +14,11 @@ function CrowdsourcedSurveyTable() {
 CrowdsourcedSurveyTableController.$inject = ['$scope', '$translate', 'FormEndpoint', 'FormStatsEndpoint', 'FormContactEndpoint', 'PostEndpoint', '_', 'PostFilters', 'Session'];
 function CrowdsourcedSurveyTableController($scope, $translate, FormEndpoint, FormStatsEndpoint, FormContactEndpoint, PostEndpoint, _, PostFilters, Session) {
     $scope.crowdsourcedSurveyStatsByForm = [];
-    $scope.totalAllReduced = 0;
-    $scope.totalWebReduced = 0;
-    $scope.totalEmailReduced = 0;
-    $scope.totalSmsReduced = 0;
-    $scope.totalTwitterReduced = 0;
+    $scope.totalAll = 0;
+    $scope.totalWeb = 0;
+    $scope.totalEmail = 0;
+    $scope.totalSms = 0;
+    $scope.totalTwitter = 0;
     $scope.noSurveys = false;
     $scope.unstructuredStats = {};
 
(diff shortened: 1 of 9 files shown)


app/main/activity/crowdsourced-survey-table.directive.js
```diff
@@ -0,0 +1,97 @@
...
+        template: require('./crowdsourced-survey-table.html')
+    };
+}
+
+CrowdsourcedSurveyTableController.$inject = ['$scope', '$translate', 'FormEndpoint', 'FormStatsEndpoint', 'FormContactEndpoint', 'PostEndpoint', '_', 'PostFilters', 'Session'];
+function CrowdsourcedSurveyTableController($scope, $translate, FormEndpoint, FormStatsEndpoint, FormContactEndpoint, PostEndpoint, _, PostFilters, Session) {
+    $scope.crowdsourcedSurveyStatsByForm = [];
+    $scope.totalAllReduced = 0;
+    $scope.totalWebReduced = 0;
+    $scope.totalEmailReduced = 0;
```
rowasc: maybe a silly question but... why "reduced"?

app/main/activity/crowdsourced-survey-table.directive.js
```diff
@@ -0,0 +1,97 @@
...
+        $scope.totalWebReduced = 0;
+        $scope.totalEmailReduced = 0;
+        $scope.totalSmsReduced = 0;
+        $scope.totalTwitterReduced = 0;
+        $scope.unstructuredStats = {};
+    }
+
+    function getFormStats() {
+        FormEndpoint.query().$promise.then((forms) => {
+            if (crowdsourcedSurveysExist) {
```
rowasc: you can do if (!crowdsourcedSurveysExist) { 
                $scope.noSurveys = true;
return; 
} 
{.... code that was in the crowdsourcedSurveysExist  if .... }
To make it a little easier to read. But also feel free to ignore if you think it's not worth it / disagree

rowasc: also I think it should be crowdsourcedSurveysExist(forms) ? right?

app/main/activity/crowdsourced-survey-table.directive.js
```diff
@@ -0,0 +1,97 @@
...
+                $scope.noSurveys = false;
+                forms.forEach((form) => {
+                    if (!form.targeted_survey) {
+                        let query = PostFilters.getQueryParams($scope.filters) || {};
+                        var postQuery = _.extend({}, query, {
+                            formId: form.id
+                        });
+
+                        FormStatsEndpoint.query(postQuery).$promise.then((stats) => {
+                            stats.name = form.name;
```
rowasc: I think you could do `                            $scope.crowdsourcedSurveyStatsByForm.push(form.name);` directly

app/main/activity/crowdsourced-survey-table.directive.js
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    function getFormStats() {
+        FormEndpoint.query().$promise.then((forms) => {
+            if (crowdsourcedSurveysExist) {
+                $scope.noSurveys = false;
+                forms.forEach((form) => {
+                    if (!form.targeted_survey) {
+                        let query = PostFilters.getQueryParams($scope.filters) || {};
+                        var postQuery = _.extend({}, query, {
```
rowasc: I think postQuery should be using `let` here and take advantage of block level scope, specially because this requests won't necessarily all run before the unstructuredStats section below and can cause issues

app/main/activity/crowdsourced-survey-table.directive.js
```diff
@@ -0,0 +1,97 @@
...
+                            $scope.totalWebReduced += parseInt(stats.total_by_data_source.web);
+                            $scope.totalEmailReduced += parseInt(stats.total_by_data_source.email);
+                            $scope.totalSmsReduced += parseInt(stats.total_by_data_source.sms);
+                            $scope.totalTwitterReduced += parseInt(stats.total_by_data_source.twitter);
+
+                        });
+                    }
+                });
+                let query = PostFilters.getQueryParams($scope.filters) || {};
+                var postQuery = _.extend({}, query, {
```
rowasc: same, might want to use `let`

app/main/activity/crowdsourced-survey-table.directive.js
rowasc: as a side note:  maybe we want to take advantage of promise.all here?

app/main/activity/crowdsourced-survey-table.directive.js
crcommons: Because originally I was going to iterate of the forms and reduce the number for each data source. But then I realized that would require an extra loop. I suppose total and reduced is redundant though. I'll change it :)

app/main/activity/crowdsourced-survey-table.directive.js
crcommons: I need to add the name to the object that I'm pushing to the array, rather than just the name itself.

app/main/activity/crowdsourced-survey-table.directive.js
crcommons: And this is what happens when you copy and paste code from other directives

app/main/activity/crowdsourced-survey-table.directive.js
crcommons: I think the promise.all would be tough because we're not getting back the form name with the stats. So it would be a pain to add the form name to each result after we get them all back.

app/main/activity/crowdsourced-survey-table.directive.js
rowasc: Cool. Ping me when ready :D

app/main/activity/crowdsourced-survey-table.directive.js
rowasc: mmmm.  OK, I think (think!) this would work anyway as long as the `var` is changed `let` for some block scope sanity 👍

PR #1231, opened 2018-08-21
fix(password-reset-confirm-controller) urldecode reset token from URL param 

kinstelli:
- Previously, when a user clicked the reset link, the token that is sent to the API was still urlencoded. However, when the user copies and pastes the reset token into the Reset Password form, it was not.

This should ensure that _all_ tokens sent to the reset controller via POST are urldecoded before they're received.

Testing checklist:
- [ ] Before login, submit a forgot password request.
- [ ] Either from a response email, or directly from the database, enter the token into the Reset Password form that remains after this 'forgot password' request. Confirm that this allows you to reset your password.
- [ ] Submit another forgotten password request.
- [ ] Click the link that is received via email -- note that the token param is urlencoded.
- [ ] Confirm that using this link allows a user to reset the password.

Fixes part of ushahidi/platform#3193 .

Ping @ushahidi/platform


app/common/auth/password-reset-confirm.controller.js
```diff
@@ -3,7 +3,7 @@ module.exports = PasswordResetConfirmController;
 PasswordResetConfirmController.$inject = ['$rootScope', 'PasswordReset', '$location', '$transition$'];
 function PasswordResetConfirmController($rootScope, PasswordReset, $location, $transition$) {
     var $scope = $rootScope.$new();
-    $scope.token = $transition$.params().token;
+    $scope.token = decodeURIComponent($transition$.params().token);
```
rjmackay: I don't understand how this is working in production since it's the same there.
Are you sure ui-router doesn't decode the url component for us?

rowasc:
I was able to use the token directly and also reset my password with the link in production, closing 
- [x] Before login, submit a forgot password request.
- [x] Either from a response email, or directly from the database, enter the token into the Reset Password form that remains after this 'forgot password' request. Confirm that this allows you to reset your password.
- [x] Submit another forgotten password request.
- [x] Click the link that is received via email -- note that the token param is urlencoded.
- [x] Confirm that using this link allows a user to reset the password.

PR #1251, opened 2018-08-31
Show validation messages for users

rowasc:
- shows validation errors to users in the csv import
- Depends on https://github.com/ushahidi/platform/pull/3263
Testing checklist:
- [ ]https://github.com/ushahidi/platform/issues/3260

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/settings/data-import/data-import.directive.js
+++ b/app/settings/data-import/data-import.directive.js
@@ -252,8 +252,11 @@ function (
                 var duplicateVars = checkForDuplicates();
 
                 // third, warn the user which keys have been duplicated
-                if (duplicateVars.length > 0) {
+                if (duplicateVars.length > 0 && duplicateVars[0] !== '') {
                     Notify.error('notify.data_import.duplicate_fields', {duplicates: duplicateVars.join(', ')});
+                    return false;
+                } else if (duplicateVars.length > 0 && duplicateVars[0] === '') {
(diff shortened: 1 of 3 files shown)


app/settings/data-import/data-import.directive.js
```diff
@@ -252,9 +252,12 @@ function (
                 var duplicateVars = checkForDuplicates();
 
                 // third, warn the user which keys have been duplicated
-                if (duplicateVars.length > 0) {
+                if (duplicateVars.length > 0 && duplicateVars.filter((o)=> o === '').length !== duplicateVars.length) {
```
crcommons: Is this filter for duplicateVars that are an empty string and then joining the empty string together? Shouldn't you filter for those that are not an empty string?

rowasc: good question. 
-   duplicateVars.filter((o)=> o === '').length !== duplicateVars.length 
what this is doing is checking if duplicate vars has something other than empty values basically.  ('') 
In that case, we want to tell the user "hey, {value} is duplicated" . 
- If it goes into the else, we are dealing with only '' as the duplicate, which points to empty columns (so ',,,,,' are all dupes, and the user tried to match them  . But anyway this is wrong becaause it could be that you have ',,,,,,pepe,pepe' and in that case it would not be empty. 

Brb need to fix the mess :D

PR #1258, opened 2018-09-11
feat: Add PUBLIC_PATH env option to set public path for assets

rjmackay:
- Add PUBLIC_PATH env option to set public path for assets

Testing checklist:
- [ ] Run `PUBLIC_PATH=http://assets.test.com gulp dist`
- [ ] Check generated index.html refs correct domain
- [ ] Load the app through some http server (`gulp serve:static`) and confirm all assets work

Fixes ushahidi/platform#1610

Ping @ushahidi/platform

### DIFF
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -215,6 +215,7 @@ gulp.task('serve:static', function() {
         ui: false,
         codeSync: false,
         open: false,
+        cors: true,
         middleware: [
           historyApiFallback()
         ]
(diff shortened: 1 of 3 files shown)


rjmackay:
It looks like the way we load SVGs trigger security blocks when loading from another domain. We might have to load the SVGs from the same domain somehow

tuxpiper:
Oh that thing with the svgs is annoying. It will complicate things when deploying.

I was thinking about this again, and I think maybe all we need is having separate folders (i.e. , index.html, static and locales ). That makes rule writing easier in the web server / CDN configs, while still keeping everything under the same domain. 

Created new PR with that idea: https://github.com/ushahidi/platform-client/pull/1262

rjmackay:
Closing in favour of #1262

PR #1262, opened 2018-09-22
Add PUBLIC_PATH env option to set folder for assets 

tuxpiper:
Asserts that the PUBLIC_PATH variable is an absolute path (i.e. '/assets') and configures webpack to drop all the js, css, svgs, etc in a folder with that name. index.html, config.js and locales/ are still placed in the top level folder. 

Fixes ushahidi/platform#1610 .

Ping @ushahidi/platform

### DIFF
--- a/gulpfile.babel.js
+++ b/gulpfile.babel.js
@@ -61,9 +61,22 @@ gulp.task('dist:config', () => {
 
 // Build webpack for production
 gulp.task('dist:webpack', (done) => {
+  // process request for static assets
+  if (process.env.PUBLIC_PATH) {
+    if (process.env.PUBLIC_PATH[0] !== '/') {
+      throw Error('PUBLIC_PATH must begin with /');
+    }
+    if (process.env.PUBLIC_PATH.slice(-1) !== '/') {
+      // ensure public path ends with '/'
+      process.env.PUBLIC_PATH = process.env.PUBLIC_PATH + '/';
+    }
+    // anchor the generation of index.html
+    process.env.INDEX_PATH = paths.dest;
+  }
+  //
   const config = require('./webpack.dist.config');
   config.entry.app = paths.entry;
-  config.output.path = paths.dest;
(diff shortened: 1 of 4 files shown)


tuxpiper:
Added test checklist

gulpfile.babel.js
```diff
@@ -61,9 +61,22 @@ gulp.task('dist:config', () => {
 
 // Build webpack for production
 gulp.task('dist:webpack', (done) => {
+  // process request for static assets
+  if (process.env.PUBLIC_PATH) {
+    if (process.env.PUBLIC_PATH[0] !== '/') {
+      throw Error('PUBLIC_PATH must begin with /');
```
rjmackay: This is weird. Shouldn't public path be able to be a full domain rather than a path?

webpack.config.js
```diff
@@ -72,7 +72,7 @@ module.exports = {
         use: {
           loader: 'svg-url-loader',
           options : {
-            publicPath: '/',
+            publicPath: process.env.PUBLIC_PATH || '/',
```
rjmackay: I think if you do this it'll just try to serve the SVGs from the static domain again, and that'll break because of security checks.

rjmackay:
I'm not sure I quite understand what you're trying to do now?
I thought the idea was to serve all the static files from `static.ushahidi.io` or something similar? I still think thats a good idea, since it allows the cache to be shared across different deployments.

If you also want to move the assets under a subfolder, I think maybe that should be a different option? ie. have ASSET_PATH and ASSET_DOMAIN as different config vars?

tuxpiper:
On second thought, I realised I was going down a more complicated path (pun not intended)

I'm just going to change PUBLIC_PATH to ASSET_DOMAIN , so it's clear what the intended use of the variable is

rowasc:
@tuxpiper this seems like a good candidate for a release soon no? (ofc after retest and doc updates needed :) )

tuxpiper:
ah yep @rowasc  this one is useful in the long term :+1: let's label as something for the next release?

PR #1268, opened 2018-10-09
fix (post-view-data.directive.js): updated total posts count after bulk delete

sinneruy:
- Updates total posts count after bulk delete

Testing checklist:
- [ ] go to views/data
- [ ] use the bulk actions to delete all the available posts
- [ ] the posts count should display 0 of 0 posts
- [ ] the bulk actions button should be disabled

Fixes ushahidi/platform#3324.

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/views/post-view-data.directive.js
+++ b/app/main/posts/views/post-view-data.directive.js
@@ -57,7 +57,7 @@ function PostViewDataController(
     $scope.itemsPerPageOptions = [10, 20, 50];
     $scope.itemsPerPage = $scope.itemsPerPageOptions[1];
     // until we have the correct total_count value from backend request:
-    $scope.totalItems = $scope.itemsPerPage;
+    $scope.totalItems = 0;
     $scope.posts = [];
     $scope.groupedPosts = {};
     $scope.deletePosts = deletePosts;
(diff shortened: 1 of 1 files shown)


sinneruy:
[clabot:check]

app/main/posts/views/post-view-data.directive.js
```diff
@@ -456,7 +457,7 @@ function PostViewDataController(
     function resetPosts() {
         $scope.posts = [];
         $scope.groupedPosts = {};
-        $scope.totalItems = $scope.itemsPerPage;
+        $scope.totalItems = 0;
```
rowasc: @Angamanga do you remember if there's a good reason why we had itemsPerPage (ie what might break with this fix? ) 
It fixes the issue of `0 of 20 posts` being shown when there are 0 posts (it shows 0 of 0 , instead), and it also makes it possible to block the bulk actions when there's zero posts.. but I'd appreciate if you took a look and let us know.

rowasc:
@Angamanga do you mind taking a look?

app/main/posts/views/post-view-data.directive.js
Angamanga: I have really tried to break it but I can't. I suspect it is remains of all the refactors we did on that page when building it? Otherwise it looks good, but found 2 other bugs (not related to this change), will file them separately.

PR #1278, opened 2018-10-31
Fix first load of saved search resetting filters

rjmackay:
- fix: don't trigger doLogout when activating authentication-events
  This was causing a race condition, where if the reset filters didn't run before the saved search loaded, then the saved search filters weren't applied. This wasn't actually needed, since the default state for the scope is logged out and it was only introduced when the rest of the log out logic was moved to Authentication.service startup.
- refactor: Remove unused redirect param from doLogout

Testing checklist:
- [x] Load a deployment
- [x] Navigate to a saved search
- [x] Clear local storage
- [x] Reload the page (with the saved search URL)
- [x] Confirm the saved search filters applied correctly
- [x] Log in
- [x] Log out
- [x] Confirm filters have now been cleared

Run checklist from this issue again https://github.com/ushahidi/platform-client/pull/903

Fixes ushahidi/account-management#64 .

Ping @ushahidi/platform

### DIFF
--- a/app/common/auth/authentication-events.run.js
+++ b/app/common/auth/authentication-events.run.js
@@ -117,17 +117,13 @@ function AuthenticationEvents($rootScope, $location, Authentication, Session, _,
             });
     }
 
-    function doLogout(redirect) {
+    function doLogout() {
         $rootScope.currentUser = null;
         $rootScope.loggedin = false;
         DemoDeploymentService.demoCheck();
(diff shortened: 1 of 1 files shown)


rowasc:
Passed both checklists. Only concern I had was that when I am logged out and go to a saved search, I don't see the saved search filters in the dropdown (as bug icons) but the filters are applied. That's a pre-exisiting issue related to https://github.com/ushahidi/platform/issues/3104, so this should be merged anyway .

PR #1280, opened 2018-11-19
Remove old socket handling code

rjmackay:
- Remove all socket handling code.

Testing checklist:
- [ ] npm run test

Refs https://github.com/ushahidi/platform/pull/3382

Ping @ushahidi/platform


rjmackay:
If we want to bring this back we should consider using https://github.com/tlaverdure/laravel-echo-server and https://github.com/laravel/echo

PR #1297, opened 2019-05-23
fix(sorting-attributes): fix attribute display order according to backend priority

rowasc:
- Fixes issue that caused the data view to render attributes in the wrong order. 

Testing checklist:
- [x] Create a survey with a few different fields , notice the fields order 
- [x] Check the fields are displayed in the correct order in the survey form
- [x] Check the fields are displayed in the correct order in the data view

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/detail/post-detail-data.directive.js
+++ b/app/main/posts/detail/post-detail-data.directive.js
@@ -192,8 +192,9 @@ function PostDetailDataController(
     };
 
     $scope.isPostValue = function (key) {
-        return $scope.form_attributes[key] && $scope.post_task &&
+        let ret =  $scope.form_attributes[key] && $scope.post_task &&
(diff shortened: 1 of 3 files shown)


app/main/posts/detail/post-detail-data.directive.js
```diff
@@ -111,10 +113,17 @@ function PostDetailDataController(
                     .sortBy('priority')
                     .value();
 
-                angular.forEach(attributes, function (attr) {
-                    this[attr.key] = attr;
+                attributes.forEach(attr => {
```
rowasc: ah so I forgot to explain why this happened in the first place. 
We had attributes in an object with their id as the key, so they were not sorted, so what I did was convert that to a sorted array of objects (attributes is sorted a few lines before this) and change the view to use that correctly

PR #1301, opened 2019-06-07
#3555 save button language gets updated when saving the language

rowasc:
- Change language of the SAVE button to the current language. https://github.com/ushahidi/platform/issues/3555

Testing checklist:
- [ ] Go to Settings -> General . Change language to Spanish (or some other language that isn't the current site's language)
- The save button is in Spanish, not English

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/settings/site/editor.directive.js
+++ b/app/settings/site/editor.directive.js
@@ -125,6 +125,8 @@ function (
                         let userLanguage = Session.getSessionDataEntry('language');
                         if ((userLanguage === undefined || userLanguage === null) && $scope.SystemLanguage !== newSystemLanguage) {
                             TranslationService.translate(newSystemLanguage);
+                            $scope.save = $translate.instant('app.save');
+                            $scope.saving = $translate.instant('app.saving');
                         }
                         $scope.SystemLanguage = newSystemLanguage;
                         Notify.notify('notify.general_settings.save_success');


Angamanga:
I don't get it to work, it's still the same language. If I change to 
`<button type="submit" translate>app.save</button>` however, it works.

app/settings/site/editor.directive.js
```diff
@@ -125,6 +125,8 @@ function (
                         let userLanguage = Session.getSessionDataEntry('language');
                         if ((userLanguage === undefined || userLanguage === null) && $scope.SystemLanguage !== newSystemLanguage) {
```
Angamanga: @rowasc I find this very confusing, if someone has selected a language in the language-switcher from map-view, they wont see any changes made in the settings-view if they change to another language. Could we remove the check entirely and each time a user switch language here, it actually change the language they see in the view? I know its not covered by this issue, but it confused me a lot while testing 🤔

app/settings/site/editor.directive.js
```diff
@@ -125,6 +125,8 @@ function (
...
                             TranslationService.translate(newSystemLanguage);
+                            $scope.save = $translate.instant('app.save');
+                            $scope.saving = $translate.instant('app.saving');
                         }
```
Angamanga: This + row 45-46 can be removed completely and changed to translation directly in the html-file :) 👍

PR #1302, opened 2019-06-09
fix: 2372-platform no more button in filters

rowasc:
- Remove the "more" overflow button in the filters dropdown per https://github.com/ushahidi/platform/issues/2372

Testing checklist:
- [x] Add many surveys, categories, and posts for all statuses (or use rominacsvsms.ushahidi.io as the backend) 
- [x] Check that the filters dropDown does not have a "more" button under each, instead they all expand without user action
See gif of expected results here. 
no-overlflow

Fixes ushahidi/platform# .

Ping @ushahidi/platform


rowasc:
@Angamanga this is a tiny fix to avoid an extra click

PR #1315, opened 2019-07-20
hotfix(video-url): fix issue where a blank video url caused the video component to crash

rowasc:
- If a video url field is empty, we no longer attempt to embed a video

Testing checklist:
- [ ] Create a survey with an optional video field
- [ ] Submit two posts. One with a valid video URL value,  and one with an empty video url.
- [ ] Verify both posts load the field correctly (one shows a video, the other is empty) 

Fixes ushahidi/platform#3610 .
Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/detail/post-video-value.directive.js
+++ b/app/main/posts/detail/post-video-value.directive.js
@@ -26,6 +26,8 @@ function PostVideoValueController(
     activate();
 
     function activate() {
-        $scope.videoUrl = $sce.trustAsResourceUrl($scope.videoUrl[0]);
+        if ($scope.videoUrl && $scope.videoUrl.length > 0) {
+            $scope.videoUrl = $sce.trustAsResourceUrl($scope.videoUrl[0]);
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


app/main/posts/detail/post-video-value.directive.js
```diff
@@ -26,6 +26,8 @@ function PostVideoValueController(
     activate();
 
     function activate() {
-        $scope.videoUrl = $sce.trustAsResourceUrl($scope.videoUrl[0]);
+        if ($scope.videoUrl && $scope.videoUrl.length > 0) {
```
rowasc: if this is not the case, we just then let the videoUrl continue as undefined / empty so it doesn't render

app/main/posts/detail/post-video-value.html
```diff
@@ -2,7 +2,7 @@
   <h2 class="form-label">
     <bdi>{{label}}</bdi>
   </h2>
-  <div class="video_embed-fluid">
+  <div class="video_embed-fluid" ng-if="videoUrl">
```
rowasc: so, this was basically causing an issue where you'd see an attempt to load platform, instead of just an empty field

PR #1316, opened 2019-08-02
adding eslint-file (translated from jscrc and jshint-file) #3587

Angamanga:
- Translates our current linter-rules for jscs and jshint to eslint

Fixes ushahidi/platform#3587 (part of) .

Ping @ushahidi/platform

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,109 @@
+{
+    "env": {
+      "browser": true,
+      "node": true,
+      "es6": true
+    },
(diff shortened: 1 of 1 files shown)


Angamanga:
@rowasc Have you had a chance to test this? Can we merge it?

rowasc:
Worked fine for me. I think we might want to move into prettier for easier formatting TBH but this is a good start to migrate later if we want it , too

PR #1322, opened 2019-08-21
Adding oauth check

Angamanga:
- Adds status-check for endpoint oauth/token
- Adds structure-check for the response from oauth/token
- Uses the received token for a request, checks if it works
- Refactors the code to reuse bits and pieces

Testing checklist:
- [ ] Run gulp verify
- [ ] You should see a status-result for /oauth/token
- [ ] You should see a structure-result for /oauth/token
- [ ] You should see a result for a token-test

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/gulp/verifier.js
+++ b/gulp/verifier.js
@@ -31,14 +31,14 @@ module.exports.verifyEnv = function() {
     if (!process.env.BACKEND_URL) {
         log.error(
         c.red('BACKEND_URL not found in .env file. ' +
-                'Please add this URL to the .env file to connect to the Ushahidi platform server.'
+                'Please add this URL to the .env file to connect to the Platform API.'
             )
         );
     }
     if (process.env.BACKEND_URL && !isUrl(process.env.BACKEND_URL)) {
         log.error(
         c.red('BACKEND_URL found in .env file. Is not a valid URL.' +
-                'Please fix the API endpoint URL in the .env file to connect to the Ushahidi platform server.'
+                'Please fix the BACKEND_URL in the .env file to connect to the Platform API.'
             )
         );
     }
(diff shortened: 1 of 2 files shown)


gulp/verifier.js
```diff
@@ -174,10 +206,43 @@ const isCheckDisabled = function(name) {
     return checks.indexOf(name) >= 0;
 };
 
-const compareKeys = function(a, b) {
+const checkStructure = function(a, b, url) {
     const aKeys = Object.keys(a).sort();
     const bKeys = Object.keys(b).sort();
-    return JSON.stringify(aKeys) === JSON.stringify(bKeys);
+    log.info(c.bold(`Structure-result for ${url}:`));
+    if (JSON.stringify(aKeys) === JSON.stringify(bKeys)) {
```
rowasc: this is good. One caveat is that maybe we want to check that all the keys are in there, but not that the stringify is equal, in case the order of the keys change at some point or something.

gulp/verifier.js
```diff
@@ -166,6 +114,90 @@ module.exports.verifyEndpointStructure = function() {
...
+                log.eror(c.red('Please check storage/logs in the platform-api, and see what the API logs say about this error.'));
+                break;
+            case '401':
+                log.error(c.red('Did you add your own client_secret and client_key?'));
+                log.error(c.red('Make sure you updated the database with the same values'));
+                log.error(c.red('If you did not add your own secret and key, check that the database-migration has ran.'));
+                break;
+            case '403':
+                log.error(c.red('There is a problem with the oauth-scopes'));
+                log.error(c.red('Check that your api is set up and that all migrations has ran.'));
```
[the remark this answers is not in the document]
Angamanga: :) But ofcourse!

gulp/verifier.js
Angamanga: Mm, yes that is true, it may happen... Fixing...

Angamanga: But they are sorted before, so should be ok?

Angamanga:
@rowasc Have updated with your comments!

gulp/verifier.js
rowasc: oh. Wow I'm dumb.
You are absolutely right, sorry

PR #1323, opened 2019-08-29
Browser installer helper

Angamanga:
- Adds basic FE-verifier with Angular, go to /verifier to see.
- Moves all verifier-tests to a specific file, that can be used both for the gulp-verifiers and browser-verifiers.
- Refactors the gulp-verifier to use the specific verifier-test-file instead
- Adds missing checks for dbConnection and env in the api (needs branch `db-installer-helper` in the api)
- This works fine as long as there are no errors :D If there are errors, the "Sorry something went wrong" message takes over.

@rowasc What I need help with/want to pair on:
I am not sure if this is the best strategy for the browser-tests. I am having problems sidetracking the "Sorry something went wrong" for this particular route (/verifier). Do you have any other ideas? I experimented a bit with other solutions like doing a plain js-file on its own route but never managed to get anything useful working... Also general sanity-check/review on the code would be appreciated! 🙏 

Testing checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/common/verifier/verifier.js
+++ b/app/common/verifier/verifier.js
@@ -104,7 +104,8 @@ const verifyEndpointStructure = function (env) {
     return Promise.all(requests)
     .then(function (responses) {
         return responses.map(function (response) {
-            return response.json().then(function (jsonData) {
+            return response.json()
+            .then(function (jsonData) {
                 let structure = {};
                 switch (response.url.substring(response.url.lastIndexOf('/') + 1)) {
                     case 'tags':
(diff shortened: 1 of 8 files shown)


app/common/verifier/verifier.js
```diff
@@ -0,0 +1,256 @@
...
+                };
+            return results;
+        });
+};
+
+const verifyEnv = function (env) {
+    let messages = [];
+    try {
+        // fs is not working in the browser, how handle it?
+        fs.accessSync('.env');
```
Angamanga: fs is not working in the browser, but waiting for us to solve the main-problem with where to put the browser-verifier before finding a solution for it.

PR #1326, opened 2019-08-30
Installer web checks

Angamanga:

- Adds FE-verifier with Angular, go to /verifier to see.
- Moves all verifier-tests to a specific file, that can be used both for the gulp-verifiers and browser-verifiers.
-Refactors the gulp-verifier to use the specific verifier-test-file instead
-Adds missing checks for dbConnection and env in the api (needs branch db-installer-helper in the api)

Testing checklist:
- in the cli, run "gulp verify"
- [x] A list with checks and statuses is listed
- run gulp dev:verifier
- go to /verifier
- [x] A list with checks and statuses is listed
- Destroy something, like stopping the database or remove the BACKEND_URL from the .env-file
- [x] Reload, the browser, it should show the checks, with errors
- Go to the cli again, run gulp verify
- [ ] It should list the same errors as in the browser

 I certify that I ran my checklist
Ping @ushahidi/platform

### DIFF
--- a/app/common/verifier/verifier.js
+++ b/app/common/verifier/verifier.js
@@ -161,14 +161,14 @@ const verifyOauth = function (env) {
 
 const verifyAPIEnvs = function (env) {
     if (isCheckDisabled(env, 'API_ENVS')) {
-        return;
+        return new Promise((resolve, reject) => resolve('DISABLED'));
     }
     return checkAPI(`${env.BACKEND_URL}/api/v3/verifier/env`);
 };
 
 const verifyDbConnection = function (env) {
     if (isCheckDisabled(env, 'DB_CONNECTION')) {
-        return;
+        return new Promise((resolve, reject) => resolve('DISABLED'));
     }
     return checkAPI(`${env.BACKEND_URL}/api/v3/verifier/db`);
 };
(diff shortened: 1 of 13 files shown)


rowasc:
@Angamanga I sent a few fixes, would you mind taking a look? 
The only difference I was able to spot with outputs is that the frontend gives less detail
This error in the CLI
> [15:54:03] FetchError: invalid json response body at http://192.168.33.110/api/v3/forms reason: Unexpected token < in JSON at position 0
Results in only this error in the web client 
>SyntaxError: Unexpected token < in JSON at position 0


🤔

webpack.config.js
```diff
@@ -99,7 +99,9 @@ module.exports = {
     new webpack.DefinePlugin({
         BACKEND_URL: JSON.stringify(process.env.BACKEND_URL || 'http://backend.url.undefined'),
         ENVIRONMENT: JSON.stringify(process.env.ENVIRONMENT || 'dev'),
-        GIT_COMMIT: JSON.stringify(GIT_COMMIT || false)
+        VERIFIER: JSON.stringify(process.env.VERIFIER || false),
+        GIT_COMMIT: JSON.stringify(GIT_COMMIT || false),
+        USH_DISABLE_CHECKS: JSON.stringify(process.env.USH_DISABLE_CHECKS) || ' '
```
rowasc: curious, why do we need the empty space there? (I'm probably missing something obvious sorry) 😬

Angamanga: Parsing gets wierd in app.js with an empty string there, I now changed to false instead.

app/common/verifier/verifier.js
```diff
@@ -273,8 +273,14 @@ const isCheckDisabled = function (env, name) {
 };
 
 const checkStructure = function (a, b, url) {
+    delete a.$promise;
+    delete a.$resolved;
+    delete b.$promise;
+    delete a.$resolved;
```
Angamanga: @rowasc Ugh this looks not so nice, do you have any other idea?

rowasc: underscore's _filter maybe ?

PR #1328, opened 2019-09-03
checking if post is a tweet

Angamanga:
- Checks if a post really is a tweet before showing the twitter-widget.

Testing checklist:
- [ ] Go to a deployment with tweets and emails
- [ ] Emails should show up as emails, without a tweet embedded (check many emails, this only happens if the message_id happens to correspond with an existing tweet)
- [ ] Tweets should show up as tweets.

Fixes ushahidi/platform#3438 .

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/views/card.html
+++ b/app/main/posts/views/card.html
@@ -14,7 +14,7 @@
             <div class="postcard-title">
                 <div class="postcard-field">
                 <div class="tweet-container">
-                  <twitter-widget ng-if="post.data_source_message_id" twitter-widget-id="post.data_source_message_id">
+                  <twitter-widget ng-if="post.data_source_message_id && post.source === 'twitter'" twitter-widget-id="post.data_source_message_id">
                   </twitter-widget>
                 </div>
 


app/main/posts/views/card.html
```diff
@@ -14,7 +14,7 @@
             <div class="postcard-title">
                 <div class="postcard-field">
                 <div class="tweet-container">
-                  <twitter-widget ng-if="post.data_source_message_id" twitter-widget-id="post.data_source_message_id">
+                  <twitter-widget ng-if="post.data_source_message_id && post.source === 'twitter'" twitter-widget-id="post.data_source_message_id">
                   </twitter-widget>
```
Angamanga: Keeping the message_id check, because if no message_id we cannot render the tweet anyway...

rowasc:
Nice. testing it  after meetings 🎉  looks good

rowasc:
Ugh just to be clear I have like 3hs scheduled in a row and then a few urgent issues... so you will see me testing this after that...

Angamanga:
@rowasc I tested in steve-buscemi and tweets show up. In order to check for emails, I need help to alter the db and change a data_source_message-id to something that corresponds with a tweet (for example 1163536651261534210). I tested again locally and it works there on develop, tweets only show up when the data_source_message_id corresponds to a tweet and the type is tweet, emails only show the contents of the email.

Angamanga:
@rowasc @tuxpiper Helped me changing in the db and confirming that this is now working as expected! 💃

rowasc:
excellent, thank you !

PR #1335, opened 2019-09-13
adding notification-bar with warning if debug is switch on

Angamanga:
- Adds a notify-bar if debug-mode is switched-on in the backend

Testing checklist:
- [ ] Disable debug-mode in the backend (composer installdebug:disable)
- [ ] Reload client, a demo-bar should not appear
- [ ] Enable debug-mode in the backend (composer installdebug:enable)
- [ ] Reload client, a demo-bar should appear with a warning.

NOTE This only shows up at login + at reload if user already is logged in and the bar can be dismissed. Do we want to always show it?

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -172,12 +172,6 @@ angular.module('app',
         angular.element(document.getElementById('bootstrap-loading')).addClass('hidden');
         DemoDeploymentService.demoCheck();
     }])
-    .run(['Verifier','Notify', function (Verifier, Notify) {
-        Verifier.checkDebugMode(BACKEND_URL)
-        .then(function (result) {
-            if (result) {
-                Notify.notifyPermanent(`You have debug-mode switched on, for security reasons, we recommend NOT leaving this check routine enabled in the API. 
-                    You may disable the check by running the "composer installdebug:disable" command in the API folder.`);
-            }
-        });
+    .run(['VerifierService', function (VerifierService) {
+        VerifierService.debugModeCheck();
     }]);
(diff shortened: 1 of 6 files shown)


tuxpiper:
I was thinking, maybe it would be better to show this only to logged in users (dare I even say, users of the admin group?) . Otherwise, it may just annoy most site visitors without really a point.

Angamanga:
@tuxpiper Yep, that makes sense, added that now.

rowasc:
@Angamanga @tuxpiper I think it's probably necessary to show it to non-logged in users, or many deployers won't notice it as they may not log in at all when they do the setup.

tuxpiper:
> I think it's probably necessary to show it to non-logged in users, or many deployers won't notice it as they may not log in at all when they do the initial setup.

That also sounds reasonable to me 🤷‍♂  . Although in that case, I think it would be an improvement to change the wording, taking into consideration that not everyone that sees the banner may be able to take action on it.

> (that however is also a problem, because we need to make them login somehow to change their password!!)

:) maybe another check?

rowasc:
>:) maybe another check?
Totally :)

rowasc:
re extra login check, ideally what I would _like_ to see is a forced password change in the backend installation (ie after migration) + this warning which would only show up if you somehow manage to skip that password change . Creating a ticket for it since I'm still editing videos and don't want to forget ha :D

Angamanga:
@rowasc @tuxpiper Updated to remove the admin-check + changed the language to make it clear its for admins.

rowasc:
@Angamanga I think a few tests are failing, could you check ?

rowasc:
Nice 🎉 . I think it can be merged since it passed the tests now :)

PR #1352, opened 2019-10-23
Fixing debug-mode-check

Angamanga:
- Uses Util.apiUrl instead of BACKEND_URL to check if debug-mode is on.

Testing checklist:
- [ ] 

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/common/verifier/verifier.js
+++ b/app/common/verifier/verifier.js
@@ -231,7 +231,7 @@ const checkTokenStructure = function (env, options) {
 };
 
 const checkDebugMode = function (url) {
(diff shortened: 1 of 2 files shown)


Angamanga:
@rowasc In the controller you mean? We send the whole CONST object in the controller, not only the backend-url. The verifier-code itself is outside of Angular. There is no way to to get the whole CONST-object from Util, but I could add it, it would definitely be prettier...

rowasc:
no, I mean just applying the same logic to the db check and env check endpoint calls, but this looks good and ready to merge for the scope of the fix now anyway

PR #1362, opened 2019-11-15
Fix for 'There's a random hover state for the time metadata' #2277

AlanaFarkas:
**Expected behaviour**
hover state on postcard should show full date and time

**Actual behaviour**

image

**New behaviour**

image

- Add aria-label for accessibility
- Change font color slightly for better contrast
- Now displays full date and time in tooltip on hover

Testing checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/sass/overrides/_modal.scss
+++ b/sass/overrides/_modal.scss
@@ -12,6 +12,4 @@ modal {
     }
 }
 
-.tooltip-visible-el {
-    color: #FEFEFE;
-}
+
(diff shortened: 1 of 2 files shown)


AlanaFarkas:
[clabot:check]

sass/overrides/_modal.scss
```diff
@@ -11,3 +11,7 @@ modal {
         pointer-events: none;
     }
 }
+
+.tooltip-visible-el {
```
rowasc: I think this may be better moved to the pattern library, but if you remove this change from here,  I will approve the PR since it works so much better now 🎉 thank you so much for this fix

AlanaFarkas: 👍

PR #1382, opened 2019-11-28
3781 validating img

Angamanga:
- connects the imageId to ngModel in post-media-directive (in order to check if an image is there or not)
- adds a very basic validation on mime-type and size

Testing checklist:
-  Start with a survey with 1 image, make the image-field required
- Add a post with an image with a size < 1 MB
- [ ] It should display the image
- Go to data-view or map-view
- [ ] The image should be visible
- Click on edit. The image should be visible
- [ ] Click on delete
- [ ] There should be a warning, click delete again
- [ ] The image should disappear
- [ ] Click on "add new"
- Select an image < 1 MB
- [ ] It should display the image
-  Click on "replace image"
- [ ] Select an image with a size > 1MB
- [ ] The image should not appear
- [ ] There should be a notification-bar telling you that the image was too  big
- Select an image < 1 MB
- [ ] It should display the image
- Click on save
- [ ] It should save
- [ ] Add a post with an image with a size > 1MB
- [ ] The image should not appear
- [ ] There should be a notification-bar telling you that the image was too  big
- [ ] There should be an error-label below the field saying this field is required
- [ ] Add an image < 1MB, the error-label should disappear

- Repeat with a survey with 2 or more fields

NOTE: This is the first step in fixing the image-upload in the client
Things to do are listed in #3781

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/common/directives/file-upload.directive.js
+++ b/app/common/directives/file-upload.directive.js
@@ -7,29 +7,42 @@ function FileUpload() {
         replace: true,
         scope: {
             container: '=',
-            model: '='
+            model: '=',
+            validation: '='
         },
         controller: [
-            '$scope', '$attrs',
+            '$scope', '$attrs', 'Notify',
             function (
-                $scope, $attrs
+                $scope, $attrs, Notify
             ) {
                 $scope.required = typeof $attrs.required !== 'undefined';
                 $scope.uploadFile = function ($event) {
-                    $scope.container.file = $event.target.files[0];
-                    var reader = new FileReader();
-                    reader.onload = function () {
-                        var dataURL = reader.result;
-                        $scope.container.dataURI = dataURL;
-                        $scope.container.changed = true;
-                        $scope.container.deleted = false;
-                        $scope.model = 'changed';
-                        $scope.$apply();
-                    };
-                    reader.readAsDataURL($event.target.files[0]);
-
-
+                    if (validateFile($event.target.files[0])) {
+                        $scope.container.file = $event.target.files[0];
+                        var reader = new FileReader();
+                        reader.onload = function () {
+                            var dataURL = reader.result;
+                            $scope.container.dataURI = dataURL;
+                            $scope.container.changed = true;
+                            $scope.container.deleted = false;
+                            $scope.model = 'changed';
+                            $scope.$apply();
+                        };
+                        reader.readAsDataURL($event.target.files[0]);
+                    } else {
+                        Notify.error('post.media.error_in_upload');
+                    }
                 };
+
+                function validateFile(container) {
+                    if ($scope.validation === 'image') {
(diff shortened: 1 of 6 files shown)


app/common/directives/file-upload.directive.js
```diff
@@ -7,29 +7,42 @@ function FileUpload() {
...
+                    } else {
+                        Notify.error('post.media.error_in_upload');
+                    }
                 };
+
+                function validateFile(container) {
+                    if ($scope.validation === 'image') {
+                        var mimeReg = /[\/.](gif|jpg|jpeg|png)$/i;
+                        var mimeCheck = mimeReg.test(container.type);
+                        var sizeCheck = container.size < 1000000;
```
rowasc: 1048576 this is the currently set size in the API. 

Also thinking about it, it's actually really problematic that we don't have a  media_max_size coming from the config itself in features, because if we change the API for an environment this has to be changed as well. 
I think it's fine for this specific fix but for the next version you are working on, I'd suggest adding this as a /config/features item as well, If you agree (I can probably add it quickly in the backend)

app/common/locales/en.json
```diff
@@ -840,7 +840,8 @@
             "upload" : "Upload",
             "add_photo": "Add photo",
             "delete_photo": "Delete photo",
-            "replace_image": "Replace image"
+            "replace_image": "Replace image",
+            "error_in_upload": "Your image is too large or has the wrong format, please check and try again!"
```
rowasc: can you provide information about the max image size and the supported formats?

app/main/posts/modify/post-data-editor.directive.js
```diff
@@ -242,10 +242,6 @@ function PostDataEditorController(
             attributes.map(function (attr) {
                 // Create associated media entity
                 if (attr.input === 'upload') {
-                    var media = {};
```
rowasc: why is this being removed? (just curious, I think I am a little lost 😬 ) is this now handled with the viewChange /

app/common/directives/file-upload.directive.js
Angamanga: Yes, i think that is a good idea to fetch it from the backend, lets work on that next?

app/main/posts/modify/post-data-editor.directive.js
Angamanga: This was not used, made the conclusion its forgotten remains from a previous refactoring.

app/common/locales/en.json
Angamanga: Sent a fix for this!

app/common/directives/file-upload.directive.js
rowasc: excellent

PR #1383, opened 2019-11-29
3781 validating img

Angamanga:
- Moves the deletion of an image to the save-step instead of instant deletion. It prevents the scenario "if a user deletes an image, and then change their mind and leave without saving, their image is still deleted"

Testing checklist:
- Select a post with an image
- Delete the image
- Do not save the post, navigate away or click cancel
- [ ] Go to the post again, the image should still be there
- Select the post again and edit it
- Delete the image
- Save the post
- [ ] Go to the post again, the image is deleted!
- [ ] Repeat above with a survey with several images

Fixes ushahidi/platform#3781 (part of it) .

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/modify/post-media.directive.js
+++ b/app/main/posts/modify/post-media.directive.js
@@ -25,7 +25,7 @@ function (
         template: require('./media.html'),
         link: function ($scope, element, attr, ngModel) {
             // Initialize media object
-            $scope.media = { id: null, file: null, caption: '', dataURI: null, changed: false};
+            $scope.media = { id: null, file: null, caption: '', dataURI: null, changed: false, deleted: false };
             $scope.mediaId = null;
             $scope.showAdd = showAdd;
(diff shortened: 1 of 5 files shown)


app/main/posts/modify/post-media.directive.js
```diff
@@ -25,7 +25,7 @@ function (
         template: require('./media.html'),
         link: function ($scope, element, attr, ngModel) {
             // Initialize media object
-            $scope.media = { id: null, file: null, caption: '', dataURI: null, changed: false };
+            $scope.media = { id: null, file: null, caption: '', dataURI: null, changed: false};
```
rowasc: should this have a `deleted: false` as well?

Angamanga: Yep, true, they should all be there for clarity. Added it now :)

PR #1388, opened 2019-12-06
3781 delete image

Angamanga:
- Detects changes on media.deleted. This will make the viewValue update if both deleted and deleted and then replaced
- Sends media-id for deletion if an image is deleted.

Testing checklist:
- Delete an image in a survey with required images
- [ ] You should immediately get a red box that a required field is missing
- [ ] Try saving, you should immediately get a message that a required field is missing
- Add a new image
- Delete the new image, 
- [ ] You should immediately get a red box that a required field is missing
- [ ] Try saving, you should immediately get a message that a required field is missing

- [x ] I certify that I ran my checklist

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/modify/post-media.directive.js
+++ b/app/main/posts/modify/post-media.directive.js
@@ -104,9 +104,7 @@ function (
 
             function deleteMedia(mediaId) {
                 // Mark for deletion
-                $scope.media = {id: mediaId};
-                $scope.media.changed = true;
-                $scope.media.deleted = true;
+                $scope.media = {id: mediaId, changed: true, deleted: true};
             }
         }
     };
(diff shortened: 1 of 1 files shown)


app/main/posts/modify/post-media.directive.js
```diff
@@ -91,7 +104,7 @@ function (
 
             function deleteMedia(mediaId) {
                 // Mark for deletion
-                $scope.media = {};
+                $scope.media = {id: mediaId};
```
rowasc: not a blocker if you can't do it now, but could you move changed and deleted to inside this declaration? since they are the next two lines?

Angamanga: Heh, yes, i'll fix it now

PR #1409, opened 2020-01-10
fix(media-all): only call media.get if there is an id

rowasc:
-

Testing checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/detail/post-media-value.directive.js
+++ b/app/main/posts/detail/post-media-value.directive.js
@@ -13,6 +13,7 @@ module.exports = ['MediaEndpoint', '_', function (MediaEndpoint, _) {
 
     function MediaValueLink($scope) {
         function loadMedia() {
+            $scope.mediaLoaded = true;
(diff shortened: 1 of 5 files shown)


Angamanga:
@Obadha2 This fix has no issue so adding the QA-checklist here:
- Add a post to a survey with images, but leave the image-field empty (no image)
- Save the post
- Go to dataview and look at the post
- [ ] The post should load without any 404-responses from the /media endpoint (you can check this in the inspector, 'network'-tab and filter on "media": 
Screenshot 2020-01-16 at 10 01 14
- Select edit for this post
- [ ] Still no 404-responses from the /media endpoint (you can check this in the inspector, 'network'-tab and filter on "media": 
Screenshot 2020-01-16 at 10 01 14
- Add an image
- [ ] The image should load
- Save the post
- [ ] The image should load when viewing the post

Obadha2:
QA'd, passes. 👍

PR #1412, opened 2020-01-14
Fix bug that generated empty HDX tag ids

rowasc:
-

Testing checklist:
- Works together with https://github.com/ushahidi/platform/pull/3831

Fixes ushahidi/platform# .

Ping @ushahidi/platform


rowasc:
sending this for review. WIll need help from Walter when we start testing  to write more tests, since the test suites for this module are pretty incomplete apparently? 
Depends on a backend PR in https://github.com/ushahidi/platform/pull/3831

Angamanga:
Copying over test-checklist from backend-pr:
Test checklist:

- [ ] I can setup my API key and userrname for HDX
- [ ] I can change my API key and username
- [ ] I can upload all my data to HDX
- [ ] I can select only some fields to upload to HDX
- [ ] I can select tags only (and it generates the CSV+HXL file correctly with the correct tags)
- [ ] I can select tags and attributes (and it generates the CSV+HXL file correctly with the correct tags)
- [ ] I can see the HDX exports in the export history tab
- [ ] I get notified when an HDX upload succeeds
- [ ] I can download a CSV+HXL file instead of uploading to HDX
- [ ] I can leave the page while the CSV and HDX processes run

Angamanga:
@rowasc The code looks good, but there are things I found, that might have to do with that I am missing some step (like trigger the upload to hdx)?: 
- (probably not related to this change) The notification-bar after saving the api-key stays in the tagging-view. Does not disappear after clicking "Start tagging data" or "Got it".
- It says "Invalid date" in the export-history for exports to hdx. 
Screenshot 2020-01-28 at 11 48 07
- There is no expire-date for any export
- My data does not turn up on HDX (do i need to trigger the upload-job in some way? Doing this locally)

rowasc:
### Not related to this (pre existing issues):
- The notification-bar after saving the api-key stays in the tagging-view. Does not disappear after clicking "Start tagging data" or "Got it".

### This is either new or I didn't see it when I tried so I'll check it out
- There is no expire-date for any export
- It says "Invalid date" in the export-history for exports to hdx.

### Misconfig on your account maybe?*
- My data does not turn up on HDX (do i need to trigger the upload-job in some way? Doing this locally)

### Normal or expected
There doesn't say anywhere that the row where the hdx-tags are is actually hdx-tags?

rowasc:
@Angamanga merging since we discussed that the expire issue was pre existing , and checking that we have issues for those already (or creating them to address them ASAP if not)

PR #1414, opened 2020-01-16
adding back correct verifier-task

Angamanga:
- Adds back verifier-task that was changed by mistake

Testing checklist:
- Run `gulp dev:verifier` from the command line
- [ ] It should start a dev server
- Navigate to /verifier
- [ ] The verifier-view should load

Fixes ushahidi/platform# .

Ping @ushahidi/platform


Angamanga:
@rowasc I’m adding back to the documented task. For some reason my brain malfunctioned and added watch:verifier instead of dev:verifier when updating gulp. watch:verifier was never documented/used. Found out when the user in community-support couldn’t use dev:verifier...

rowasc:
thank you <3 super excited that this got solved so fast, and before hitting prod

PR #1416, opened 2020-01-22
Adding Accept-Language-header to interceptor

Angamanga:
-Adding Accept-Language header to send the user-selected language to Auth-interceptor

Testing checklist:
- [ ] Switch language to Spanish (Spanish is the only translations we have right now)
- [ ] Try signing up with an email address already known
- [ ] The Error-message should be in Spanish
- [ ] GO to settings-categories
- [ ] Add a category with the same name as one that already exists
- [ ] The error message (or part of it, its not all translated yet) should be in Spanish
- [ ] Any other error-messages showing up as the error-toast-bar should be translated (it there are corresponding translations in the backend)

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/common/auth/authentication-interceptor.config.js
+++ b/app/common/auth/authentication-interceptor.config.js
@@ -85,6 +85,8 @@ function AuthInterceptor($rootScope, $injector, $q, CONST, Session, _) {
         var deferred = $q.defer();
 
         config.ignorable = shouldIgnoreAuthError(config);
+        // Revisit and possibly move out to own inceptor
+        config.headers['Accept-Language'] = Session.getSessionDataEntry('language') ? Session.getSessionDataEntry('language') : 'en-US';
 
         if (config.url.indexOf(CONST.API_URL) === -1) {
             deferred.resolve(config);


rowasc:
Lets get this into the release since David will need it, and make sure Walter tests it? @Angamanga 
this should be the only extra change going into it except for the bugfix you are doing for filters , sorry 😬 !

Angamanga:
@rowasc Yep, plus the backend-changes, right?

rowasc:
yep sending both

Angamanga:
@Obadha2 This will be included in the release too, could you test in Steve Buscemi? 🎉 🙏

tuxpiper:
just a heads up : I've had trouble testing this with firefox on my localhost , the header was not changed after switching languages.

will give it another try tomorrow with other browsers / steve-buscemi

tuxpiper:
mmm ok nevermind about my previous comment. I must have been too tired.

Obadha2:
@Angamanga while running smoke tests, i noticed the following: 
From an old account, when I change language from General settings, language changes don't come into effect when I click Save, I have to log out of the account. When logged out, then the new language changes are seen.
However, I created a new account. When I change language from General Settings in a newly created account, when I click Save, language changes are seen immediately.

tuxpiper:
> @Angamanga while running smoke tests, i noticed the following:
> From an old account, when I change language from General settings, language changes don't come into effect when I click Save, I have to log out of the account. When logged out, then the new language changes are seen.
> However, I created a new account. When I change language from General Settings in a newly created account, when I click Save, language changes are seen immediately.

I think this more or less matches the expected behavior I read from the code: if the user account has a language preference, that should always take precedence over the language chosen in "General settings".  (I'm assuming the "old account" had some language preference saved in the database)

Angamanga:
@tuxpiper @Obadha2 Yep, that is the expected behaviour right now at least, but it is confusing. There is a wip pr that deals with it, but its abandoned (by me) right now. Plan to take up work on it.

PR #1434, opened 2020-02-20
Reset zoom feature added (fixes #3848)

rohit645:
- UI enhancement.
- Adds the reset zoom button. As user clicks on the reset zoom button map is set to default center [latitude, longitude] and zoom level.

Testing checklist:
- [x] Verified code style using jscs
- [x] Verified eslint checklist 

Fixes ushahidi/platform#3848 .

**Demo**
resetZoom

### DIFF
--- a/app/common/services/maps.js
+++ b/app/common/services/maps.js
@@ -39,13 +39,15 @@ function Maps(ConfigEndpoint, L, _, CONST) {
         pointToLayer: pointToLayer,
         getConfig: getConfig,
         getLayer: getLayer,
-        pointIcon: pointIcon
+        pointIcon: pointIcon,
+        setDefaultView: setDefaultView,
+        defaultConfig: defaultConfig
     };
 
     function createMap(element) {
         return getLeafletConfig().then(function (config) {
             var map = L.map(element, config);
-
+            
             map.attributionControl.setPrefix(false);
             map.zoomControl.setPosition('bottomleft');
             map.setMaxBounds([[-90,-360],[90,360]]);
@@ -58,9 +60,38 @@ function Maps(ConfigEndpoint, L, _, CONST) {
 
             // Add a layer control
             // L.control.layers(getBaseLayersForControl(), {}).addTo(map);
+            
+            //  easyButton is the leaflet plugin to reset the map-view
+            // var resetButton  = L.easyButton( '<span id="reset"> <i class = "fa fa-home"> </i> </span>', function(){
+            //     var defaultview = defaultValues();
+            //     map.setView([defaultview.lat, defaultview.lon], defaultview.zoom);
+            // });
+            var resetButton  = L.easyButton({
+                id: "reset-button",
+                position: "topright",
+                type: 'replace',  
+                leafletClasses: true,
+                states:[{                 
+                    // specify different icons and responses for your button
+                    stateName: 'reset-button',
+                    onClick: function(){
+                        var defaultview = defaultValues();
+                        map.setView([defaultview.lat, defaultview.lon], defaultview.zoom);
+                    },
+                    title: 'Reset to default view',
+                    icon: '<i class = "fa fa-home"></i>',
+                  }]
+            });
+
+            resetButton.addTo(map);
 
             return map;
         });
+    }
+
+    function setDefaultView() {
+        var defaultview = defaultConfig();
+        map.setview(defaultview.center, defaultview.zoom);
     }
 
     function getLeafletConfig() {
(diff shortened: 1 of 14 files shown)


rowasc:
Wow this looks fantastic, @rohit645 . Thank you so much! 
I'm assigning @Angamanga for a detailed review and she'll get back to you ASAP.

app/common/services/maps.js
```diff
@@ -58,11 +59,34 @@ function Maps(ConfigEndpoint, L, _, CONST) {
...
+                id: 'reset-button',
+                position: 'bottomleft',
+                type: 'replace',
+                leafletClasses: true,
+                states:[{
+                    // specify different icons and responses for your button
+                    stateName: 'reset-button',
+                    onClick: function() {
+                        var defaultview = defaultValues();
+                        map.setView([defaultview.lat, defaultview.lon], defaultview.zoom);
```
Angamanga: Send config to the defaultValues-function

app/common/services/maps.js
```diff
@@ -140,4 +163,14 @@ function Maps(ConfigEndpoint, L, _, CONST) {
             popupAnchor: [0, 0 - size[1]]
         });
     }
+
+    // default view is centered on nairobi
+    function defaultValues(config) {
```
Angamanga: Here, if you use config.center and config.zoom, you will get the default-values for the deployment, this defaults to Nairobi if the user has not set any own values but sometimes they have (this is set in Settings->General).

app/common/services/maps.js
```diff
@@ -58,11 +59,34 @@ function Maps(ConfigEndpoint, L, _, CONST) {
...
+                }]
+            });
+
+            resetButton.addTo(map);
 
             return map;
         });
     }
 
+    function setDefaultView() {
```
Angamanga: Is this function used? I cannot find it but maybe its a Leaflet-default-thing?

app/common/services/maps.js
```diff
@@ -112,9 +136,9 @@ function Maps(ConfigEndpoint, L, _, CONST) {
 
     function defaultConfig() {
         return {
-            scrollWheelZoom: false,
+            scrollWheelZoom: true,
             center: [-1.2833, 36.8167], // Default to centered on Nairobi
-            zoom: 8,
+            zoom: 3,
```
Angamanga: Nice catch! Zoom 8 is really too much :)

sass/overrides/_leaflet.scss
```diff
@@ -65,3 +65,10 @@ input.leaflet-control-layers-selector {
 .leaflet-map-pane svg.iconic {
     position: initial;
 }
+
+//  Override styles for reset button
+#reset-button {
+    color: black;
+    background-color: white;
+    padding: 8px 8px;
+}
```
Angamanga: Could you check the styles for the small maps in edit-mode and general-view? There is an odd shadow-outline outside of the icon: 
Screenshot 2020-02-24 at 10 46 01

app/common/services/maps.js
rohit645: Yes, you're right :). I should have seen that earlier. This piece of code is redundant.

> Is this function used? I cannot find it but maybe its a Leaflet-default-thing?

Yes, you are right :). This piece of code is redundant. I should have seen that earlier.

rohit645: > Is this function used? I cannot find it but maybe its a Leaflet-default-thing?

Yes, you are right. This piece of code is redundant. I should have seen that earlier :)

rohit645:
Thanks for the feedback @Angamanga . I have made the required changes. Kindly review and let me know if any further changes are required.

app/index.html
```diff
@@ -14,6 +14,7 @@
         <meta name="apple-itunes-app" content="app-id={{appStoreId}}, app-argument={{currentFullUrl}}>">
 
         <link rel="manifest" href="/manifest.json">
+        <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/4.7.0/css/font-awesome.min.css">
```
Angamanga: Remove this and use an icon from the pattern-library instead.

sass/overrides/_leaflet.scss
Angamanga: Great work getting rid of this :)

app/main/posts/views/post-view-map.html
```diff
@@ -3,7 +3,7 @@
     <div ui-view="mode-context" noanimation="true" filters="filters"></div>
     <main role="main">
 		<post-toolbar is-loading="isLoading" filters="filters" edit-mode="{ editing: false }"></post-toolbar>
-    	<div class="map-view">
+    	<div class="map-view" id="map-view">
 		    <div id="map-full-size" class="map full-size" ng-show="!noui"></div>
```
Angamanga: I don't think the id is needed here, we use the ids in the child-divs instead, one for using the normal view you see when navigating to the map-view and one for specific use cases when no ui around the map is needed.

rohit645:
Hey @Angamanga , I have made the suggested changes. Please review my PR.

Angamanga:
@Angamanga Note to self, merge after release

PR #1438, opened 2020-02-28
Fix- Can't type (URL) into video field, can only copy-paste.

noone0212:
- Enabled user typing into the video field.

Testing checklist:
- [ ]

Fixes ushahidi/platform#3613 .

Ping @ushahidi/platform


noone0212:
clabot:check

noone0212:
@Angamanga please review my pull request.

Angamanga:
@noone0212 Thank you for your contribution 🎉 I will review shortly!

Angamanga:
@noone0212 This is a super-tricky one. I would like to make another approach here than doing the debounce of 4 seconds. Even though it works, it may not be very user-friendly, maybe the user needs more time to write. Or maybe the user pastes a link in, then it gets a bit odd to wait 4 seconds.

I have another idea, if you take a look at this row: https://github.com/ushahidi/platform-client/blob/develop/app/main/posts/modify/post-video.directive.js#L87

Instead of notifying the user here, we could turn on and off an error-label in the html-template. 

But this also means that we need to add validation on save, to not save a faulty url. The post-validation after save happens in a service here: https://github.com/ushahidi/platform-client/blob/develop/app/main/posts/modify/post-edit.service.js#L24

Unfortunately, we have now entered the most gnarly parts of our codebase 😬. I am happy to jump on a short call and explain whats going and pair on the issue if you don't figure it out yourself! Just ping me on the community-chat or send me an email and I'll set something up!

PR #1448, opened 2020-03-10
Select all button in the filter

himil-vasava:

**Before Changes**
-Multiple select fields didn't have a select all button. 

Screenshot from 2020-03-10 11-55-45

**After Changes**
-A select all button is there using which all the fields can be selected at once.

Screenshot from 2020-03-10 12-02-14

Testing checklist:

- [X] I certify that I ran my checklist

Fixes ushahidi/platform#2376 .

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/views/filters/filter-form.html
+++ b/app/main/posts/views/filters/filter-form.html
@@ -1,7 +1,7 @@
 <fieldset overflow-toggle has-overflow="forms.length > 1" ng-model-options="{ updateOn: 'default' }">
     <label translate="app.surveys">Survey</label>
    
-    <div class="form-field checkbox">
+    <div class="form-field checkbox" ng-show="forms.length > 5">
         <label>
             <input  type="checkbox" name="selectedForms" ng-model="all"><em><span translate="category.select_all"></span></em>
         </label>
--- a/app/main/posts/views/filters/filter-form.html
+++ b/app/main/posts/views/filters/filter-form.html
@@ -3,12 +3,17 @@
 
     <div class="form-field checkbox" ng-repeat="(index, form) in forms">
         <label>
-            <input checklist-value="form.id" checklist-model="selectedForms" type="checkbox" name="selectedForms"> <bdi>{{ ::form.name }}</bdi>
+            <input checklist-value="form.id" checklist-model="selectedForms" type="checkbox" name="selectedForms" ng-checked="all"> <bdi>{{ ::form.name }}</bdi>
         </label>
     </div>
     <div class="form-field checkbox">
         <label>
-            <input checklist-value="'none'" checklist-model="selectedForms" type="checkbox" name="selectedForms" > <span translate="nav.unknown">Unknown</span>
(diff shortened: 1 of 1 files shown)


Erioldoesdesign:
From a design/UI pov I would suggest select all/unselect all be at the top of the list rather than the end. 

The list of surveys could potentially be 10's of items long. I wouldn't want a user, going through the list carefully selecting the ones they want to view and then accidentally click unselect at the end and lose their choices. It's better to 'make this mistake' at the beginning rather than the end of a list :)

Erioldoesdesign:
I would also suggest that there be a visual different in the select/unselect all checkbox to differentiate it from the regular selections. This could be achieved a few ways (or a combination of them):

- More padding around the unselect/select all checkbox
- Different colours, font-weight or pixel size of the checkbox
- Different coloured 'background' to this single cell within the list to show difference

himil-vasava:
@Erioldoesdesign  Is it okay now?

Screenshot from 2020-03-10 21-52-03

Erioldoesdesign:
Let's also wait for @Angamanga or @rowasc to come back on the other points :) 

I would say this is good for now but would suggest improvement tickets for the other UI improvements

himil-vasava:
@Angamanga I have made the changes. Are they okay now?

Screenshot from 2020-03-12 00-35-38

himil-vasava:
@Angamanga I tried to use ng-if directive. Although it worked correctly by displaying the select all button if surveys are >5, the select all button stopped working. Is it something due to ng-checked attribute?

Angamanga:
@himil-vasava Thank you for your patience, so sorry for the delay in response! Try ng-show, that should work :)

himil-vasava:
@Angamanga Thanks! It worked :)

rowasc:
found some issues, sent a fix for this in steve-buscemi 
``` 
#1448 caused a regression that went undetected unfortunately. 
- The filters start out with none of the surveys selected, and then when I select one, filters don't work at all (ie even the filters " count " is still zero, I have to check, then uncheck, to see filters changing... and even then it doesn't work as expected )
- Restore filters is broken :( 
``` 
Testing checklist:
- [x] Open the filters dropdown without having made any changes to filters
- [x] The survey filters should have a  "Select all" checkbox, and it should be checked
- [x] Uncheck the "Select all" checkbox for surveys. All surveys should be unchecked. 
     - [x] Apply the filters. Nothing should change (same post count should appear) since an empty array is the same as a full one here. Ideally, we should not allow an empty survey filter list in the UI either but this fix doesn't address that debt.
- [x] Check one survey filter (ie select your first survey), and apply it. 
- [x] You should see posts only from that survey
- [x] The checkbox for "Select all" should remain unchecked
- [x] Check the "Select all" checkbox. All survey filters should be checked. Apply them and see all posts.
- [x] Reset filters
- [x] The select all checkbox should be checked, and no filters should be applied


@Obadha2 can you verify in steve (I only tested locally) 

@Obadha2

Obadha2:
QA'd, passed. 👍

PR #1458, opened 2020-03-17
"Enabling User to type rather than only pasting."

noone0212:
- Enabled user to type rather than only-paste into the video field 

Testing checklist:
- [ ]

Fixes ushahidi/platform#3613 .

Ping @ushahidi/platform


noone0212:
clabot:check

PR #1475, opened 2020-03-26
3896 delete warning

Angamanga:
- Adds a new notification with an extra check where the user writes the name of the survey to confirm deletion.
- Adjusts the warning text to be more clear

Testing checklist:
- [ ] Go to settings->surveys
- [ ] Click on delete
- [ ] A notification-modal should appear:
Screenshot 2020-03-26 at 10 02 40
- Click on Cancel
 - [ ] The modal should close
- [ ] The survey should not be deleted
- Click on delete again
- Write a text in the input-box that is not the survey-name
- Click on delete
- [ ] An error-label should appear
- [ ] The survey should not be deleted 
- Write the correct survey-name in the box
- [ ] The modal should close
- [ ] The survey should be deleted

Fixes ushahidi/platform#3896 .

Ping @ushahidi/platform


app/common/notifications/notify.service.js
```diff
@@ -217,7 +218,52 @@ function Notify(_, $q, $rootScope, $translate, SliderService, ModalService, Demo
 
         return deferred.promise;
     }
+    function deleteWithInput (type, checkName) {
+        var scope = getScope();
+        var deferred = $q.defer();
+        scope.noequal = false;
+
+        const modalText = {
+            survey: {
```
Angamanga: Not sure of this solution, maybe we should send the strings as arguments. I just found it so confusing with all the arguments for notifications.

app/common/notifications/notify.service.js
```diff
@@ -217,7 +218,52 @@ function Notify(_, $q, $rootScope, $translate, SliderService, ModalService, Demo
...
+                modalBody: 'notify.form.delete_form_confirm_desc',
+                errorLabel: 'notify.form.delete_form_error',
+                button: 'notify.form.delete_form_button'
+            }
+        };
+
+        var texts = modalText[type];
 
+        $translate(texts.modalBody, {check_name: checkName}).then(show, show);
+
```
Angamanga: Need to translate the modalBody this way because of translate-values.

Erioldoesdesign:
Is this gated by user type? like only an admin can do it or users with permissions enabled can do it?

I was thinking a log of surveys deleted and/or auto-messaging to an admin/owner of a deployment when a survey is deleted would help with historical archiving for projects :) (different issue probably!)

rowasc:
Yea it's only people with the correct permissions who can delete a survey. 
I'd love to hear more about the email feature you are envisioning @Erioldoesdesign (in a separate ticket). 
@Angamanga one thing I just realized, maybe we would also want to encourage them to do an export and download it before? IDK, maybe it's too much? I'd do that in a separate ticket since I want to release this extra warnings pretty soon tho.

app/common/locales/en.json
```diff
@@ -319,7 +319,7 @@
         "delete" : {
             "desc" : "<strong>If you delete this survey</strong>, all of its posts will also be deleted. Proceed with caution.",
             "delete" : "Delete survey",
-            "delete_this" : "Delete this survey",
+            "delete_this" : "Delete this surve",
```
rowasc: why is the `y` deleted? typo?

app/common/locales/en.json
```diff
@@ -1523,8 +1523,10 @@
             "save_attribute_success" : "Field {{name}} added",
             "edit_form_success" : "Survey {{name}} updated",
             "edit_stage_success" : "Survey task {{name}} updated",
-            "delete_form_confirm" : "Are you sure you want to delete this survey?",
-            "delete_form_confirm_desc" : "This action cannot be undone. Deleting this survey will remove all of its data, including posts.",
+            "delete_form_confirm" : "Are you sure you want to delete this survey and all its posts?",
+            "delete_form_confirm_desc" : "This action cannot be undone. Deleting this survey will remove all of its data, including posts. Write the name of the survey: <br><br><strong>{{check_name}}</strong><br><br>in the box below to confirm deletion.",
+            "delete_form_error": "The text you entered does not match the survey-name, please try again",
+            "delete_form_button": "Delete this survey",
```
rowasc: `Delete this survey and all its data` maybe? or did we go against that in our chat? can't recall

app/common/notifications/notify.service.js
```diff
@@ -217,7 +218,52 @@ function Notify(_, $q, $rootScope, $translate, SliderService, ModalService, Demo
 
         return deferred.promise;
     }
+    function deleteWithInput (type, checkName) {
+        var scope = getScope();
+        var deferred = $q.defer();
+        scope.noequal = false;
```
rowasc: can we do isEqual instead? I find it more intuitive although maybe it's just me, open to discuss

app/common/notifications/notify.service.js
rowasc: you mean instead of sending survey, send each individual string? 
I tend to like grouping as objects so I think this appraoch is great actually.

app/common/notifications/notify.service.js
```diff
@@ -217,7 +218,52 @@ function Notify(_, $q, $rootScope, $translate, SliderService, ModalService, Demo
...
+        };
+
+        var texts = modalText[type];
 
+        $translate(texts.modalBody, {check_name: checkName}).then(show, show);
+
+        function show(body) {
+
+            scope.confirm = function (name) {
+                if (name === checkName) {
```
rowasc: may look a tiny bit cleaner if we do

scope.isEqual = name === checkName; 
and then 
if (!scope.isEqual) {do blah}

defaulting to false for isEqual as you do of course.

app/settings/surveys/surveys.controller.js
```diff
@@ -49,7 +49,7 @@ function (
     };
 
     $scope.deleteSurvey = function (survey) {
-        Notify.confirmDelete('notify.form.delete_form_confirm', 'notify.form.delete_form_confirm_desc').then(function () {
+        Notify.deleteWithInput('survey', survey.name).then(function () {
```
rowasc: I really like this being introduced as a general pattern, nice job

test/unit/common/notifications/notify.service.spec.js
```diff
@@ -203,6 +203,21 @@ describe('Notify', function () {
         });
     });
 
+    describe('deleteWithInput', function () {
```
rowasc: would you be able to add a small test that checks the isEqual and confirmation?

app/common/notifications/notify.service.js
rowasc: if we start doing typescript you can even state the object structure more explicitly later on (not in this PR! ha) and that helps validate

Erioldoesdesign:
>" export and download it before..." <<< this is a great option to offer! Some good recommendation help text like... 
'We recommend you download your survey structure and data before deleting it and keeping it in a secure location so if you need it in the future, perhaps for audit or data protection purposes, you'll be able to find it'

I'll raise a separate ticket with my 'Email the owner/admin when a survey is deleted' idea

app/common/notifications/notify.service.js
Angamanga: Totally! Thanks!

app/settings/surveys/surveys.controller.js
Angamanga: Yes, I think its easier to understand whats going on this way...

app/common/locales/en.json
Angamanga: typo...

app/common/locales/en.json
Angamanga: I don't think we talked about it. Do you think people understand that this means posts?

Angamanga:
@rowasc Pushed fixes after review. We might need to discuss the test, I don't find a way to check that function that is inside another function, may need to refactor, but then I step away from the pattern we already use for notifications (which might be ok I suppose).

PR #1473, opened 2020-03-24
E2E test infrastructure

individual-it:
This pull request makes the following changes:
- adding nightwatch as end-to-end testing framework
- testing fixes done in #1454 (that is why this PR is on top of #1454)

Testing checklist:
- [x] :robot: 

- [x] I certify that I ran my checklist

Fixes ushahidi/platform# .

Ping @ushahidi/platform

PR #1477, opened 2020-03-27
fix: csv column titles not shown in FF

individual-it:
- makes sure the dropdown list is shown correctly in all browsers

Testing checklist:
- [x] check the import in Chrome
- [x] check the import in Firefox

Fixes ushahidi/platform#3903 .

Ping @ushahidi/platform


individual-it:
@rowasc if you approve https://github.com/ushahidi/platform-client/pull/1473 I could make a couple of UI tests for it and make sure they also run in FF and not only in Chrome

PR #1478, opened 2020-03-30
Fix - Tell user video link is broken even though valid.

noone0212:
- This tells user if video url is playable or not even though it matches youtube/vimeo url.

Testing checklist:
- [ ]

Fixes ushahidi/platform#3612 .

Ping @ushahidi/platform


noone0212:
Presently, it works for vimeo urls successfully. I am not able to obtain API key or Oauth token for youtube urls. Please someone guide me to resolve this issue!

rowasc:
Thanks @noone0212 for your contribution!
@Angamanga @noone0212 
Let's rethink this, I think. I would prioritize first fixing the fact that if it is invalid it won't stop the user from submitting, and then we can think of the correct approach for verifying vimeo/youtube, which I believe should be done (if we do it) in a service, not the directive itself.

rowasc:
Ah, I see someone is already working on what I mentrion above for https://github.com/ushahidi/platform/issues/3904. Let's keep this on hold for now until that's resolved

PR #1479, opened 2020-03-31
Fix - Secondary toolbar and add-post-button look untidy while editing post.

noone0212:
- Hiding add-post-button while editing the post.

Testing checklist:
  - Login into the deployment.
  - Go to Data.
  - Select any post to edit.
  - Scroll down to end of page, previously add-post-button and secondary toolbar looked untidy.
  - Now, while editing post add-post-button is removed completely.
  - Saving or cancelling the action bring the button back to its original position.

Fixes ushahidi/platform#3905 .

Ping @ushahidi/platform


noone0212:
After Disabling:

Screenshot from 2020-03-31 13-38-06

Erioldoesdesign:
From a design pov I don't think this is the best visualisation of a 'disabled' add post button. It's more distracting and users might be 'Why is this transparent floating button here? is it an error?' is my machine broken? the better thing from my pov is to remove it/make fully transparent and have it come back once the click cancel or save. or hide it completely behind the modal. 

This semi-transparent low opacity is confusing I think!

noone0212:
Screenshot from 2020-03-31 18-05-33
Please review this.

Erioldoesdesign:
Would definitely want @Angamanga @rowasc and the wider team to be ok with this as I don't have authority to review and clear this kind of amend

Angamanga:
@noone0212 The best option would be to completely remove the button while editing, we are trying to remove disabled buttons as best as we can. You could try a ng-show instead 👍

noone0212:
I have removed it completely. Please see https://github.com/ushahidi/platform-client/pull/1479/commits/dcbe4516df0221b0d63e3f10c607899b14cd7519.

Angamanga:
@noone0212 Thank you! Could you also add a testing-checklist to this pr? The steps to do in order to check if this works? That helps our QA-process! 🎉

rowasc:
@Obadha2 was this tested? If not, can you test it in the morning? Thanks !

Obadha2:
@rowasc this was looked at [here](https://github.com/ushahidi/platform/issues/3905#event-3214465020)

PR #1491, opened 2020-04-10
delete survey if anything goes wrong with it's title/description fixes #3215

rohit645:
-

Testing checklist:
- [ ]

Fixes ushahidi/platform#3215 .

Ping @ushahidi/platform

### DIFF
--- a/app/settings/surveys/survey-editor.directive.js
+++ b/app/settings/surveys/survey-editor.directive.js
@@ -77,6 +77,7 @@ function SurveyEditorController(
 
     $scope.deleteAttribute = deleteAttribute;
     $scope.saving_survey = false;
+    $scope.createNew = false;
     $scope.saveSurvey = saveSurvey;
     $scope.cancel = cancel;
 
@@ -328,7 +329,14 @@ function SurveyEditorController(
 
     function handleResponseErrors(errorResponse) {
         $scope.saving_survey = false;
-        Notify.apiErrors(errorResponse);
+        $scope.createNew = true;
+
+        FormEndpoint.delete({
+            id: $scope.survey.id
+        }).$promise.then(function () {
+            Notify.apiErrors(errorResponse);
+        });
+
     }
 
     // START -- Reorder tasks
@@ -576,7 +584,7 @@ function SurveyEditorController(
         $scope.saving_survey = true;
         // Save the survey
         FormEndpoint
-        .saveCache($scope.survey)
+        .saveCache($scope.survey, $scope.createNew)
         .$promise
         .then(function (survey) {
             // If the survey is new, cache the new id
@@ -592,6 +600,7 @@ function SurveyEditorController(
         })
         .then(function () {
             // Display success message
+            $scope.createNew = false;
             SurveyNotify.success(
                 'notify.form.edit_form_success',
                 { name: $scope.survey.name },
--- a/app/settings/surveys/survey-editor.directive.js
+++ b/app/settings/surveys/survey-editor.directive.js
@@ -329,7 +329,6 @@ function SurveyEditorController(
 
     function handleResponseErrors(errorResponse) {
         $scope.saving_survey = false;
-        $scope.createNew = true;
 
         FormEndpoint.delete({
             id: $scope.survey.id
@@ -337,6 +336,7 @@ function SurveyEditorController(
             Notify.apiErrors(errorResponse);
         });
 
+        delete $scope.survey.id;
     }
 
     // START -- Reorder tasks
@@ -584,7 +584,7 @@ function SurveyEditorController(
         $scope.saving_survey = true;
         // Save the survey
         FormEndpoint
-        .saveCache($scope.survey, $scope.createNew)
+        .saveCache($scope.survey)
         .$promise
         .then(function (survey) {
             // If the survey is new, cache the new id
@@ -600,7 +600,6 @@ function SurveyEditorController(
         })
         .then(function () {
             // Display success message
-            $scope.createNew = false;
             SurveyNotify.success(
                 'notify.form.edit_form_success',
                 { name: $scope.survey.name },
--- a/app/common/services/endpoints/form.js
+++ b/app/common/services/endpoints/form.js
@@ -51,8 +51,8 @@ function (
         return FormEndpoint.query();
     };
 
-    FormEndpoint.saveCache = function (item) {
-        var persist = item.id ? FormEndpoint.update : FormEndpoint.save;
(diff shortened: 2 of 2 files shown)


rohit645:
@Angamanga, This PR makes changes as you suggested in the ushahidi/platform#3215 (i.e.deleting faulty survey in an error handler and making POST request second time instead of PUT request). But the problem is second POST request is made with the same id "x" as the first post request (and it says **method not allowed**). How do I proceed further with this issue? Any suggestions are appreciated.

app/settings/surveys/survey-editor.directive.js
```diff
@@ -576,7 +584,7 @@ function SurveyEditorController(
         $scope.saving_survey = true;
         // Save the survey
         FormEndpoint
-        .saveCache($scope.survey)
+        .saveCache($scope.survey, $scope.createNew)
```
Angamanga, line 587: You don't need the createNew attribute, simply delete the id from $scope.survey after deleting the form instead.

app/settings/surveys/survey-editor.directive.js
```diff
@@ -328,7 +329,14 @@ function SurveyEditorController(
...
     function handleResponseErrors(errorResponse) {
         $scope.saving_survey = false;
-        Notify.apiErrors(errorResponse);
+        $scope.createNew = true;
+
+        FormEndpoint.delete({
+            id: $scope.survey.id
+        }).$promise.then(function () {
+            Notify.apiErrors(errorResponse);
+        });
```
Angamanga, line 338: after successful deletion, you can delete the id from  $scope.survey.

Angamanga:
@rohit645 I added two comments in the code. The reason why you get the error the first time is because the data you send is wrong (by purpose since we wanted to trigger the error). 

The second time is a tricky one, the reason why it happens is that after saving the survey, and before deleting the survey, the survey-id it is saved in the tasks and attribute-objects too: 
https://github.com/ushahidi/platform-client/blob/develop/app/settings/surveys/survey-editor.directive.js#L614
and 
https://github.com/ushahidi/platform-client/blob/develop/app/settings/surveys/survey-editor.directive.js#L643

It complains because it tries to find a survey.id thats not in the database. So we unfortunately need to check and delete those ids as well if something goes wrong. I did not realise this from the beginning. I found a better trick to trigger the error that might make things easier to test. I added an extra attribute to the "default" attributes that are created in the beginning (https://github.com/ushahidi/platform-client/blob/develop/app/settings/surveys/survey-editor.directive.js#L141):
```
{
           cardinality: 0,
           input: 'text',
           label: 'Title',
           priority: 1,
           required: true,
           type: 'title',
           options: [],
           config: {},
           form_stage_id: getInterimId()
},
{
           cardinality: 0,
           input: 'text',
           label: 'Description',
           priority: 2,
           required: true,
           type: 'description',
           options: [],
           config: {},
           form_stage_id: getInterimId()
 },
{
             cardinality: 0,
             type: 'number',
             input: 'text',
             label: 'Field with error',
             priority: 2,
             options: [],
             config: {},
             form_stage_id: getInterimId()
  } 
```

This should make it easier to test since that field will get added to the ui when creating a new survey,  but can be removed, which title and description cannot.

Let me know if things are unclear and you want more guidance. This part of the code is super-confusing and complex so its not a very straight forward issue. Well done getting this far :)

rohit645:
> @rohit645 I added two comments in the code. The reason why you get the error the first time is because the data you send is wrong (by purpose since we wanted to trigger the error).
> 
> The second time is a tricky one, the reason why it happens is that after saving the survey, and before deleting the survey, the survey-id it is saved in the tasks and attribute-objects too:
> https://github.com/ushahidi/platform-client/blob/develop/app/settings/surveys/survey-editor.directive.js#L614
> and
> https://github.com/ushahidi/platform-client/blob/develop/app/settings/surveys/survey-editor.directive.js#L643
> 
> It complains because it tries to find a survey.id thats not in the database. So we unfortunately need to check and delete those ids as well if something goes wrong. I did not realise this from the beginning. I found a better trick to trigger the error that might make things easier to test. I added an extra attribute to the "default" attributes that are created in the beginning (https://github.com/ushahidi/platform-client/blob/develop/app/settings/surveys/survey-editor.directive.js#L141):
> 
> ```
> {
>            cardinality: 0,
>            input: 'text',
>            label: 'Title',
>            priority: 1,
>            required: true,
>            type: 'title',
>            options: [],
>            config: {},
>            form_stage_id: getInterimId()
> },
> {
>            cardinality: 0,
>            input: 'text',
>            label: 'Description',
>            priority: 2,
>            required: true,
>            type: 'description',
>            options: [],
>            config: {},
>            form_stage_id: getInterimId()
>  },
> {
>              cardinality: 0,
>              type: 'number',
>              input: 'text',
>              label: 'Field with error',
>              priority: 2,
>              options: [],
>              config: {},
>              form_stage_id: getInterimId()
>   } 
> ```
> 
> This should make it easier to test since that field will get added to the ui when creating a new survey, but can be removed, which title and description cannot.
> 
> Also, is there any reason you removed the attribute-editor? I don't get the branch to work without them?
> 
> Let me know if things are unclear and you want more guidance. This part of the code is super-confusing and complex so its not a very straight forward issue. Well done getting this far :)

Deleting form in API and from $scope.survey deletes the ID. But problem occurs when that ID is used to access attributes, tasks as you highlighted above. And this gets even more complex if we try to delete those ID's also, makes the code even more complex.

Rather I was looking for some easy and convenient fix for this. What about creating a new attribute associated with form (say for example 'valid'). VALID will be set to **true** if the survey is saved successfully. And **false** if some error occurs :) in error handler may-be, (this way our survey will be created and stored in Database and it's valid attribute will still be set to some boolean value). Finally, this valid attribute can be used to decide whether it should be displayed or not (by checking it at https://github.com/ushahidi/platform-client/blob/3a9d668c5a0434ad8f7672ebfbb4cfd1701bccd8/app/settings/surveys/surveys.controller.js#L47 )? 
But this will require changes in API also (to make API return **valid** along with other attributes like name, desc,..).
Suggestions?

Angamanga:
Hi @rohit645! We have started working on a new feature for surveys. I just realised there will be a lot of changes to this directive, and we will be able to solve this in the backend instead. I am SO sorry telling you this now, after you have done so much work on this pr!! We are super thankful for your hard work and value what you do! If you want help finding another issue to work on, please let me know!

Angamanga:
Closing this since this bug will be fixed with the introduction of v5

PR #1497, opened 2020-04-16
Production aws

Angamanga:
- merging back hotfix for build-dependencies

Testing checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


rowasc:
@Angamanga do we still want to do this now? Since the hotfix didn't fix prod maybe we should revert it? not entirely sure TBH

Angamanga:
@rowasc It fixed the issue with svgs in the notice-slider though, but we can do that the correct way from develop instead

PR #1512, opened 2020-04-28
Sort users

himil-vasava:
**Before**
Screenshot from 2020-04-28 20-27-41

**After**
Screenshot from 2020-04-28 20-25-00

Testing checklist:

Fixes ushahidi/platform#3265 .

Ping @ushahidi/platform


Angamanga:
@himil-vasava 
Try to do this:
1. We need to add order to the filters, that can be done here: https://github.com/ushahidi/platform-client/blob/develop/app/settings/users/users.controller.js#L43. We have either "asc" or "desc" to choose between so lets start with ascending (from a-z).
2. Now we need to connect this with the filter-users directive. Since filters is sent to that directive, we can use this. If you check the value on filters, its the same as in the users.controller to start with. Now connect the options to the filter-model, something like this (I changed to radio-buttons since we can only choose one of the options at the time: 
```
 <div class="form-field checkbox">
      <label>
          <input type="radio" name="order" ng-model="filters.order" value="asc">
           Ascending
       </label>
   </div>
   <div class="form-field checkbox">
       <label>
           <input type="radio" name="order" ng-model="filters.order" value="desc">
           Descending
       </label>
   </div>
```
I don't think you need to change anything in filter-users.directive.js at all :)

Let me know if this work 🎉

app/settings/users/filter-users.directive.js
```diff
@@ -24,6 +24,7 @@ function (
             _
         ) {
             $scope.filtersMenuOpen = false;
+            $scope.sortalph="realname";
```
Angamanga, line 27: You can remove this now since we don't use it anymore.

himil-vasava: @Angamanga I have removed the unnecessary line of code

tuxpiper:
@Angamanga since the change request was addressed, looks like this could be merged?

Angamanga (approved):
This works fine now! Thank you for your contribution 🎉

Angamanga:
Note: Wait with merge until the data-filters-fix is rolled-back or fixed to keep things sane.

PR #1514, opened 2020-04-30
Fixing tooltip bug.

noone0212:
- Fixing Poorly visible font/background on hint field for toggling featured option on Collections.

Testing checklist:
- [ ]

Fixes ushahidi/platform#3956 .

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/collections/editor.html
+++ b/app/main/posts/collections/editor.html
@@ -20,7 +20,7 @@
                 <svg class="iconic">
                     <use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="/img/iconic-sprite.svg#question-mark"></use>
                 </svg>
-                <span class="bug" translate>set.featured_collection_tooltip</span>
+                <span class="bug" style="color: black" translate>set.featured_collection_tooltip</span>
             </span>
         </label>
         <div class="toggle-switch">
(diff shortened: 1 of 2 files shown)


noone0212:
I also found same issue in saved search editor and fixed it here itself.
Screenshot from 2020-05-01 05-05-05

Angamanga:
@noone0212 Lets try to fix this in the pattern-library instead to avoid using inline styles. Whats happening here is that there is two different classes that clashes. We need to increase the specificity for the color for the bug class inside form-fields and tooltips. 

The pattern-library lives in another repository:

What you need to do is adding  this:
```
.bug {
            color: $color-dark-alpha;
        }
```

to the line here: https://github.com/ushahidi/platform-pattern-library/blob/master/assets/sass/2_fragments/_forms.scss#L48

Once that is done, I will create a new release for the pattern-library and add it to the platform-client! You don't _need_ to set the pattern-library up and running, but if you are curious and want to do that, you can read the instructions here: https://docs.ushahidi.com/platform-developer-documentation/development-and-code/setup_alternatives/setting-up-the-pattern-library-for-development

Let me know if you need any guidance! :tada:

tuxpiper:
Closing in favor of https://github.com/ushahidi/platform-pattern-library/pull/239

PR #1535, opened 2020-08-20
Revert "Filters mode context"

Angamanga:
- Removing attempt to disconnect the filter-dropdown and the mode-context-bar
- Reverts ushahidi/platform-client#1533


Angamanga:
@tuxpiper First rollback-pr. It looks like we need to do it in steps.

tuxpiper (approved):
Looks good to me :+1:

PR #1546, opened 2020-09-01
refactoring loading of posts to avoid duplicate requests

Angamanga:
- Refactors the way we load posts in the map to make sure we get the offset right and don't make duplicate calls.

Testing checklist:
This bug happened (sometimes) in deployments with > 200 posts with location
To test the fix:
Go to mapview in a deployment with clustering of markers enabled
- See posts loading
- Zoom out to full extent (only one marker with all posts)
- [ ] The count of posts should match the number of posts on the map (the same nb as is in the filter-dropdown and in mode-context bar).
- Open the filter-dropdown
- Add a filter
- Zoom out again if needed
- [ ] The count of posts should match the number of posts on the map (the same nb as is in the filter-dropdown and in mode-context bar).
 

Fixes ushahidi/platform#4064 .

Ping @ushahidi/platform


app/main/posts/views/post-view-map.directive.js
```diff
@@ -147,54 +145,47 @@ function PostViewMap(PostEndpoint, Maps, _, PostFilters, L, $q, $rootScope, $com
...
+            let getFirstPostChunk = PostEndpoint.geojson(conditions);
+            currentGeoJsonRequests.push(getFirstPostChunk);
+            getFirstPostChunk.$promise.then(function (posts) {
+                // Adding the first 200 posts to map here and getting the totals
                 $scope.stats.filteredPosts = posts.total;
+                addPostsToMap(posts)
+
+                // Moving on to request rest of the posts
+                if (posts.total > limit) {
+                    for (let i = 200; i < posts.total; i = i + limit) {
```
bb, line 179: Instead of the literal `200`, the limit variable could be reused.

rowasc: agree. Thanks. 
@Angamanga pinging for you to follow up on this small change and then we could merge, I think?

Angamanga:
@rowasc Fixed the change, ready for review 👍

rowasc (approved):
@Obadha2 this can be tested. 
@Angamanga thanks !

PR #1549, opened 2020-10-06
Adds check for user's permission before requesting endpoint

renujain31:
- Adds a function in [post-metadata.service.js](https://github.com/renujain31/platform-client/blob/check_permission/app/main/posts/common/post-metadata.service.js#L50) that validates the user before making request to user endpoint.
- Calls that function before making the request in [filter-transformers.service.js](https://github.com/renujain31/platform-client/blob/check_permission/app/main/posts/views/filters/filter-transformers.service.js#L12)

- Go to map or data-view without logging in
- Check the console
- [ ] No error-messages about missing auth-token should be visible in the console
- Add a number of filters
- [ ] The active-filter icons should appear in the filter-dropdown

Repeat above as logged in user

Fixes ushahidi/platform#4075 .
Also fixes ushahidi/platform#4075.
Ping @ushahidi/platform


renujain31:
@Angamanga Can you please review it? Thanks!

app/main/posts/views/filters/filter-transformers.service.js
```diff
@@ -9,21 +9,23 @@ function FilterTransformersService(_, FormEndpoint, TagEndpoint, RoleEndpoint,
...
+            return $q.all([
+                    RoleEndpoint.query().$promise,
+                    UserEndpoint.query().$promise,
+                    TagEndpoint.query().$promise,
+                    FormEndpoint.query().$promise,
+                    SavedSearchEndpoint.query({}).$promise,
+                    CollectionEndpoint.query({}).$promise
+                ]).then(function (results) {
+                roles = _.indexBy(results[0], 'name');
+                users = _.indexBy(results[1], 'id');
```
Angamanga, line 22: So if you do the change with just making the role and user-requests if the user is validated, you will need to do the same check here and only assign roles and users if those requests were made 🎉

Angamanga (changes requested):
Nice work! I added some comments for you, some of the requests still need to be made, even if the user is not validated. 🌻

renujain31:
@Angamanga I have made the changes. Can you review it! Thanks!

app/main/posts/common/post-metadata.service.js
```diff
@@ -1,3 +1,5 @@
+const { findLastIndex } = require('underscore');
```
Angamanga, line 1: It doesn't look like this is used? If not you can remove this line.

Angamanga (changes requested):
Well done! Now its working fine! Just one more tiny thing I missed last round, you can remove the "findLastIndex" since it is not used :)

renujain31:
I have made the changes @Angamanga

app/main/posts/views/filters/filter-transformers.service.js
```diff
@@ -9,21 +9,23 @@ function FilterTransformersService(_, FormEndpoint, TagEndpoint, RoleEndpoint,
...
-                CollectionEndpoint.query({}).$promise
-            ]).then(function (results) {
-            roles = _.indexBy(results[0], 'name');
-            users = _.indexBy(results[1], 'id');
-            tags = _.indexBy(results[2], 'id');
-            forms = _.indexBy(results[3], 'id');
-            savedSearches = _.indexBy(results[4], 'id');
-            collections = _.indexBy(results[5], 'id');
-        });
+        if (PostMetadataService.validateUser) {
```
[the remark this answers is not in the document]
Angamanga: Ok absolutely last thing :D You need to call the validateUser function by adding () after it like this: `PostMetadataService.validateUser()`, otherwise it just checks if the function is there or not...

renujain31:
Done. @Angamanga 

Sorry for so many iterations.😅

Angamanga (approved):
Works fine now! Thank you for your contribution!

Angamanga:
@Obadha2 Ready for QA! This removes the annoying error-messages in the console 🎉 🎉 🎉

Obadha2:
QA'd, passes. 👍

PR #1556, opened 2020-10-09
 #2384 Use native names in Select language dropdwn

siggrid:
- adds a new field `nativeName` with corresponding value for each language in app/common/locales/languages.json
- uses `nativeName` field as a display value for select option in app/common/directives/language-switch.html template (in addition to `name` field, in case native name is not provided) 

Testing checklist:
- [ ] go to Map page
- [ ] open Select language dropdown
- [ ] languages are displayed in their native names (for example, Chinese displayed as 中文)
- [ ] select any other language
- [ ] open Select language dropdown again
- [ ] languages are displayed in their native names

Fixes ushahidi/platform#2384 .

Ping @ushahidi/platform


siggrid:
[clabot:check]

siggrid:
[clabot:check]

rowasc (approved):
I love this change, thank you ! :clap: @Obadha2 could you test when you have time next week?

rowasc (changes requested):
oh,  sorry @siggrid I just realized there may be concerns on the transifex side that we need to discuss because the languages.json file is automatically generated.  I'm setting this as "rejected" but we will come back to it and we'll figure something out. I think we either need to auto-add the nativeLanguage through some scripting or maybe find a way to do it through transifex settings :thinking: @Angamanga did you have specific ideas on this one?

siggrid:
> oh, sorry @siggrid I just realized there may be concerns on the transifex side that we need to discuss because the languages.json file is automatically generated. I'm setting this as "rejected" but we will come back to it and we'll figure something out. I think we either need to auto-add the nativeLanguage through some scripting or maybe find a way to do it through transifex settings 🤔 @Angamanga did you have specific ideas on this one?

@rowasc Thanks for your feedback.
Want to suggest a possible solution: we are able to add nativeName when generating languages.json file, when running downloadLanguages in gulp. nativeName value comes from json file of each downloaded language.
As I don’t have access to transifex, it is not possible to test if this suggestion works properly.

rowasc:
@siggrid you are awesome :D Yeah this is definitely in the ballpark of what I imagined, I just didn't want to throw you into a potential rabbit hole without more guidance :) 
@Angamanga assigning to you for review , looks close!

Angamanga (approved):
Very nice solution! Approved! Thank you for your contribution 🎉

Angamanga:
@Obadha2 This is ready for QA!

Obadha2:
Hi @siggrid while testing and verifying this, I see languages still in their English names and not their native names.

siggrid:
> Hi @siggrid while testing and verifying this, I see languages still in their English names and not their native names.

Hi @Obadha2 I updated an initial message with additional steps. Please, let me know if it works now. Thanks.

siggrid:
@Obadha2 
A new list of languages after running `gulp languages` and then `gulp`

lang-list

@Angamanga  @rowasc 
I found a problem, when generated new languages.json
The newly downloaded list of languages is shorter than [previous one](https://github.com/ushahidi/platform/issues/2384#issuecomment-706120828). There are languages that weren’t injected into the list, because they are not meeting the [completion_threshold of 60% translated](https://github.com/ushahidi/platform-client/blob/64ea51e1f6804cdd5d9f5c2b1ea6347efdb00ddd/gulp/transifex-download.js#L11). Some locales were updated with new phrases and some were not. For example, Bulgarian locale [has new phrase in line 88](https://github.com/ushahidi/platform-client/blob/64ea51e1f6804cdd5d9f5c2b1ea6347efdb00ddd/app/common/locales/bg-BG.json#L88) and Italian [has not](https://github.com/ushahidi/platform-client/blob/develop/app/common/locales/it.json#L88)
Bulgarian has **77%** and is presented in the list of languages, whereas Italian has **59%** and is not presented in the list of languages.

Angamanga:
@tuxpiper Do we need to update the github actions in order to integrate this change? It adds a new gulp-command (`gulp languages`) to add the native language in languages.json.

tuxpiper:
> @tuxpiper Do we need to update the github actions in order to integrate this change? It adds a new gulp-command (`gulp languages`) to add the native language in languages.json.

@Angamanga  no, not necessarily. We will just need to run gulp languages and commit the differences before the next release. We should do that every time, but I made an oversight with the last release 🙄 

Running that command in the actions wouldn't have 100% the desired effect, as it still wouldn't make the languages available to developers cloning the repo.

Angamanga:
@tuxpiper Ah, right, since the languages are already there right now. Do you have any idea why the new phrase was added to Bulgarian but not Italian (see @siggrid comment above)? Is it a mistake from the translator (the line is in English)? I am not sure how the translation-workflow goes?

tuxpiper:
> @tuxpiper Ah, right, since the languages are already there right now. Do you have any idea why the new phrase was added to Bulgarian but not Italian (see @siggrid comment above)? Is it a mistake from the translator (the line is in English)? I am not sure how the translation-workflow goes?

mm not sure @Angamanga , I think the languages file is completely re-generated each time `gulp languages` is called . I would go look at the payloads from transifex in order to see exactly what happened.

At any rate, I see now that we need a new PR with these changes, since this one was reverted after its merge.

PR #1565, opened 2020-10-12
Adds warnings in  case of identical options in checkbox

renujain31:
- Disabled the `Add & Close` if duplicate options are present in the survey. 
- While creating a post for the survey where duplicate options are present
         - If user is admin: warning is shown.
         - Else: only unique options are made visible.

Testing checklist:
- Create a new survey by settings->surveys->create new survey.
- Add a new field "Checkbox".
- [x] While adding two similar field options the add and close button should be disabled.

- Create a new post in a survey which has two similar options.
- [x] If you are logged in as admin you would see a warning.
- [x] If you aren't logged in as admin you would see checkboxes and duplicates won't be present
- [x] No error in console about ng-repeat

Some Screenshots:

While creating survey:
image

While creating post in a survey with two similar options:
image

After logging out:
image

Fixes ushahidi/platform#3994 .

Ping @ushahidi/platform


renujain31:
@Angamanga @rowasc Can you please review this. Thanks!

Angamanga:
@renujain31 Thank you! I will review shortly!

app/settings/surveys/attribute-editor.html
```diff
@@ -26,6 +26,9 @@
    <div ng-if="canDisplay()">
        <div class="form-field" ng-if="hasOptions(editAttribute)">
           <label translate>form.field_options</label>
+          <div class="alert error" role="alert" ng-show="!valuesPermissible()">
+            <p translate>survey.duplicate_option</p>
```
Angamanga, line 30: Nice work and extra 🎉 🎉  for adding role="alert" here.

app/common/locales/en.json
```diff
@@ -646,6 +647,7 @@
         "posts_total_map": "Displaying {{posts}} of {{total_nb}} posts with location information on the map.",
         "search_results": "Results:",
         "by" : "by",
+        "duplicate_option" : "There's a problem with the checkbox configuration in the survey because of duplicates, please fix it.",
```
Angamanga, line 650: I think we should add "please notify the admin of the deployment" here instead since the users who adds post rarely are the same as the ones managing the deployments.

app/main/posts/modify/post-value-edit.html
```diff
@@ -221,7 +221,10 @@
             </div>
         </div>
         <!-- Container for fieldset structured elements -->
-        <fieldset ng-if="isFieldSetStructure(attribute) && attribute.options.length > 0">
+        <div class="alert" role="alert" ng-show="duplicatePresent(attribute) && isAdmin()">
```
Angamanga, line 224: Good work thinking about the post as well, not only the survey-config!

Angamanga (changes requested):
Well done, nearly there! I left a few comments in the code.

app/common/locales/en.json
renujain31: This message would only be shown to admin. Does it. still makes sense to add that? I fear I can't totally understand what you meant by

> since the users who adds post rarely are the same as the ones managing the deployments.

app/settings/surveys/attribute-editor.html
```diff
@@ -106,8 +109,10 @@
         </div>
    </div>
    <div class="form-field">
```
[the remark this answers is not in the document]
renujain31: I already added an warning for duplicate messages which looks something like this:
image

The code for same lies at L30 of the same file.
Do you mean that I should add it below the form instead of adding it above them?

app/common/locales/en.json
Angamanga: @renujain31 Ah, right, yes, if it is only shown to the admin its ok!

renujain31:
Hey @Angamanga , I have added the changes.

So now while creating the review, it looks like this:
Screenshot 2020-10-20 at 12 12 44 PM


The Add & Close button isn't disabled but we couldn't save the survey and a warning stating"Duplicate option is present" appears on the form.

Angamanga (approved):
This works fine now! Thank you for your contribution! 🎉

Angamanga:
@Obadha2 Ready for QA 👍

Obadha2:
@Angamanga while the button is disabled when there's duplicate options, it looks active even though it does not work. Could we change this so that it actually looks disabled(greyed out) too? Or is this desired behavior and appearance for this button.

> Hey @Angamanga , I have added the changes.
> 
> So now while creating the review, it looks like this:
> Screenshot 2020-10-20 at 12 12 44 PM
> 
> The Add & Close button isn't disabled but we couldn't save the survey and a warning stating"Duplicate option is present" appears on the form.

Angamanga:
@Obadha2 This follows the pattern that was already present in the add-field modal. If a user skip adding a title for example, the Add & Close button is still enabled and there is an error-message under the title-input-field. I think this was a design-decision at some point, to remove the greyed out, disabled buttons. I agree it is confusing but I also think its another issue than was fixed here.

Obadha2:
Got it. Passing this as is. 👍

PR #1542, opened 2020-08-25
FIX: Removing use of current location in embeds #4043

Angamanga:
This pull request makes the following changes:
- Removing the "Use your current location" in the location-picker for posts when in embeds. The use of geolocation in embeds create the error: "Geolocation error: Geolocation has been disabled in this document by Feature Policy." and confuses both deployers and people adding the posts.

Testing checklist:
In "normal" mode (not embed)
- Create or edit a post for a survey with a location-field
- click on the text-search-bar in the location-picker
- [ ] The button "Use your current location" should be there

In "embed" mode
- Create or edit a post for a survey with a location-field
- click on the text-search-bar in the location-picker
- [ ] The button "Use your current location" should not be there

- [ ] I certify that I ran my checklist

Fixes ushahidi/platform# .

Ping @ushahidi/platform

PR #1572, opened 2020-10-14
Added alert role to alerts to improve accessibility

stacytonui:
-Added alert role to alerts to improve acccessibility

Testing checklist:

- [x] All the alerts have the alert role added to them and still function as alerts

Fixes ushahidi/platform#4109

Ping @ushahidi/platform


stacytonui:
[clabot:check]

rowasc (changes requested):
Hey @stacytonui thanks for your contribution!. Please add a testing checklist as directed by the PR template, and verify that it passes. 

Here's an example of a pull request with a checklist https://github.com/ushahidi/platform-client/pull/1542 
Thanks!

stacytonui:
@rowasc Thankyou. I've added the checklist.

rowasc (changes requested):
Hi @stacytonui , thanks for the change. 

- I hadn't noticed before, but I see role is in many cases being applied to field containers (ie divs for a field) withins alerts, not necessarily to the alert container itself. This happens in multiple files in this PR. It can quickly become really confusing if every element within a form is an alert. 

- related to the above: we should be using `alertdialog` role for interactive alerts (those which have forms) https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alertdialog_role . 

- related to the above: we should be using `alert` role for non-interactive alerts (those which are text-only) https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alert_role. 


- In terms of testing checklist, it's important to detail what was tested. ie: Registration modal opens , Login , etc
cc @renujain31 copying you since it's one of the issues you created and maybe you have thoughts on the above that can help. I suspect that ticket may be focused on text-only alerts but accidentally ended up applied to forms in modals too

stacytonui:
@rowasc oh right, apologies. I'll go through the linked articles and apply changes accordingly. Thanks for the correction

Angamanga:
@stacytonui Do you plan to fix the requested changes for this pr? Otherwise I'll merge it into its own branch so someone else can continue 👍

stacytonui:
Hey @Angamanga sorry for the late reply. Could you please assign this to someone else

PR #1577, opened 2020-10-15
Fix 4103: Add alt-text to avatar-images 

bimbolabuari:
# Description of Issue

Some images do not include alternative text as this helps in web accessibility, its primary purpose is for visually impaired users, This limits its accessibility, such as an assistive technology like screen reader reading out the alternative text when on the image, making it easier for assistive technology users to help have a feel/identity an image.

## Proposed solution

 For images without an alternative text, an alternative text "user avatar" is given. For example, as shown below:

```html
  <img alt="user avatar" src="//www.gravatar.com/avatar/{{ message.user.gravatar }}?d=retro&s=40" class="avatar">
```
## Testing

The testing was done manually by turning off the website image and also using a screen reader which in turn shows alternative text. The `gulp test` command was run to make sure no test is broken/affected by the changes.

### Testing checklist:

- [x] I certify that I ran my checklist

- [ ] Turn off website images on site settings
- [ ] Alternatively use a screen reader
- [ ] Click on "Data" in the left-hand mode-bar
- [ ] Click on a current post
- [ ] The alternative text `user avatar` will display 

## Screenshots
|  Current        |    Proposed changes    |
| --------- |:----------:|
| add-alt-text-pervious    |      add-alt-text-later

Fixes ushahidi/platform#4103

Ping @ushahidi/platform @rowasc @Angamanga

PR #1583, opened 2020-10-16
Add alt tags to gravatar images

joneskj55:
- Adds alt tags to img elements that are displaying avatars.

Testing checklist:
- [x]

Fixes ushahidi/platform#4103 .

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/common/post-metadata.html
+++ b/app/main/posts/common/post-metadata.html
@@ -17,7 +17,7 @@
       </li>
 
       <li>
-        <img src="//www.gravatar.com/avatar/{{ post.user.gravatar || '00000000000000000000000000000000' }}?d=retro&s=40" class="avatar">
+        <img src="//www.gravatar.com/avatar/{{ post.user.gravatar || '00000000000000000000000000000000' }}?d=retro&s=40" class="avatar" alt="user avatar">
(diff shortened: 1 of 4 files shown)


joneskj55:
[clabot:check]

rowasc (approved):
Thanks for your contribution. @joneskj55 . This looks good/follows the spec!


Regardless of that I'd like to ask  @Angamanga a question before we can merge.  I don't think I understand why we would want a screen reader to call out the "user avatar"  image (instead of skipping it) since it's a low signal text/adds noise, no?  but at the same time assume this came from our accessibility reviews and I'm just missing something? sorry I didn't catch this at the ticket-writing stage to ask, just want to make sure I understand it now. 
Walter: hold on merge please.

Angamanga:
@rowasc We could think of it as a decorative image and add an empty alt="" to skip it. I am thinking about users that do see things but not enough and use screen-readers. Wouldn't they get confused if they see that it is something there but not what it is? Its not like there is any other explanation anywhere about what it is.

tuxpiper:
Just adding a note that ushahidi/platform#4103 is also fixed by #1577 (released today)

Angamanga:
@tuxpiper Thank you for noticing, closing this since the repo looks deleted already.

PR #1589, opened 2020-10-20
fix undefined error

helen-ndip:
-

Testing checklist:
- [ x]  Log in and log out while in `Activity` view, refresh the page and the  undefined text that usually appears is fixed.

- [ x] I certify that I ran my checklist

Fixes ushahidi/platform#4107 .

Ping @ushahidi/platform


helen-ndip:
@rowasc  I succeeded to get rid of the other commits and made a new pull request. Thanks.

rowasc (changes requested):
Thanks @helen-ndip . Just FYI you don't need to create new pull requests when we send feedback, you can just update the one you have :) .. 
I'm afraid this may not work when a deployer has setup markdown in their site description. But maybe doing the ng-if + markdown-to-html rendered would work? To verify this works when using markdown, you use a text like 
```
# A site to test regression 1 
1. Hello
2. World
``` 
and verify that it renders correctly (it should not show the `#` characters and instead render a heading (ie h1) corresponding to the markdown `#`  style

helen-ndip:
> Thanks @helen-ndip . Just FYI you don't need to create new pull requests when we send feedback, you can just update the one you have :) ..
> I'm afraid this may not work when a deployer has setup markdown in their site description. But maybe doing the ng-if + markdown-to-html rendered would work? To verify this works when using markdown, you use a text like
> 
> ```
> # A site to test regression 1 
> 1. Hello
> 2. World
> ```
> 
> and verify that it renders correctly (it should not show the `#` characters and instead render a heading (ie h1) corresponding to the markdown `#` style

@rowasc  I did  ng-if + markdown-to-html to render the text like you said in the fix.

Angamanga (approved):
This works fine now! Thank you for your contribution!

Angamanga:
@Obadha2 Ready for QA!

Obadha2:
QA'd, passes. 👍

PR #1591, opened 2020-10-20
fix(posts-form): Fixes default date field

AngadSethi:
**This pull request makes the following changes:**
Fixes the workings of the date field while creating posts. Here's what it does now:

- If there is a default value, we let JavaScript parse it and create a Date object. If it's valid we have a date, otherwise it is empty.
- If there is no default value and the field is required, we enter today's date, otherwise we pass in a bogus string, which gives an empty date.

Testing checklist:
- [ ] A survey with an unrequired date field and no default value now has an empty date field.
- [ ] A survey with a default value , regardless of its requirement, has a date object parsed from default value.

Fixes ushahidi/platform#3961


rowasc (commented):
Thanks @AngadSethi ! 
:thinking: This Date('') will result in an Invalid Date, which makes sense but is also slightly annoying. Would it be possible to change it to null instead?

AngadSethi:
The `Date('')` is very annoying and my first thought was to pass a null value. However, passing a null value results in the datepicker field not being rendered at all, so I switched back to an empty string.

rowasc (approved):
@AngadSethi ah, that makes sense! Thanks for following up. @Obadha2 let's get this tested when you are back.

Obadha2:
QA'd, passes. 👍

tuxpiper:
Reverted by #1613

PR #1608, opened 2020-10-30
fix(posts-form): Fixes default number field

AngadSethi:
- If a default value is submitted in decimal field, it now shows the entire decimal, instead of just the integral part.
- If there is both an integer field and decimal field, `parseInt` is used for the integer field and `parseFloat` is used for the decimal field. (To sidestep ushahidi/platform#3847)
- Removes the trailing zeros from the record fetched from the database while displaying the post

Testing checklist:
- [ ] Add a 'decimal' input field to a survey.
- [ ] Specify a default decimal value like '5.64'
- [ ] Add a response. The field now shows '5.64' instead of just '5'
- [ ] Submit the response. The response now shows '5.64' instead of '5.6400'

Fixes ushahidi/platform#3954


Obadha2:
@Angamanga this is due for a review. Just pinging you so that it's not lost.

Angamanga (approved):

Obadha2:
QA'd, passes. 👍

Obadha2:
QA'd, passes. 👍

PR #1610, opened 2020-10-30
#4138 Improved header logical tab order

bukolabisuga:
-

Testing checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


bukolabisuga:
Hi @rowasc @Angamanga Kindly review this PR for #4138 https://github.com/ushahidi/platform/issues/4138

Angamanga (changes requested):
@bukolabisuga Even if we manage to tab to this, it still doesn't open once it gets focus. I think this is due to the dropdown-functionality we use, it might work if the whole <div> gets focus instead.

bukolabisuga:
Hi @Angamanga , thanks for the feedback!
You are right, you need to tab twice for the dropdown to show,
I can work on improving the experience such that the dropdown shows once you tab into the arrow (if you believe this is a better approach)

Angamanga:
@bukolabisuga Thank you! I think a solution where the dropdown opens once you press space would be the best solution, thats what we have for example in the language-dropdown and other similar places.

Angamanga:
@bukolabisuga Closing this, if you plan to work on it I am happy to reopen :)

PR #1612, opened 2020-11-03
Updating the filter searchbar

AngadSethi:
- Rectifies the filter search bar
- The user searching function now works consistently

Testing checklist:
- [ ] Sign in as an admin, go to 'Settings', then click on 'Users'
- [ ] Search for a user.
- [ ] Hit enter, or click on the button.
- [ ] The users list is shown correctly
- [ ] Only one request is sent to the backend per search or reset.

Fixes ushahidi/platform#3936

_Every time the enter key would be hit, the value of `model.q` would be reset to an empty string, which was changing the value of filters and thus calling the watch function._

### DIFF
--- a/app/common/directives/filter-system/filter-searchbar.html
+++ b/app/common/directives/filter-system/filter-searchbar.html
@@ -11,7 +11,7 @@
                 ng-model="model.q"
                 ng-focus="showSearchResults()"
                 ng-keyup="detectSubmit($event)"
-                ng-value="model.q=''"
+                ng-value="model.q"
             >
             <svg class="iconic" ng-show="!form.q.$viewValue.length">
               <use xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="/img/iconic-sprite.svg#magnifying-glass"></use>


Angamanga:
@AngadSethi This change does not seem to work, the first time the search is performed it works fine but if you try to search again, the search does not happen. Someone made an attempt to fix this last spring, and thats where they added "model.q=''" line. Previously it was model.q as you added back now so I suspect there is something more to this that we have not figured out yet :/. If you have time, any research into it is welcome 🙏

Angamanga (changes requested):
@AngadSethi This change does not seem to work, the first time the search is performed it works fine but if you try to search again, the search does not happen. Someone made an attempt to fix this last spring, and thats where they added "model.q=''" line. Previously it was model.q as you added back now so I suspect there is something more to this that we have not figured out yet :/. If you have time, any research into it is welcome 🙏

AngadSethi:
@Angamanga Thank you for the review!

> This change does not seem to work, the first time the search is performed it works fine but if you try to search again, the search does not happen.

You are absolutely right about this. I missed this when I was testing the fix. I'll explore this issue, and, hopefully, submit a fix soon

Angamanga:
@AngadSethi Thank you 🎉

PR #1614, opened 2020-11-04
Fix: Starting with null for datetime if no default is present

Angamanga:
- Starting with an empty date if no default-value is present when adding/editing a post

Testing checklist:
- [ ] Add a date-time field in a survey
- [ ] Add a post for that survey
- [ ] Leave the date-time field empty
- [ ] Save
- [ ] The post should save without a date-time value
- [ ] Add a new post and select a date
- [ ] The date-picker should work
- [ ] The time-picker should work
- [ ] The value should save

- Repeat above for editing a post

Fixes ushahidi/platform#4155 .

Ping @ushahidi/platform


Angamanga:
Added this change to v5-main, closing this pr!

PR #1617, opened 2020-11-10
fix(surveys): makes the first input focusable on every page reload

renujain31:
Makes the first input box in the following forms focusable after every page reload
- While creating/updating a new survey
- While adding/updating fields in a survey
- While adding tasks in the survey
- While updating user settings for hdx maintainer 

Testing checklist:
- [ ] While creating or updating new survey the survey title field gets focusable
- [ ] While adding or updating any new field the first input gets focusable
- [ ] While adding or updating any task  the first input gets focusable.
- [ ] While updating user settings for hdx maintainer the first input gets focusable.

Fixes ushahidi/platform#2073 .

Ping @ushahidi/platform

### DIFF
--- a/app/settings/surveys/attribute-editor.directive.js
+++ b/app/settings/surveys/attribute-editor.directive.js
@@ -48,7 +48,7 @@ function (
             };
 
             initiateEditor();
-
+            document.getElementById('attribute_name').focus();
             $scope.save = function (editAttribute, activeTask) {
                 editAttribute.instructions = $scope.editor.getMarkdown();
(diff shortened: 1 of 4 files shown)


Angamanga (commented):
@renujain31 I changed the base to v5-main. It would be neat to add the translation-editor to this as well so we can keep this pr open until you start coding again. I am also curious if there is another "AngularJs-way" to set focus. You could explore ng-focus for example. I am a bit unsure if the template always is available for the directive immediately, it might fail if the DOM is not loaded properly by the time it reaches that line of code?

renujain31:
Made a different pull request: #1657 because of a lot of conflicts.

PR #1620, opened 2020-11-16
fix(MapView): chaining since .all doesn't respect order and we have a…

rowasc:
… requirement for it here

-

Testing checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


tuxpiper (approved):
Yea I think this makes sense :+1:

PR #1621, opened 2020-11-16
Locale fix

rowasc:
- start by setting "en" as default locale then attempt to require a new one
- if the correct moment locale isn't available we default to the 2 letter code

Testing checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


tuxpiper (approved):
alrite 👍

PR #1622, opened 2020-11-16
fix(MapView stats): use the stats endpoint instead of counting the geojson entries since we can get the same info

rowasc:
fix(MapView stats): use the stats endpoint instead of counting the geojson entries since we can get the same info from the stats endpoint. 

- Changes a call to geojson for a call to stats which is less resource intensive.

Testing checklist:
- [ ] Load the map view with a deployment with many posts. 
- [ ] Verify that your filters dropdown displays the correct count for mapped vs unmapped posts (x of y. ) 
      - Note: the wording there is a bit odd because it implies its not showing the full dataset when in reality what is does is show the count of mapped vs totals, but that's a separate thing, didn't want to change it without more talking about it.

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/views/post-view-map.directive.js
+++ b/app/main/posts/views/post-view-map.directive.js
@@ -149,7 +149,7 @@ function PostViewMap(PostEndpoint, Maps, _, PostFilters, L, $q, $rootScope, $com
                     function (a,b) {
                         return a.total + b.total
                     }
-                );
+                ) - result.unmapped;
 
                 $scope.stats.unmapped = result.unmapped;
             });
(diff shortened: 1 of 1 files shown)


tuxpiper (approved):
Niice :+1: I favor this.

I was just wondering, may there be differences in count due to permissions (not all that familiar with the stats repository code in the API)

Angamanga:
@rowasc @tuxpiper In production right now, we display the result in the filter-dropdown as nb of posts in the search-result vs all posts with *location information*:


With this change, we would display the result as nb of posts vs ALL posts, unmapped and mapped:
Screenshot 2020-11-17 at 11 44 54

Is this a change in behaviour we want to do? For context, we discussed the behaviour here: https://ushahidi.slack.com/archives/C03UU507R/p1598257290004300

rowasc:
@Angamanga ooof sorry I totally missed that was going on . Fixing

Angamanga (approved):
Works fine 🎉

tuxpiper:
@Obadha2 did this pass QA?

PR #1631, opened 2020-12-02
#4150 - Fix link/button order on survey page

Ifycode:
**This pull request makes the following changes:**
- Fixes ushahidi/platform#4150

**Issues, observations and steps taken:**
- Tab order for links or buttons on the survey page goes backward (from right to left). Rearranged the html elements so that elements receive focus from left to right.
- New survey link/button receives focus twice as it's a button wrapped in an < a > tag and both receive focus. Removed the button since it leads to some other page.

**Associated platform pattern library PR:**
- [Platform pattern library PR](https://github.com/ushahidi/platform-pattern-library/pull/262)

**Testing checklist:**
- [x] Add assets folder from platform-pattern-library containing the new changes made to the platform-client which also contains some new changes and run the project.
- [x] Login, and use tab key to navigate to buttons/links on the survey page.
- [x] Check that links/buttons receive focus in the correct order.

Fixes ushahidi/platform# .

Ping @ushahidi/platform


Angamanga (changes requested):
@Ifycode This works fine but the button does not get the blue outline as the link when in focus. I see this is also happening in production so it is nothing new, but it would be great if you could figure out how to add that too 🙏

Ifycode:
Hi @Angamanga. Thank you for the review and happy new year to you. Sorry for my very late response. 

I will work on how to make the links get the blue outline.

Ifycode:
Hi @Angamanga, I checked out why the button doesn't get the blue outline even though it is a link. At first I thought it was some change made in the css, but soon found that the default focus ring or outline for focus-able elements has been updated to black (for bright backgrounds) - including links.

|Simple link I made with default settings | More info from [this source](https://blog.chromium.org/2020/03/updates-to-form-controls-and-focus.html) |
|------------ | ----------- |
| link-default | focusring |

This I guess they did for accessibility reasons. Would you want me to override the styles for links and change them to blue?

Angamanga:
@Ifycode Ah, that makes sense then. As long as the focus is visible in one way or the other, when I tested, the focus was not visible for that specific button I think. But now when I try to reproduce it, I do see focus so I am not sure if it was a temporary glitch in my browser :/.

Ifycode:
Oh ok @Angamanga No problems with this PR again I guess.

renujain31:
Hey @Angamanga I checked the PR and it is working fine in my local. Maybe we can merge it!

Angamanga (approved):
Works fine now! Approved 🎉

Angamanga:
@Obadha2 Ready for QA 🎉

PR #1651, opened 2021-02-04
fix(post-detail,categories): restores behavior of parent categories b…

tuxpiper:
…eing hidden in post detail view

The v5 API now returns parent categories , this has the side-effect in the client of showing those parent categories along their children. Because our display currently is just a comma separated list, this can be confusing.

It's also detrimental to some of our deployers, who had intentionally created specific survey fields for specific family groups of the defined categories.

Since there is little information to be gained from listing the parent categories along their children categories in a comma separated list (there's no display of their hierarchical relationship), we should restore the original behavior.

- This PR restores the original behavior of not listing parent categories

Testing checklist:
- [x] View a post with a category field , where some of the selected categories have parent categories
- [x] Ensure none of the categories displayed are parent categories

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/detail/post-category-value.directive.js
+++ b/app/main/posts/detail/post-category-value.directive.js
@@ -16,9 +16,16 @@ function PostCategoryValue() {
 PostCategoryValueController.$inject = ['$scope', '_', 'CategoriesSdk'];
 
 function PostCategoryValueController($scope, _, CategoriesSdk) {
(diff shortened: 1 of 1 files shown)


Angamanga (approved):
Works fine 🎉

Obadha2:
QA'd, passes. 👍

PR #1654, opened 2021-02-18
fix(post-editor): Copying post-object before preparing it for save)

Angamanga:
- uses a copy of the post-object to do pre-saving formatting
- this makes it possible to continue edit and save a post again after validation-errors from the backend

Testing checklist:
- Add a post to a survey with required fields
- Leave one of the required fields empty
- Save the post
- [ ] You should see an error-message
- Add a response in the required field
- Save again
- [ ] The post should save

- [ ] Repeat above for when editing a post

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/main/posts/modify/post-editor.directive.js
+++ b/app/main/posts/modify/post-editor.directive.js
@@ -157,11 +157,12 @@ function PostEditorController(
         // Create/update any associated media objects
         // Media creation must be completed before we can progress with saving
         resolveMedia().then(function () {
-            $scope.post.base_language = $scope.languages.active;
-            $scope.post.type = 'report';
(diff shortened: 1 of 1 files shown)


tuxpiper (commented):
@Angamanga  Makes sense to me :+1: Going to test it shortly. 

What was failing due to modifying base_language, type, form_id , etc .. before making the copy?

Angamanga:
@tuxpiper It was failing because we had removed the form-object from the post 😬

tuxpiper (approved):
Tested it, working well :+1:

Angamanga:
@Obadha2 Ready for QA!

PR #1659, opened 2021-03-28
Add keyboard accessibility to the 'who can see this' section in the edit collection modal

emmanuelhashy:
- Add keyboard accessibility to the 'who can see this' section in the edit collection modal

Testing checklist:
- [ ] Go to collections
- [ ]  Click the create new button
- [ ] Click tab on your keyboard until 'who can see this' section is reached
- [ ] Click the enter key to select an option

Fixes ushahidi/platform#4130.

Ping @ushahidi/platform


emmanuelhashy:
[clabot:check]

emmanuelhashy:
@Angamanga I have finished the task. Kindly review. Thanks!

app/common/directives/role-selector.directive.js
```diff
@@ -37,10 +37,12 @@ function RoleSelectorController($scope, RoleEndpoint, $translate, _) {
         if (_.indexOf($scope.model.role, admin.name) === -1) {
             $scope.model.role.push(admin.name);
         }
+        document.getElementById('add_roles').click();
     }
 
     // adding all available roles to model if user clicks 'Everyone'
     function setEveryone() {
         $scope.model.role = null;
+        document.getElementById('add_everyone').click();
```
Angamanga, line 46: Instead of doing the click here and on row 40, we need mimic the behaviour that is going on in the input-elements, otherwise this function will be run twice. What is happening in the input-elements is that the model used, "$scope.everyone", is set to either true or false. This makes the dropdown with the roles visible or hidden (see this line: https://github.com/ushahidi/platform-client/blob/develop/app/common/directives/role-selector.html#L49). So in the setEveryone-function, we need to set $scope.everyone = true and in the setAdmin-function, we need to set it to false.

package.json
```diff
@@ -131,7 +131,7 @@
     "socket.io-client": "2.0.3",
     "sortablejs": "1.10.0",
     "underscore": "^1.7.0",
-    "ushahidi-platform-pattern-library": "4.4.9",
+    "ushahidi-platform-pattern-library": "^4.4.9",
```
Angamanga, line 134: Remove the ^ here, we need to lock the pattern-library to this specific version, otherwise styling-changes that are about to be published gets published too early (for example, you may have seen the date-picker appearing in strange ways on some pages. Thats because changes to the date-picker is pending in a later minor, release of the pattern-library :) )

app/common/directives/role-selector.html
```diff
@@ -40,7 +40,7 @@
                     ng-model="everyone"
                     ng-click="setAdmin()"
                 >
-                <span translate="app.specific_roles">Specific roles...</span>
+                <span translate="app.specific_roles" tabindex="0" ng-keyup="$event.keyCode == 13 && setAdmin()">Specific roles...</span>
```
Angamanga, line 43: Use `===` here instead of `==`. It doesn't make much difference in this case but it is a good practice to use `===` since `==` may give unexpected results. If you want to read more about it check out these two: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness and https://dorey.github.io/JavaScript-Equality-Table/)

Angamanga (changes requested):
Thank you for your contribution! I have added some comments and requests for change in the code. 🎉

Ifycode:
@emmanuelhashy Can you see my comment? 

Sorry @Angamanga I'm trying to help Emmanuel see your review comments. Looks like he needs help with this from the discussion we had.

emmanuelhashy:
Thanks, @Ifycode. Yes, I can see.

Ifycode:
Alright. Best of luck @emmanuelhashy

app/common/directives/role-selector.html
emmanuelhashy: Alright! I will fix this.

package.json
emmanuelhashy: Alright! I will fix this

app/common/directives/role-selector.directive.js
emmanuelhashy: Alright, I will fix this.

emmanuelhashy:
@Angamanga I have made the changes as you recommended. Kindly review. Thanks

Angamanga (approved):
Approved! Thank you for your contribution 🎉

Angamanga:
@Obadha2 Ready for QA!

Obadha2:
QA'd, passes! :+1:

PR #1660, opened 2021-04-02
Add alert role where there's an alert css class in the template to improve accessibility.

emmanuelhashy:
- Enable screen readers accessibility on all the alert boxes across the platform-client's codebase.

Testing checklist:
- [ ]

Fixes ushahidi/platform#4109 .

Ping @ushahidi/platform

### DIFF
--- a/app/common/directives/category-selector.html
+++ b/app/common/directives/category-selector.html
@@ -45,7 +45,7 @@
         </label>
         </div>
     </div>
-    <div ng-if="categoryTranslationUnavailable.length > 0 && displayWarning" class="alert">
+    <div ng-if="categoryTranslationUnavailable.length > 0 && displayWarning" class="alert" role="alert">
         <span ng-if="categoryTranslationUnavailable.length === 1">Please note: The following category is not available in {{'languages.' + activeLanguage.language | translate}}. Its default-language will be displayed instead:</span>
         <span ng-if="categoryTranslationUnavailable.length > 1">Please note: The following categories are not available in {{'languages.' + activeLanguage.language | translate}}. Their default-languages will be displayed instead:</span>
         <ul>
--- a/app/main/posts/detail/post-detail-data.html
+++ b/app/main/posts/detail/post-detail-data.html
@@ -67,7 +67,7 @@
     </nav>
     <div class="listing-item tasks-tabs tabs-target active" ng-show="post.post_content.length > 1"
       >
-      <div class="alert"
+      <div class="alert" role="alert"
           ng-repeat="task in post.post_content | limitTo: 1 - post.post_content.length" 
           ng-if="!taskHasValues(task)"
           ng-show="task.id === visibleTask">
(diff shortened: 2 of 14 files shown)


emmanuelhashy:
Hi @Angamanga, I have finished the task. Kindly review. Thanks!

Angamanga (changes requested):
Thank you for working on this! While the changes are according to the instructions in the issue description, I want us to understand why we are adding the role="alert" to the elements. role="alert" is used when we want to alert a user who uses assistive technologies, like a screen-reader or braille, when browsing the web. When an element has role="alert", the reader interrupts what it reading and instead reads out the content in the alert element (More about it here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alert_role). I realize that in some places we use the class "alert" for styling, without it being an actual alert. Please test yourself with voice-over (I am not sure what os you are on, google "voice over" for your specific os) if you haven't already. Go through the changes you have made and see if it is an alert that needs interruption or simply extra information? I am happy to discuss specific places because I realize this could be confusing. Also, let me know if you need help to find the places the code adjusts in the UI 🎉

emmanuelhashy:
Hi @Angamanga Can you explain this "Please test yourself with voice-over" please!
I did some google search on "voice over" software and I got this meaning "Voice over software records your voice in high quality and adds it to your videos". Do you mean I should do some voice recording?

Angamanga:
@emmanuelhashy I was confusing, I meant a screen reader (https://en.wikipedia.org/wiki/Screen_reader). Voice-over is the name of the built-in screen reader for iOS 😃 .

emmanuelhashy:
@Angamanga While using the windows 10 screen reader "Narrator", I notice that in some pages the screen reader says "alert" without reading the content of the alert message. I think keeping the role="alert" attribute in such pages is redundant. What do you think?

emmanuelhashy:
Also, I need some clarity on this "Go through the changes you have made and see if it is an alert that needs interruption or simply extra information"

Angamanga:
@emmanuelhashy Could you give an example of a place where the screenreader only says "alert"? It may be redundant and it may be that the `role="alert"` attribute needs to be on another element. It depends on the message a person who doesn't use a screen reader sees :). With "Go through the changes you have made and see if it is an alert that needs interruption or simply extra information" I meant, is the alert an alert that needs attention now (=the screenreader stops what it is reading and reads out the info in the alert) or extra information that the screen reader can read when the user gets to that part of the UI?

emmanuelhashy:
@Angamanga I'd like to get a screenshot of the alert screens present on the website.

Angamanga:
@emmanuelhashy We use alerts for many different things, for example notifications when something goes wrong or is ready:
Screenshot 2021-04-26 at 10 21 40

or as error-messages in field-validation:

Screenshot 2021-04-26 at 10 22 14

PR #1655, opened 2021-02-25
Add flatpickr library for date picker

renujain31:
This pull request makes the following changes:
- Adds a new date picker library, [flatpickr](https://github.com/flatpickr/flatpickr), to make the datepicker accessible and removes the angular-datepicker library used earlier.

Testing checklist:
- [ ] Open the filters and navigate to the date field.
- [ ] The datepicker opens up and you can navigate through the dates using the arrows.
image

- [ ] Create a new survey that requires the datetime field.
- [ ] Add a post to that survey.
- [ ] The datepicker does have a time field too and you can set the date and time.
image

- [ ] Create a new survey that requires the date field.
- [ ] Add a post to that survey.
- [ ] The datepicker only have the date field and you can navigate through the dates using the arrows. 
image

- [ ] Go to the activity menu.
- [ ] Select Custom date range from the dropdown menu at top.
- [ ] The datepicker only have the date field and you can navigate through the dates using the arrows. 
image

- [ ] I certify that I ran my checklist

Fixes ushahidi/platform#4190 .

Ping @ushahidi/platform

PR #1661, opened 2021-04-04
4198 - Add focus-trap for sort & filter dropdown

Ifycode:
- Fixes ushahidi/platform#4198

#### Details:
* Adds focus trap for sort & filter drop-down watching the presence of the `.active` class on the container/form.
* trap is activated either when the `search` input receives focus or when the sort & filter button is clicked on using the `enter` button.
* Focus trap is deactivated once `esc` button is pressed.

#### Testing checklist:
- [x] Go to map page & use tab key to navigate the sort & filter drop-down area.
- [x] Check that focus is trapped when navigating inside the sort and filter drop-down.
- [x] Check that focus trap is deactivated when `esc` button is pressed.

#### Check for these 3 cases:
- [x] When `search` input receives focus and sort & filter menu drops down. 
- [x] When `search` input receives focus but no drop down menu (usually happens after `esc` has been used previously), and the `arrow-down` key is pressed to make sort & filter menu drop down again. 
- [x] When sort & filter `button` is clicked on using `enter` button.

Fixes ushahidi/platform# .

Ping @ushahidi/platform


Ifycode:
Hi @Angamanga I have finished making changes to sort & filter drop down. Help review. Thanks.

Angamanga (changes requested):
Very smart solution with the Mutationobserver 😍!! I like it and it works fine. However, I see some tests are failing, could you please check? Run `gulp test` in your terminal and you will see the test results. I think its that the FocusTrap library is not in the test app, you can see how I solved it here: https://github.com/ushahidi/platform-client/pull/1655/files#diff-8df717880024c569eb888c6a9bfa096bf25468f6a00a43a8f29eaddb65046d87R26

Ifycode:
Thank you @Angamanga 😊
Alright, I will run the test command and see what the problem is.

Ifycode:
Hi @Angamanga .

Ok, so the first thing I did was to run the `gulp test` command in the develop branch. It throws an in the first place before I created a branch from it and added  the new code. Here's the screenshot - ***develop***:

ushahidi-test-fail-dev

***Back to the my feature branch***:

Using `require('focus-trap');`, it gives the error `SourceMap position not found for trace: NotFoundError: NotFoundError: DOM Exception 8 (line 65295)`. Then about 6 tests fail from `filter-posts.directive.spec.js`.

```
.factory('FocusTrap', function () {
      return require('focus-trap');
})
```

I've also tried these below, no luck. The 6 tests also fail as the previous, then I get either `SourceMap position not found for trace: TypeError: undefined is not
a constructor (evaluating 'FocusTrap.createFocusTrap('#dropdown-window')') (line 65260)`
 or `SourceMap position not found for trace: Error: [$injector:undef] Provider 'FocusTrap' must return a value from $get factory method`.

```
.factory('FocusTrap', function () {
      return function () {};
       //return require('focus-trap').default; 
       //return require('focus-trap').createFocusTrap('#dropdown-window'); 
       //return module.exports = function () {};
})
```

I tried using ideas from the link you sent and from some other online resources. Haven't been able to get any that helps with this issue.

Angamanga:
@Ifycode Ahh, it's the MutationObserver! 
`.factory('FocusTrap', function () {
      return require('focus-trap');
})` is correct, but then it jumps on line 40 in the directive instead and that is the MutationObserver. I don't have time to look into it right now but if you google for "mock MutationObserver Jasmine" or similar I think you will figure it out! Let me know how it goes and I'll look into it later today as well 👍 .

Ifycode:
Oh... Interesting. I will look it up and let you know the outcome. Thank you.

Ifycode:
Hi @Angamanga. I searched like you asked. [This stack overflow question](https://stackoverflow.com/questions/50398703/jasmine-how-to-mock-mutationobserver) is the closest to the problem I think, but  could not get an idea of how it is solved since no answer was supplied.  [This on the other hand](https://stackoverflow.com/questions/51325374/failing-angular-directive-unit-test-on-mutationobserver-parameter-1-is-not-of-t) is not angularjs but I came across this and some others suggesting that setting a timeout or delay may help. If this is true, I'm not sure how to go about it. 

Although, I tried injecting it in the `beforeEach` function inside the `filter-posts.directive.spec.js` file like this:
````
beforeEach(angular.mock.inject(function (_$rootScope_, _$compile_, _PostFilters_, FocusTrap) {
````

Just so I can get to try out some of these info found, probably it will work. But I receive this error:
`SourceMap position not found for trace: An error was thrown in afterAll
ReferenceError: Can't find variable: FocusTrap in http://localhost:9876/base/test/unit/spec.bundle.js?585f83a09bb4f44fc3d69e898897617a1e315940 (line 240333)`

I thought I should still try out more things before coming here to type anything.

Angamanga:
@Ifycode For filter-posts.directive.spec.js, check how PostFilters is injected, you need to first define the FocusTrap variable in the beginning of the file (check line 5 for PostFilters) and then add it is an argument like this:
`beforeEach(angular.mock.inject(function (_$rootScope_, _$compile_, _PostFilters_, _FocusTrap_) ` (its confusing, I know :D ).

For the mock, since we are not testing the MutationObserver functions in this test, I actually think we can do this:
https://stackoverflow.com/a/50626200

Simply add that to the filter-posts.directive.spec.js, before the beforeEach and it should hopefully pass. That adds a new mock-MutationObserver to the file.

I hope this unblocks you, it was an unexpected tricky one 👍

Ifycode:
Hmm. Alright, thank you. I will try it out.

Ifycode:
Wow! It works @Angamanga 
Although I put it in the inner describe function instead, because when it's in the outer describe function (before the beforeEach functions), some times it shows this error:
````
PhantomJS 2.1.1 (Windows 7.0.0) setting create targeted survey controller controller-functions Step Three Completion should fail to complete the step when no questions is added FAILED
        Error: <toHaveBeenCalled> : Expected a spy, but got Function.
        Usage: expect(<spyObj>).toHaveBeenCalled() thrown
````
I don't know if this is just on my computer. But the inner describe gives me no issue.

Angamanga (approved):
Approved! Works fine now 🎉

Angamanga:
@Obadha2 Ready for QA 🎉

Angamanga (approved):

Angamanga:
@Obadha2 I need this one before fixing the date-picker issue since it affects the same code. Could you have a look?

PR #1664, opened 2021-04-13
fix language not showing in side bar

helen-ndip:
-

Testing checklist:
- [ x] Log in and navigate to  /settings/categories/create, the drop-down now has the default language selected.
-  [ x] Log in and navigate to  /settings/surveys/create, the drop-down now has the default language selected.
-  [ x] Language select on the sidebar now has the default language.
- [ x] I certify that I ran my checklist

Fixes ushahidi/platform#4227.

Ping @ushahidi/platform

### DIFF
--- a/app/common/directives/language-switch.directive.js
+++ b/app/common/directives/language-switch.directive.js
@@ -17,11 +17,17 @@ function LanguageSwitchController($scope, Languages, TranslationService, $locati
     $scope.changeLanguage = changeLanguage;
     $scope.$on('event:authentication:login:succeeded', TranslationService.setStartLanguage);
     $scope.$on('event:authentication:logout:succeeded', TranslationService.setStartLanguage);
+    $scope.site = {};
     activate();
 
     function activate() {
         Languages.then(function (languages) {
             $scope.languages = languages;
+            TranslationService.getLanguage().then(lang =>{
+                let deflangs = languages.filter(language => (language.code).toString().split('-')[0] === lang.split('-')[0]);
+                $scope.site.language = deflangs.length === 1 ? deflangs[0].code : lang;
+            })
+
         });
     }
 
--- a/app/settings/categories/categories-edit.html
+++ b/app/settings/categories/categories-edit.html
@@ -123,9 +123,7 @@
                     <select
                           ng-model="selectedLanguage"
                           ng-options="lang.code as lang.name for lang in languagesToSelect"
-                          ng-value="lang.code"
                           ng-change="selectLanguage(selectedLanguage)"
-                          ng-selected="lang.code === selectedLanguage"
                       >
                       </select>
                 </div>
--- a/app/settings/surveys/survey-editor.directive.js
+++ b/app/settings/surveys/survey-editor.directive.js
@@ -183,6 +183,7 @@ function SurveyEditorController(
                         surveyLanguages: [$scope.survey.enabled_languages.default, ...$scope.survey.enabled_languages.available]
                     }
                     }
+                    $scope.selectedLanguage = language;
                 });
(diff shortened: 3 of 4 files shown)


Angamanga:
@helen-ndip I am so sorry I thought I commented on friday but apparently not. Could you please try to only commit the actual changes to this fix? I see a lot of cosmetic changes such as indentation and row-breaks and that makes it really hard to know what has to do with the issue itself :) 👍

helen-ndip:
> @helen-ndip I am so sorry I thought I commented on friday but apparently not. Could you please try to only commit the actual changes to this fix? I see a lot of cosmetic changes such as indentation and row-breaks and that makes it really hard to know what has to do with the issue itself :) 👍 If you remove those changes and make a new commit it should be fine.

Right away.

helen-ndip:
> @helen-ndip I am so sorry I thought I commented on friday but apparently not. Could you please try to only commit the actual changes to this fix? I see a lot of cosmetic changes such as indentation and row-breaks and that makes it really hard to know what has to do with the issue itself :) 👍 If you remove those changes and make a new commit it should be fine.

I am done here. Pushed a new commit.

app/common/directives/language-switch.directive.js
```diff
@@ -17,11 +17,17 @@ function LanguageSwitchController($scope, Languages, TranslationService, $locati
     $scope.changeLanguage = changeLanguage;
     $scope.$on('event:authentication:login:succeeded', TranslationService.setStartLanguage);
     $scope.$on('event:authentication:logout:succeeded', TranslationService.setStartLanguage);
+    $scope.site = {};
```
Angamanga, line 20: `$scope.site` is sent from the directive using the language-switch (see the parameter here: https://github.com/ushahidi/platform-client/blob/develop/app/common/directives/language-switch.directive.js#L9). You don't need to define it here since it overrides the information sent into the directive. This may be a bit confusing but the main reason is that we often have this information already in the place where we are using the directive, and also when using a directive in many different places, we want to have the flexibility to choose what information to send to the directive as a starting value. That is not possible if we define it again inside of the directive.

Instead, pass the values we need as a parameter from the mode-context-form-filter. We fetch the language here: https://github.com/ushahidi/platform-client/blob/develop/app/main/posts/views/mode-context-form-filter.directive.js#L94 and the variable holding the information we want is called `$scope.userLanguage`. Therefore, we can pass this information to the language-switch like this:
`<language-switch site="{language: userLanguage}"></language-switch>`

app/common/directives/language-switch.directive.js
```diff
@@ -17,11 +17,17 @@ function LanguageSwitchController($scope, Languages, TranslationService, $locati
...
     activate();
 
     function activate() {
         Languages.then(function (languages) {
             $scope.languages = languages;
+            TranslationService.getLanguage().then(lang =>{
```
Angamanga, line 26: This can be removed if you do above change.

app/settings/categories/categories-edit.html
```diff
@@ -123,9 +123,7 @@ <h1 class="mode-context-title" ng-show="!category.id" translate>app.add_category
                     <select
                           ng-model="selectedLanguage"
                           ng-options="lang.code as lang.name for lang in languagesToSelect"
-                          ng-value="lang.code"
```
Angamanga, line 126: Nice catch and solution 🎉

Angamanga (changes requested):
Just a small change for the language switch, see my comment in the code. Let me know if you need any clarification 👍

helen-ndip:
Hi @Angamanga  I am closing this because this branch is far behind develop. I will push a new branch with the requested fixes.

PR #1668, opened 2021-04-24
4044 - Remove or fix default values that maintain previous behavior

### LINKED ISSUE #4044: Remove default value for fields that doesn't use it
**Describe the bug**
There is an input field for default-values for all field-types but not all fields can use a default value. This is confusing to the users and causes bugs. 
TODO:
- Check which fields that use the default values
- Remove the possibility to set default value for the fields that don't

Aha! Link: https://ushahiditeam.aha.io/features/PROD-68

Ifycode:
**This pull request makes the following changes**:
* Fixes issue ushahidi/platform#4044 

**Details**:

Removes or fixes [fields whose default values maintain their their previous (buggy) behaviors](https://github.com/ushahidi/platform/issues/4044#issuecomment-822915377) - except for date, date and time. 

Fields involved in this PR: 
* Location 
* Related posts 
* Integer 
* Decimal

**Removed these**: Removed location and related posts with `canDisplayDefaultValue()`.

**Problem with numbers (integer and decimal) and solution**:

u-form-int

Doesn't switch for int or decimal based on findings from the [attribute-create directive](https://github.com/ushahidi/platform-client/blob/develop/app/settings/surveys/attribute-create.directive.js). Therefore, it defaults to displaying input type text instead - screenshot above. Added switch for checking `editField.type` in the case of numbers, and then conditioned input type text to only show up when input is not number for it to work as expected.

**Testing checklist**:
- [x] Check that default value input have been removed for fields that don't use it.
- [x] Check that others (integer and decimal) now show up on survey post created.

Fixes ushahidi/platform# .

Ping @ushahidi/platform


Ifycode:
Hi @Angamanga happy weekend. I have some questions.

**Question A** - Is on date and date & time. The solution I added for date works, but is just added temporarily (since it just uses the native browsers date picker I think). I also added it to help better explain while I ask this question. So here is the problem - For date and time to work, the user must type in the exact thing at the right place for it to be added correctly in [the input on preview page] and then finally in the post created - that is if we stick to using input type text. I'm thinking a date (and time) picker could work here better instead of adding validation on input type text as was the case before - The user won't have to bother about the number of spaces to be added etc. I also noticed a [pull request](https://github.com/ushahidi/platform-client/pull/1655/files) has been submitted for changing the picker... I wanted to get your opinion on this first. Do you think a date (and time) picker might work here? or would you prefer I go on to manually configure the input with javacript?

**Question B** - Is about app/common/locales/en.json file. I noticed there are other json files for translations I guess. Is there some mechanism set in place to translate and send the translation into other files? or I shouldn't bother about those?

Angamanga:
@Ifycode For question A, you can actually skip the validation for date and time for now, a pull request with a new date picker is waiting for QA so we can look into that once that PR is released.

Question B You don't need to bother about that, they are set automatically when a string is translated.

Ifycode:
Alright. I've restored anything related to date back to it's intial state.

PR #1674, opened 2021-05-31
4258 - Add webpack-bundle-analyzer

Ifycode:
- Also part of solution for ushahidi/platform#4258
- Fixes issue ushahidi/platform#4246 

#### Details:
- Adds `webpack-bundle-analyzer` plugin to analyze dependencies in bundles generated. Generates a stats.json file, and opens up the browser, when you run `gulp` or `npm run build` commands.

#### Testing checklist:
- [x] webpack-bundle-analyzer is installed.

Ping @ushahidi/platform


Angamanga (changes requested):
Thank you for adding this plugin. I really like it! However, I think we could move the use of it to a specific gulp-task, so we only run it when we want to check something. In that way, the stats are not generated and the webpage doesn't open every time we start the dev-server. Take a look at this gulp-function: https://github.com/ushahidi/platform-client/blob/develop/gulpfile.babel.js#L152 and the task "default": https://github.com/ushahidi/platform-client/blob/develop/gulpfile.babel.js#L188. We could add a new function and task that adds the plugin to the config in a similar way as we set the entry when starting the devServer.

An example of how you run many functions in a series in a gulp-task can be seen here: https://github.com/ushahidi/platform-client/blob/develop/gulpfile.babel.js#L316

Let me know if you need any guidance in setting the task and function up :) 👍

Ifycode:
Hmm. Interesting. Thank you. I will reach out to you once I'm stuck at any point, when I begin work on this.

Ifycode:
@Angamanga I have written gulp tasks for the analyzer without generating the stats.json. Can you help take a look at this too?

PR #1687, opened 2021-06-30
Split and lazy load activity module

Ifycode:
#### This pull request makes the following changes
- Fixes ushahidi/platform#4289

#### Details
- Add oclazyload
- Split and lazy load activity view/module

#### Testing checklist
- [x] Go to map page localhost:9090.
- [x] Open up browser dev tool (network tab).
- [x] Click on activity-view from the modebar.
- [x] Check that activity.js file shows up.

Ping @ushahidi/platform


Ifycode:
@Angamanga here is the new pull request created for that of activity-view.

legacy/app/activity/activity-routes.js
```diff
@@ -0,0 +1,3 @@
+angular.module('ushahidi.activity.routes', [])
+
+.config(require('./activity-routes-lazyload.js'));
```
Angamanga, line 3: You could add the whole "activity-routes-lazyload.js" to the config-parenthesis here instead of requiring the file, maybe that makes things more intuitive? In that case, you can remove the activity-routes-lazyload.js entirely.
Like this:

```
angular.module('ushahidi.activity.routes', [])
.config(
    [
        '$stateProvider',
        '$urlMatcherFactoryProvider',
    function (
        $stateProvider,
        $urlMatcherFactoryProvider
    ) {
        $urlMatcherFactoryProvider.strictMode(false);
        $stateProvider
        .state({
            name: 'activity',
            url: '/activity',
            controller: require('./activity.controller.js'),
            template: require('./activity.html'),
            lazyLoad: function ($transition$) {
                const $ocLazyLoad = $transition$.injector().get('$ocLazyLoad');
                return System.import('/activity.js').then(mod => {
                    $ocLazyLoad.load(mod.ACTIVITY_MODULE);
                });
            }
        });
    }]
);
```

Angamanga (commented):
Just one comment, see the code, you can decide if you want to change it or keep it the way it is :)

legacy/app/activity/activity-routes.js
Ifycode: Honestly, I thought of this. I also think it will make this part less complicated. I like the idea of using this method.

Angamanga (approved):
Approved! Well done!

PR #1693, opened 2021-07-29
Experiment 1 of 2: Move posts module remnants to map module

Ifycode:
Move collection, filters and saved searches to map-view. No lazy load. This works. Merge into splits-view branch?

https://github.com/ushahidi/platform-client/pull/1693#issuecomment-888772961

PR #1694, opened 2021-07-29
Experiment 2 of 2: Lazy load posts remnant after moving to map

Ifycode:
#### Experimenting

Collections, filters and savedSearches are only visible when added to map. Meaning map bundle will increase in size a bit. Trying to see if it's possible to reduce the size of the map bundle a bit by lazyloading these 3 when needed. On click or on focus.

#### Observation/issue:

|Screenshot 2021-07-29 at 2 52 27 AM|Screenshot 2021-07-29 at 2 52 39 AM|
|--|--|

Looks like once it's cut from the map module, the structure for filters and collections disappears. But if you remove the comment from line 2: https://github.com/Ifycode/platform-client/blob/24709c780342daeb7d76e312c4b76b66472257cd/legacy/app/map/map-module.js#L2 and line 42: https://github.com/Ifycode/platform-client/blob/24709c780342daeb7d76e312c4b76b66472257cd/legacy/app/map/map-module.js#L42 the structure for the affected parts show up as it should.

#### My conclusion:
Lazy loading doesn't work. Just use code in [Experiment 1 of 2: Move posts module remnants to map module](https://github.com/ushahidi/platform-client/pull/1693) instead.


Angamanga:
@Angamanga For reference. Contains experiments with lazyload on focus that might be useful!

PR #1697, opened 2021-08-11
Fix error from last merge to feature/migration

Ifycode:
- Fixes error from last merge to feature/migration (removes main module from app.js, turns out one of the pull requests contained the commit that moved donation into common module. Main module no longer exists as a result. This causes error when running npm run serve).

#### Screenshots from feature/migration
Screenshot 2021-08-11 at 5 58 07 AM

Ping @ushahidi/platform


Ifycode:
@Angamanga could you merge this in so that the issue is fixed before my next pull request?

PR #1706, opened 2021-09-01
Using chart.js instead of d3

Angamanga:
- Using chart.js instead of  d3 in activity-view
- Updating configs and data formatting to fit chart.js

Testing checklist:
This pr does not change the behaviour of the activity-view, just the library used to display the graphs (and a little bit of the appearance, for example, the bars are now vertical instead of horizontal, but they still display the same thing). The easiest way of testing is to look at the activity-view before merging this pr, take note of the behaviour for different filters, merge the pr and compare the filters. They should reflect the same thing. Or merge this pr and compare the behaviour with another deployment. 

Note: There are known issues with the activity-view (see comment below) and they are not fixed by this PR, they will be handled separately.

Fixes ushahidi/platform#4258 and ushahidi/platform#429  .

Ping @ushahidi/platform


Angamanga:
@tuxpiper @webong This PR contains the changes needed to switch from d3 to chart.js. When working on this I found some odd things in the responses from the API. When calling the Post-stats-endpoint with some of the filters used, I get results I cannot figure out. Could we try to clarify or identify if there is a bug?

List of the weirdnesses:
1. When calling the `/api/v3/posts/stats?group_by=&timeline=1&timeline_attribute=created` it looks like I could expect post-count/day a post has been submitted during the time the deployment has existed. For annatest1000.ushahidi.io I get back 2 posts only which is a bit odd since there are 22620 posts in my deployment. 
2. When calling `/api/v3/posts/stats?group_by=tags&timeline=1&timeline_attribute=created` I always get an empty array back. It looks like what the client expects is the number of posts added/day grouped by categories.
3. When calling `/api/v3/posts/stats?group_by=form&timeline=1&timeline_attribute=created` I get the survey names and a total of 1 for each survey, which I suspect is the number of surveys created for each survey but not sure since not all my surveys are included in the response 🤔 . The value I would like is the number of posts/day grouped by survey.
4. When calling `/api/v3/posts/stats?group_by=tags` I also always get an empty array back, but what I expect is the total number of posts grouped by categories.

Either the API has changed over time and we have forgotten to update the activity view or the API has a bug in some way. Not sure if we should merge this before we have figured it out? The behaviour already exists in production so we would not be introducing a new bug, but still, it's annoying to push something half-broken.

Angamanga:
@tuxpiper @webong Ready for review 🎉

Ifycode:
Yay!

app/main/activity/bar-chart.directive.js
```diff
@@ -89,13 +47,87 @@ function ActivityBarChartController($scope, $translate, PostEndpoint, d3, _, Pos
...
-            if (results.totals[0]) {
-                results.totals[0].key = $scope.options.chart.yAxis.axisLabel;
-            }
             $scope.data = results.totals;
+            generateGraph();
             $scope.isLoading = false;
         });
     }
 
+    function getRandomColour() {
```
tuxpiper, line 56: I just smiled at the simplicity of this function :)

Something for the future (not this PR), but maybe let's look for a palette of colors to randomly pick from, for better visual results.

tuxpiper (commented):
Leaving the approval up to @webong  . I just added a little comment of mine :) Excited to see this!

Angamanga:
@webong Could you do a final review? 🙏

Angamanga:
@tuxpiper QA passes, just need an approval for the last small changes (using dayjs instead of moment in activity-view)

tuxpiper (approved):
Yes, I like it 👍  ( @webong  please take look too 🙏 )

webong (approved):

PR #1714, opened 2021-09-23
Fix(auth): Using refresh-token to renew session

Angamanga:
- Adds a timer if there is an expiry-date and refresh-token available
- Renews the auth-token before session ends (in the background, user should not notice)

Testing checklist:
- Login
- Take note of the expiry-time in the Session
- Start adding a post or a survey
- Leave tab and post/survey open over the expiry-time
- Save post/survey after the first expiry-time
- [ ] Post/survey should save
- [ ] If a post, your logged in user should be the author of the post
- [ ] You should not lose your data

- Logout
- Login again
- Close the browser
- Open the browser and open the page
- Check that you are still logged in
- Take note of the expiry-time in the Session
- Start adding a post or a survey
- Leave tab and post/survey open over the expiry-time
- Save post/survey after the first expiry-time
- [ ] Post/survey should save
- [ ] If a post, your logged in user should be the author of the post
- [ ] You should not lose your data
  
Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/app/common/auth/authentication.service.js
+++ b/app/common/auth/authentication.service.js
@@ -66,6 +66,49 @@ function (
         loginStatus = false;
     }
 
+    function handleRequestSuccess(authResponse) {
+        var accessToken = authResponse.data.access_token;
+        Session.setSessionDataEntry('accessToken', accessToken);
+        if (authResponse.data.expires_in) {
+            Session.setSessionDataEntry('accessTokenExpires', Math.floor(Date.now() / 1000) + authResponse.data.expires_in);
+        } else if (authResponse.data.expires) {
+            Session.setSessionDataEntry('accessTokenExpires', authResponse.data.expires);
+        }
+        // Setting a timer to refresh the session if a refresh-token is available
+        if (authResponse.data.refresh_token) {
+            scheduleRefreshAuth(authResponse.data.refresh_token);
+        }
+        Session.setSessionDataEntry('grantType', 'password');
+    }
+
+    function handleRequestError () {
+        setToLogoutState();
+        $rootScope.$broadcast('event:authentication:login:failed');
+    };
+
+    function scheduleRefreshAuth (refreshToken) {
+        if (Session.getSessionDataEntry('accessTokenExpires')) {
+            /* Getting a random number between 5 minutes and 11 minutes before the expiry time.
+            (to avoid simultanous requests if the user has 2 or more tabs open)*/
+            const minutesBefore = (Math.floor(Math.random() * 11) + 5) * 60;
+            const timeToExpire = Session.getSessionDataEntry('accessTokenExpires') - Date.now() / 1000;
+            const timeToRefresh = timeToExpire - minutesBefore;
+            // Setting the timer to refresh token
+            setTimeout(function () {
+                let payload = {
+                    refresh_token: refreshToken,
+                    grant_type: 'refresh_token',
+                    client_id: CONST.OAUTH_CLIENT_ID,
+                    client_secret: CONST.OAUTH_CLIENT_SECRET,
+                    scope: CONST.CLAIMED_USER_SCOPES.join(' ')
+                };
+                //TODO: If we want to show the login-box if the refresh-token is invalid, use error-handler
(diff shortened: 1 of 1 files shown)


tuxpiper (dismissed):
I've added changes and notes as discussed when we were pairing on this @Angamanga  . Would you like to take another look before we merge?

Angamanga:
@tuxpiper Works fine for newly logged in windows and closed and re-opened windows. 2 tabs at the same time seem to log the person out though, but I think that was expected so this is at least an improvement 👍

Angamanga:
@tuxpiper So have removed the request for the client-credentials token. It was triggered when the geojson-request fails (which is something we probably also need to look into) and then messed up the refresh-flow. However, left a few questions in the code, where I am not sure what to do. Lets discuss, either async or pairing!

tuxpiper:
> @tuxpiper So have removed the request for the client-credentials token. It was triggered when the geojson-request fails (which is something we probably also need to look into) and then messed up the refresh-flow. However, left a few questions in the code, where I am not sure what to do. Lets discuss, either async or pairing!

:+1: let's do async for now, I will take a look during the next couple of days

legacy/app/auth/authentication-interceptor.config.js
```diff
@@ -122,29 +52,16 @@ function AuthInterceptor($rootScope, $injector, $q, CONST, Session, _) {
...
                         return deferred.promise;
                     }
-
-                    concurrentGetClientCredsToken(rejection.config).then(
-                        function (config) {
-                            deferred.resolve($http(config));
-                        },
-                        deferred.reject
-                    );
+                    // Here we got the client-creds token previously. What to do now? Just reject the promise?
```
tuxpiper, line 61: > What to do now? Just reject the promise?
I think so. This function seems to be about error handling, so in that case it makes sense to me to exit it by propagating the error (to other error handlers, I assume)

In general, my opinion is that error handling routines should be as simple as possible. For that reason, I wouldn't say I like the original approach of getting a client token here... I'm glad we are getting rid of it.

legacy/app/auth/authentication.service.js
```diff
@@ -27,16 +27,25 @@ function (
...
-        Session.getSessionDataEntry('grantType') === 'password' &&
         !!Session.getSessionDataEntry('userId')
     ) {
         // If the access token is expired
         if (Session.getSessionDataEntry('accessTokenExpires') <= Math.floor(Date.now() / 1000)) {
+            /*
+             * TODO: if there's a refresh token in the session, we should attempt obtaining a new
+             *       access token. BUT should we intercept and defer other requests until we obtain
+             *       the token?
+             */
```
tuxpiper, line 38: > BUT should we intercept and defer other requests until we obtain the token?
I am a bit torn about that... part of me says yes, but another says it may be avoidable complexity. I think we need to reason it a bit more. So, maybe just not for this PR?

In general, I'm wondering if we wouldn't be better off with a fresh approach in the future. I can think of other corner cases that don't seem to be well handled, such as dealing with flaky network connections.

legacy/app/auth/authentication.service.js
```diff
@@ -61,13 +70,68 @@ function (
...
 
     function setToLogoutState() {
+         let refreshToken = Session.getSessionDataEntry('refreshToken');
+
+         let payload = {
+             refresh_token: refreshToken,
+             client_id: CONST.OAUTH_CLIENT_ID,
+             client_secret: CONST.OAUTH_CLIENT_SECRET
+         };
+         // Do we need to cancel the scheduled refresh?
```
tuxpiper, line 80: > // Do we need to cancel the scheduled refresh?
Yes I think so..  or at the very least make sure it will do nothing if it triggers

tuxpiper (approved):

legacy/app/auth/authentication-interceptor.config.js
Angamanga: Agreed!

legacy/app/auth/authentication.service.js
Angamanga: Yep, I think so too, time has passes and it looks like there were plans to continue working on this but that never happened. I think a good point working on this is when we migrate this part of the code.

PR #1722, opened 2021-10-20
Adjusting import map

Angamanga:
- Hashing the names of the js-files in production-builds.
- Using keys from import-map to lazy-load modules in legacy-app
- Generating importmap for production-file-names when building the app

Testing checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


Angamanga:
@tuxpiper I somehow messed up something when building it looks like, I'll let you know when I have figured it out...

PR #1724, opened 2021-11-08
Add gtm

Angamanga:
- Adding google-tag-manager script if window.ushahidi.googleTagManager is present
- Moving google-scripts to end of body-tag

Fixes ushahidi/platform# .

Ping @ushahidi/platform


tuxpiper (approved):
In general looks good to me, but just wondering:

* why is the addition of googleTagManager repeated? I see it both in lines 100 and 198
* didn't we have analytics and maps before? Why are they added here but not removed from anywhere else?

Angamanga:
@tuxpiper The duplicated scripts are now removed :D

tuxpiper (dismissed):
Looks good to me ! :+1:

tuxpiper (approved):

PR #1756, opened 2021-12-03
fix: Saving data from date-fields in the format YYYY-MM-DD

Angamanga:
- Initialises and stores data from date-fields in the format YYYY-MM-DD
- Removes the type="date" from date-fields since the validator does not like the new format

Testing checklist:
- Add a new post to a survey with date-fields
- Select a date
[ ] Look at the data sent to the API, it should be in the format YYYY-MM-DD without any time-stamp
[ ] The "value_meta" should contain the time-zone
- Edit the same post
- Make some changes to the dates
- Save
- Look at the data sent to the API
[ ] The "value" should be in the format YYYY-MM-DD without any time-stamp
[ ] The "value_meta" should contain the time-zone

- Add a new post to a survey with a date-time-field
- Select a date and time
- Look at the data sent to the API,
[ ] The value should be the same date and time added converted to UTC.
[ ] The "value_meta" should contain the time-zone and the offset in minutes
- Edit the same post
[ ]  The pre-filled values should be the same as the value added to the post in the first place <-- this one fails!
- Make some changes to the date and time
- Save
- Look at the data sent to the API,
[ ] The value should be the same date and time added converted to UTC.
[ ] The "value_meta" should contain the time-zone and the offset in minutes
 

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/legacy/app/data/common/post-edit-create/post-datetime-value.directive.js
+++ b/legacy/app/data/common/post-edit-create/post-datetime-value.directive.js
@@ -10,13 +10,6 @@ module.exports = [
             require: 'ngModel',
             template: require('./post-datetime-value.html'),
             link: function ($scope, element, attrs, ngModel) {
-                // Split date time in time and date fields
-                $scope.timeOptions = {
-                    format: 'HH:i',
-                    interval: 15,
-                    onClose: save
-                };
-                $scope.dateOptions = { format: 'yyyy-mm-dd', onClose: save };
                 $scope.model = null;
 
                 // If no ngModel, skip the rest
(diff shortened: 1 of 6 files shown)


tuxpiper:
@Angamanga I've made a first quick test, and it seems to be essentially working 👍 

I suspect there are a few loose ends to tie: I see javascript console errors when loading the data view and some weird behavior when editing a post (the title and description of the post temporarily switch to show as "undefined" in the data view, until the view is reloaded).

Could you please check?

tuxpiper:
I think this also may be doing something to datetimes, I no longer seem to be able to set a time. Regardless of what I place in the picker, "12:00:00Z" is always sent.

Angamanga:
@tuxpiper odd, this should only affect the date-field. But checking 😬

PR #1758, opened 2022-01-11
fix: Saving data from date-fields in the format YYYY-MM-DD

Angamanga:

Initialises and stores data from date-fields in the format YYYY-MM-DD
Removes the type="date" from date-fields since the validator does not like the new format
Testing checklist:

Add a new post to a survey with date-fields

Select a date
[ ] Look at the data sent to the API, it should be in the format YYYY-MM-DD without any time-stamp
[ ] The "value_meta" should contain the time-zone

Edit the same post
Change the date
Save

Look at the data sent to the API
[ ] The "value" should be in the format YYYY-MM-DD without any time-stamp
[ ] The "value_meta" should contain the time-zone

- Change time-zone on your computer
- Edit the same post again but do not touch the date-field
[ ] The "value" should be in the format YYYY-MM-DD without any time-stamp
[ ] The "value_meta" should contain the time-zone that the post was originally saved in (it should not update the timezone to your new setting)

Edit the same post again and make a change to the date
Save

Look at the data sent to the API
[ ] The "value" should be in the format YYYY-MM-DD without any time-stamp
[ ] The "value_meta" should contain the new time-zone

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/legacy/app/data/common/post-edit-create/post-value-edit.directive.js
+++ b/legacy/app/data/common/post-edit-create/post-value-edit.directive.js
@@ -49,7 +49,7 @@ function PostValueEditController($rootScope, $scope, _, Flatpickr, SurveysSdk) {
         addDefaultValue();
         if (isDate($scope.attribute)) {
             $scope.$watch('attribute.value.value', (newValue, oldValue) => {
-                if (newValue !== oldValue) {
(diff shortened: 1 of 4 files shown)


tuxpiper (approved):
This is looking good to me 👍

PR #1760, opened 2022-01-31
Fix: Not using default date if invalid

Angamanga:
- Ignores default date if its an Invalid Date on editing and creating a post

Testing checklist:
- Add a new post to a survey with a date-field with a default-value that is not a date (for example a text-string)
- Look at the date-field
- [ ] It should be empty and not say "Invalid Date"
- Edit a post to a survey with a date-field with a default-value that is not a date (for example a text-string). The post should not have any data for the date-field
- Look at the date-field
- [ ] It should be empty and not say "Invalid Date"

Fixes ushahidi/platform# .

Ping @ushahidi/platform

### DIFF
--- a/legacy/app/data/post-create/post-editor.directive.js
+++ b/legacy/app/data/post-create/post-editor.directive.js
@@ -132,8 +132,11 @@ function PostEditorController(
                             attr.value.value = dayjs(attr.value.value).format('YYYY-MM-DD');
                         } else if (attr.default) {
                             try {
-                                attr.value.value = dayjs(new Date(attr.default).format('YYYY-MM-DD'));
-                            } catch (err) {
+                                let defaultValue = dayjs(new Date(attr.default)).format('YYYY-MM-DD');
+                                // Safeguarding against invalid default dates below. We should add validation in the survey setup instead
+                                if (defaultValue !== 'Invalid Date') {
+                                    attr.value.value = defaultValue;
+                                }                            } catch (err) {
                                 // What do do if the default-value is in the wrong format? Probably validate that field in the first place?
                             }
                         } else {
--- a/legacy/app/data/post-create/post-editor.directive.js
+++ b/legacy/app/data/post-create/post-editor.directive.js
@@ -132,14 +132,16 @@ function PostEditorController(
                             attr.value.value = dayjs(attr.value.value).format('YYYY-MM-DD');
                         } else if (attr.default) {
                             try {
-                                let defaultValue = dayjs(new Date(attr.default)).format('YYYY-MM-DD');
+                                let defaultValue = dayjs(new Date(attr.default))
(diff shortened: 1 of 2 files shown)


legacy/app/data/post-create/post-editor.directive.js
```diff
@@ -132,8 +132,11 @@ function PostEditorController(
                             attr.value.value = dayjs(attr.value.value).format('YYYY-MM-DD');
                         } else if (attr.default) {
                             try {
-                                attr.value.value = dayjs(new Date(attr.default).format('YYYY-MM-DD'));
-                            } catch (err) {
+                                let defaultValue = dayjs(new Date(attr.default)).format('YYYY-MM-DD');
+                                // Safeguarding against invalid default dates below. We should add validation in the survey setup instead
+                                if (defaultValue !== 'Invalid Date') {
+                                    attr.value.value = defaultValue;
+                                }                            } catch (err) {
```
tuxpiper, line 139: Please check the indentation here, I don't usually look too much into it but this is a bit too much off :)

legacy/app/data/post-create/post-editor.directive.js
```diff
@@ -132,8 +132,11 @@ function PostEditorController(
                             attr.value.value = dayjs(attr.value.value).format('YYYY-MM-DD');
                         } else if (attr.default) {
                             try {
-                                attr.value.value = dayjs(new Date(attr.default).format('YYYY-MM-DD'));
-                            } catch (err) {
+                                let defaultValue = dayjs(new Date(attr.default)).format('YYYY-MM-DD');
+                                // Safeguarding against invalid default dates below. We should add validation in the survey setup instead
+                                if (defaultValue !== 'Invalid Date') {
```
tuxpiper, line 137: Are we reasonably confident this is going to be "Invalid Date" regardless of environment, new versions etc.. ?  Is there something easily available but also less likely to change? (i.e. an object type or property)

tuxpiper (changes requested):

tuxpiper:
Would it be safe to assume we should follow up with one like this, but for datetime?

legacy/app/data/post-create/post-editor.directive.js
Angamanga: That is awful! Fixed!

legacy/app/data/post-create/post-editor.directive.js
Angamanga: Good point. Changed this a bit.

tuxpiper (approved):
Looks good to me 👍

Angamanga:
@Obadha2 Ready for QA 🎉

tuxpiper:
Goes together with ushahidi/platform#4395

PR #1777, opened 2022-04-03
close notifications after a set timeut of 6s or when user navigates f…

helen-ndip:
…rom a page

-

Testing checklist:
- [x] automatically dismiss notification when user navigates away from a page.
- [x] automatically dismiss notification after 6s if user decides to stay on a page.

Fixes ushahidi/platform#4384.

Ping @ushahidi/platform

### DIFF
--- a/legacy/app/common/notifications/slider.service.js
+++ b/legacy/app/common/notifications/slider.service.js
@@ -46,13 +46,13 @@ function SliderService($rootScope, $q, $templateRequest) {
     function closeOnNavigationOrTimeout() {
         setTimeout(() => {
             close();
-            window.removeEventListener("popstate", sliderClose, false);
+            window.removeEventListener('popstate', sliderClose, false);
         }, 6000);
 
         let sliderClose = () => {
             close();
-            window.removeEventListener("popstate", sliderClose, false);
+            window.removeEventListener('popstate', sliderClose, false);
         }
-        window.addEventListener("popstate", sliderClose, false);
+        window.addEventListener('popstate', sliderClose, false);
     }
 }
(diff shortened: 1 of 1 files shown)


Angamanga (changes requested):
Thank you for your contribution @helen-ndip! I like the solution but I would like to remove the timeout and only keep closing it for navigation-changes. For some of the notifications, the slider actually needs to be open to prevent the user from missing it, for example csv exports so its actually better to keep it.

Ako-Njang:
> Thank you for your contribution @helen-ndip! I like the solution but I would like to remove the timeout and only keep closing it for navigation-changes. For some of the notifications, the slider actually needs to be open to prevent the user from missing it, for example csv exports so its actually better to keep it.

Ok I think so too. I will update that right away.

helen-ndip:
> Thank you for your contribution @helen-ndip! I like the solution but I would like to remove the timeout and only keep closing it for navigation-changes. For some of the notifications, the slider actually needs to be open to prevent the user from missing it, for example csv exports so its actually better to keep it.

I think so too. I will update that right away.

helen-ndip:
@Angamanga  I have done the suggested changes.

helen-ndip:
> Thank you for your contribution @helen-ndip! I like the solution but I would like to remove the timeout and only keep closing it for navigation-changes. For some of the notifications, the slider actually needs to be open to prevent the user from missing it, for example csv exports so its actually better to keep it.

@Angamanga  do I still need to make some more changes? I think I am done with this

Angamanga (approved):
Thank you for your contribution 🙏 Approved!

Angamanga:
@Obadha2 Ready for QA.

Obadha2:
@Angamanga QA'd, passes. 👍

Angamanga:
@tuxpiper To include in next relase!

tuxpiper:
Wonderful 👌, thank you all!

PR #1779, opened 2022-04-13
Fixed the deployment to heroku

ps-19:
After brain storming for many hours on heroku and repo i got the silly reason for the issue. 

Testing checklist:
- [x] I have tested the build.

Fixes ushahidi/platform#[4288](https://github.com/ushahidi/platform/issues/4288).

Ping @ushahidi/platform

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ All our issues live in https://github.com/ushahidi/platform/issues .
 
 ## Try it out on Heroku
 
-[![Deploy](https://www.herokucdn.com/deploy/button.png)](https://heroku.com/deploy?template=https://github.com/ushahidi/platform-client/tree/master)
+[![Deploy](https://www.herokucdn.com/deploy/button.png)](https://heroku.com/deploy?template=https://github.com/ushahidi/platform-client/tree/develop)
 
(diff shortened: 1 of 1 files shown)


Angamanga:
@tuxpiper Could you review this? 🙏

tuxpiper:
I think that's fine, but since it's the develop branch, a disclaimer should be added that the deployed site may not work or be usable (unstable version).

Otherwise, it should point to the "main" branch instead.

Could you please adjust @ps-19 ?

tuxpiper (changes requested):
Changes requested

ps-19:
@tuxpiper sir where should i add the disclaimer, in the readme file. And even if it is pointing towards `main` still "No app.json located in the repo URL provided" is showing.

tuxpiper:
Oh I see, so it looks like we are actually missing more bits here in order to make this PR viable..

@Angamanga do you recall if we had that app.json before and was removed as part of the Single-SPA changes?

tuxpiper:
I think that without that missing `app.json` file, we should close this PR. Anyone willing to spend a few minutes, I'm pretty sure they can fish that out from previous commits.

PR #1790, opened 2022-05-18
Adding loading-bar to survey-view

Angamanga:
- Adds a loading-bar/line to the create-survey-screen.

Testing checklist:
- Click on the + sign and select a survey to add a new post to
- [ ]  While the survey is loading, an orange bar/line that is moving should be visible just below the toolbar.

Fixes ushahidi/platform# .

Ping @ushahidi/platform


webong (approved):

webong:
All good to me! 

There's a side-effect I notice though, once I'm done adding a post to a survey, the orange bar/line loader shows, but even after the post is submitted and redirects to its post-view, the load line doesn't go away (keeps loading).

Angamanga:
@webong Yea, that is the functionality of the bar I think, already present in production. Since we use the same bar for most loading events, it shows up whenever a request is done and the client is waiting for a response from the API. It could be loading posts, loading map, checking for exports etc etc.

tuxpiper:
@Angamanga @webong  can we say this is good for QA?

tuxpiper:
@AmTryingMyBest could you please test this at the earliest opportunity? 🙏

PR #1796, opened 2022-06-09
Poc modebar

Angamanga:
- Adds an example of modebar in react. WIP. For more info, read here: https://www.notion.so/ushahidiorg/Platform-client-ac8e4d0cdd8045c7aaab2560603c6c19

Testing checklist:
- [ ]

Fixes ushahidi/platform# .

Ping @ushahidi/platform


tuxpiper:
To do for completion:

* use i18n strings
* plug in to authentication, so that only relevant icons are shown, according to user roles
   - choice: moving the authentication away from legacy first or last
     - probably last, because that way we need to change less things in the legacy app
     - authentication state is available anyway from local storage
* review the styles architecture, because dealing with the root DOM shouldn't be done inside modules , breaks separation of concerns

PR #1807, opened 2022-07-28
Jumping Map View

webong:
- Modify the `addPostsToMap` function to accepts a boolean parameter that determines if the status of the received posts
- Avoid looping over geojson layers using the `cluster.addLayers` to set layers to the cluster group

Test this by:
- [ ] The map loads to show all the markers loaded into it, no markers are left outside of the initial view after it finishes loading.
- [ ] When the map has more than 400 points , the view auto-adjusts progressively as points are loaded into it, but it doesn't feel like it's "shaking" or "jumping"

Fixes ushahidi/platform#2265.

Ping @ushahidi/platform

### DIFF
--- a/legacy/app/map/post-view-map.directive.js
+++ b/legacy/app/map/post-view-map.directive.js
@@ -77,32 +77,30 @@ function PostViewMap(PostEndpoint, Maps, _, PostFilters, L, $q, $rootScope, $com
             }
         }
 
-        function addPostsToMap(posts) {
+        function addPostsToMap(posts, shouldFitMapBoundary = true) {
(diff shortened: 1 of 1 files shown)


Ifycode (commented):
Works nice @webong 👍🏽

tuxpiper:
I think it needs another look 😬   I commented in Slack: https://ushahidi.slack.com/archives/CFC4WQG4V/p1659540157443289

tuxpiper (approved):
Ah yes, it was actually so simple :)

tuxpiper:
@AmTryingMyBest this is ready for QA

PR #1808, opened 2022-07-30
Fix: Platform is filtering out USSD messages

Ifycode:
**This pull request makes the following changes**
- Fixes ushahidi/platform#4430

**More on changes made**
- Video 1: https://ushahidi.slack.com/archives/CFC4WQG4V/p1659203074593859
- Video 2: https://ushahidi.slack.com/archives/CFC4WQG4V/p1659203132561779
- Video 3: https://ushahidi.slack.com/archives/CFC4WQG4V/p1659203162753399

**Testing checklist**
- [x] USSD filtering works as expected

Ping @ushahidi/platform


Ifycode:
Ready for review @tuxpiper @webong

tuxpiper:
I've added WhatsApp .

@Ifycode maybe we could make the change to show all the sources selected (checkboxes "ticked") by default?

I would just love if there was a place where we could write that once (`['sms', 'twitter','web', 'email','whatsapp','ussd']`), instead of hard-coding the same array all over the place. Do you think we could do that?

Ifycode:
Ok @tuxpiper please check if the latest commit solves both things you requested for.

tuxpiper:
This looks good :+1: I'm just thinking now that we should be thinking again before merging it into the project trunk, because 99.99% of the users don't have USSD or WhatsApp available.

We'll use it for Uchaguzi for now.

In the future, we should have some way for the backend to tell the frontend which datasources are actually relevant to the deployment.
(cc: @Mh-Asmi for awareness)

tuxpiper:
Converting to draft just to signify that we shouldn't be merging this to develop for now. We should not be releasing it to the wider audience.

legacy/app/common/global/event-handlers.js
```diff
@@ -14,4 +14,6 @@ function (
     $rootScope.toggleModalVisible = function (state) {
             $rootScope.modalVisible = (typeof state !== 'undefined') ? state : !$rootScope.modalVisible;
     };
+
+    $rootScope.sources = ['sms', 'twitter','web', 'email', 'whatsapp', 'ussd'];
```
tuxpiper, line 18: @Ifycode as one last thing to do before we can merge this to trunk branches...

Could this variable come from config.js ? Something like 'enabled_sources' that defaults to SMS, twitter, web and email.

That way we could keep away whatsapp and ussd for now, and only enable them in the few deployments where they are actually used.

tuxpiper (changes requested):

legacy/app/common/global/event-handlers.js
Ifycode: All right @tuxpiper. I will check out how to add it to config.

tuxpiper:
This should be good to proceed after #1825 is merged :+1:

tuxpiper (approved):

tuxpiper (approved):

PR #1816, opened 2022-08-10
Changes on the search to only search upon completion of typing and using the space bar to trigger the search.

chegejohn159:
-

Testing checklist:
- [ ]

Fixes ushahidi/platform#4474 .

Ping @ushahidi/platform


tuxpiper:
Thank you @chegejohn159 ! @Ifycode , could you please review?

tuxpiper:
@chegejohn159  mentioned to me that the use of the spacebar to send the query was taken because there were some security limitations with the use of the Enter key.

@chegejohn159 maybe you would like to talk a bit more about that here, for Mary and QA to take into consideration during review.

chegejohn159:
> @chegejohn159 mentioned to me that the use of the spacebar to send the query was taken because there were some security limitations with the use of the Enter key.
> 
> @chegejohn159 maybe you would like to talk a bit more about that here, for Mary and QA to take into consideration during review.

Hello, I used the spacebar instead of enter key as there was a problem to do with reloading caused by the enter key, I had to prevent default

Ifycode:
Oh, Just seeing this @tuxpiper @chegejohn159. Sorry about this.
 Expect feedback soon.

legacy/app/data/common/post-edit-create/location.directive.js
```diff
@@ -187,11 +187,15 @@ function PostLocationDirective($document, $http, L, Geocoding, Maps, _, Notify,
...
+
         function handleActiveSearch(event) {
-            var del = event.keyCode === 8 || event.keyCode === 46;
-            var letter = event.keyCode > 47 && event.keyCode < 58;
-            var number = event.keyCode > 64 && event.keyCode < 91;
-            if (del || letter || number) {
+            // var del = event.keyCode === 8 || event.keyCode === 46;
+            // var letter = event.keyCode > 47 && event.keyCode < 58;
+            // var number = event.keyCode > 64 && event.keyCode < 91;
+            if (event.keyCode === 32) {
```
Ifycode, line 198: A cool attempt at solving this @chegejohn159 

I have the following comments:
- Even though I see that you added this `<p>Press Space Bar After Typing Location</p>` in the html to direct the user to use the space bar, the problem is that (from experience) many users naturally expect to press enter key in this case based on the norm/convention. Some won't even read what you put there and just assume the feature doesn't work.
- About enter key and reload: It is the default behavior for enter key to reload. It is also a normal thing to use event.preventDefault() to prevent that reload. There is nothing wrong with doing that.

So I'd suggest removing the html instruction you added and use the enter key based on norm or what a user would naturally expect.

Here's what your handleActiveSearch function can look like:

 ```
function handleActiveSearch(event) {
            if (event.keyCode === 13) {
                event.preventDefault();
                $scope.processing = true;
                if ($scope.searchTimeout) {
                    clearTimeout($scope.searchTimeout);
                }
                $scope.searchTimeout = setTimeout($scope.searchLocation, 250);
            }
        }
```

**Review by Ifycode — CHANGES_REQUESTED** [review]

**Comment by chegejohn159** [comment]
@Ifycode 
- We addressed the changes requested. We changed the search to be triggered upon click of the enter key as opposed to the spacebar.

File: legacy/app/data/common/post-edit-create/location.directive.js [hunk 2]
```diff
@@ -187,11 +187,15 @@ function PostLocationDirective($document, $http, L, Geocoding, Maps, _, Notify,
...
+        function test1() {
+            console.log('chege test');
+        }
+
         function handleActiveSearch(event) {
-            var del = event.keyCode === 8 || event.keyCode === 46;
-            var letter = event.keyCode > 47 && event.keyCode < 58;
-            var number = event.keyCode > 64 && event.keyCode < 91;
-            if (del || letter || number) {
+            // var del = event.keyCode === 8 || event.keyCode === 46;
```
Ifycode, line 195: @chegejohn159 You can also clean up the code: remove the comments and remove the test1 function you added earlier. After that, let's see if @tuxpiper is ok with the latest changes.

Also I see some other files apart from the ones you edited. The entire .idea folder. You may want to remove those too.

Ifycode (changes requested):

legacy/app/data/common/post-edit-create/location.directive.js
chegejohn159: Done, removed .idea, removed the comments and the extra function

tuxpiper:
@Ifycode can you please confirm that the changes satisfy your request, and then let QA know they test this one?

Thanks everyone! 🙏

legacy/app/data/common/post-edit-create/location.directive.js
Ifycode: It is not complete @chegejohn159 
Restore the package-lock files too to be on the safe side. The only file you made a change to is the location directive. So when you click on the "Files changed" tab here on this PR, the location directive file should be the only file shown as changed there.
That is what I meant by restoring all other changes.
Once you do this, PR will be approved.

Ifycode:
Sorry @tuxpiper Just seeing this. Left a comment for @chegejohn159 

> It is not complete @chegejohn159 Restore the package-lock files too to be on the safe side. The only file you made a change to is the location directive. So when you click on the "Files changed" tab here on this PR, the location directive file should be the only file shown as changed there. That is what I meant by restoring all other changes. Once you do this, PR will be approved.

tuxpiper:
Dear @chegejohn159 , could you please make the changes asked by @Ifycode above?

Otherwise, we would need to re-create this pull request ourselves and your attribution wouldn't be so strong 🙏

chegejohn159:
@tuxpiper Kindly confirm the requested edits

Ifycode:
@chegejohn159 please send me a dm on Uchaguzi slack channel.

tuxpiper:
@Ifycode what is the status of review for this one? Is it still pending changes?

Ifycode:
Yes @tuxpiper It still has pending changes.
@chegejohn159 was supposed to send a new PR since it looks like the latest requested changes became confusing for him.

tuxpiper:
Closing in favor of #1823

PR #1820, opened 2022-08-18
Fix: Filters are reset when the page is reloaded 

Ifycode:
- Fixes ushahidi/platform#4469

Testing checklist:
- [x] Filter is retained on page reload (but reset on login or logout)

Ping @ushahidi/platform

### DIFF
--- a/legacy/app/auth/authentication-events.run.js
+++ b/legacy/app/auth/authentication-events.run.js
@@ -91,6 +91,7 @@ function AuthenticationEvents($rootScope, $location, Authentication, Session, _,
     }
 
     function doLogin(redirect, noReload) {
+        localStorage.removeItem('ush-filterState-2');
         TermsOfService.getTosEntry()
             .then(function () {
                 loadSessionData();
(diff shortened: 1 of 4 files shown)


Ifycode:
@tuxpiper @webong This is ready for review

Ifycode:
Has issues. Converting to draft for now, till it is resolved.

tuxpiper:
From chat:

> There seems to be a slight issue (with the fix for filters reset when the page is reloaded). For every other filtering, everything works fine. But in the case of filtering using a  saved search that e.g. only admins are permitted to access… If the admin logs out while such filter is still on, trying to access the data view and map view from the modebar brings up the login modal instead of taking the not logged in user to the data or map view.

> The restore default button also has stopped working as expected.

Ifycode:
Update: The two issues attached to the restore default button after page reload have been fixed. Only the admin logout issue is left - https://ushahidi.slack.com/archives/CFC4WQG4V/p1660854328246289

tuxpiper:
@webong Could you please review?

tuxpiper:
@webong is it feasible for you to review this?

webong:
Tested and works great ... Good work @Ifycode 

@AmTryingMyBest ready for QA

webong (approved):

